PR #56, opened 2015-08-06
Displaying error if no default acc record type has been set.

ceiroa:
@kbromer, this includes the change you just requested. What do you think?


kbromer:
Perfect - thx

PR #109, opened 2015-11-06
W-008542: schedule seasonal addresses updates from install script.

ceiroa:


davidhabib:
@ceiroa Why did you need to do this?  does HEDAP no longer use our Schedulable object?  If not, how will you avoid push upgrade issues?

ceiroa:
@davidhabib ~ Good point. We considered it unnecessary to use the Schedulable framework that we have in HEDA, given that the limits have increased to 100 jobs in the queue, and that we only have schedulable. But we should still instantiate the class dynamically to avoid locking it (or avoid having to perform two push upgrades, one to cancel the job and another to update the class and reschedule the job). I have created a work item for it: https://foundation--agf.na6.visual.force.com/apex/adm_userstorydetail?id=a2x800000001h9lAAA

PR #119, opened 2015-11-30
Update README

ceiroa:


ceiroa:
@daveperry ~ I added some content to the README in order to have something there until @jvarese has the chance to work on this item. Please review and merge if you think this is ok.

PR #126, opened 2015-12-11
Updating name of static resource.

ceiroa:


ceiroa:
@daveperry - Could you take a quick look and merge? It's just an update to remove version number from the SLDS static resource.

PR #128, opened 2015-12-21
Feature/8837 accessibility

ceiroa:
@daveperry - I'm going to merge in order to test it as managed package and see if there are prefix-related issues.


daveperry:
No problem.

*Dave Perry*
*Technical Architect - Higher Education*
Salesforce.org
385-233-8075

2016 Higher Ed Summit registration is LIVE!
<http://www.salesforce.org/highered/hesummit16/>



<http://www.salesforce.org>

--

PR #129, opened 2015-12-21
Feature/8837 accessibility

ceiroa:


src/aura/STG_CMP_Base/STG_CMP_BaseHelper.js
```diff
@@ -130,5 +130,16 @@
 			prefix = namespacePrefix;
 		}
 		component.set(messageElement, labels[prefix][label]);
+	},
+	
+	//Any label referenced in any of the components will only be available if it's being called like this
+	//somewhere in the code. Note that this method isn't even called from anywhere.
+	hugeHack : function() {
+		$A.get("$Label.c.noAfflMappings");
```
ceiroa: @kbromer ~ We need to talk about this, and probably tell core about it. It's nuts we have to do this...

PR #135, opened 2015-12-22
Displaying delete icon only in Edit mode.

ceiroa:


ceiroa:
@daveperry ~ Could you take a quick look and merge? Thanks!

PR #136, opened 2015-12-22
Feature/8888 field changes

ceiroa:
W-008955: s/HIPPA/HIPAA/
W-008888: Class Year field on Program Enrollment should be text, not a number


ceiroa:
@daveperry ~ Could you please review and merge? thanks

PR #138, opened 2015-12-23
Adding graduation year field (instead of class year, which we had to …

ceiroa:
…delete because it was of the wrong type).


ceiroa:
@daveperry ~ Could you review and merge? I have changed the field name because we cannot create another field with the same name but different type. We can always change the label.

PR #139, opened 2015-12-28
Feature/8899 no admin affl

ceiroa:
I have commented out the code that automatically creates an Affiliation to its parent Account when a Contact is created, and I have fixed the tests to not assume that Affiliation will be there (and also fixed a bug in an unrelated area which surfaced due to these changes). 

In addition to what is done here, we could make it configurable, and let the administrator decide whether the Affiliation will be created.

Also, this made me realize that none of the Account record types are in the package. We should consider including at least Organizational and Household, specially because many tests depend on these. Installation of managed package will fail if they aren't there.


ceiroa:
@daveperry ~ We could merge it now, or wait until we decide whether we are going to make this configurable. After looking into it a little bit more, I don't think we lose any important functionality with these changes.

ceiroa:
Hey, @daveperry 

Could you please review and merge? It turns out I do need this code in, or I'm going to have a conflict with the branch I'm working on right now.

Thanks

PR #141, opened 2015-12-30
Feature/8631 extra affl

ceiroa:
Removing comments, pulling in some minor metadata changes from the org, and putting the Account Name back at the top in the Contact layout.


ceiroa:
@daveperry ~ Please take a look and merge when ready. Thanks.

PR #142, opened 2015-12-31
Feature/7887 compact layouts

ceiroa:
Compact layouts and updated related lists.


ceiroa:
@daveperry ~ Could you please review and merge? Also, an additional configuration step is now required to make the HEDA Compact Layouts the default in the org (it's assigned globally, not per profile). Let me know if you have any questions regarding this configuration. Thanks.

PR #143, opened 2016-01-05
Feature/7633 start end year

ceiroa:


ceiroa:
Hey @davidhabib, could you please take a look at this and let me know where else you think changes need to be made to support seasonal addresses start and end year? (There are couple more small things in this pull request, but they are just small metadata changes. I think it's easy to spot them.)

src/classes/ADDR_Addresses_TDTM.cls
```diff
@@ -370,15 +370,17 @@ public class ADDR_Addresses_TDTM extends TDTM_Runnable {
     * @return Date
     */ 
     private static Date dtStartOrEndOfAddrSeasonal(Address__c addr, boolean fStart) {    
+        integer startYear = integer.valueOf(addr.Seasonal_Start_Year__c);
```
davidhabib: won't this throw an exception if Seasonal_Start_Year__c is null?

davidhabib: won't isSeasonalOverlap() need to be modified to look at the years as well (if they are specified)?

ceiroa: I didn't modify isSeasonalOverlap because of the comment in line 583 (expand code to see) that says "we don't need to know what year or month we are currently in, but we need to take into account days (since we could be in the same month)." But I didn't dig into the logic in depth. I'll take a look.

ceiroa: Regarding the first comment, I tested the logic in the Anonymous window of the debug console with the following code, and no exception is thrown:

Decimal test = null;
integer startYear = integer.valueOf(test);
System.debug('Start year: ' + startYear);

Which makes sense (though Apex sometimes surprises you), since the Integer value of null should just be null.

ceiroa:
@davidhabib ~ The pull request is finally ready for review. I have made the changes we talked about, fixed couple of bugs, and created a set of tests for the additional functionality. I have also changed the name of a couple of methods and variables, just to make the code easier to read. Hopefully that won't be too confusing. :-) Oh, and I finally decided to remove that section of code that I had initially commented out just for cleanliness. The file is long enough already. I don't thing this makes the diff harder to read.

@daveperry ~ Please feel free to review it as well and make any comments, but please don't merge it. I'd like to get the thumbs up from DH, since he wrote all the original seasonal addresses code. Thanks.

src/classes/ADDR_Addresses_TDTM.cls
```diff
@@ -366,31 +366,44 @@ public class ADDR_Addresses_TDTM extends TDTM_Runnable {
...
+        //start and end year have been provided (we don't allow providing only start or only end - addrSeasonalPartialYear error)
+        if(startYear != null && endYear != null) {
+            startDate = date.newInstance(startYear, startMonth, startDay);
+            endDate = date.newInstance(endYear, endMonth, endDay);
+        //start and end year have not been provided
         } else {
-            dtStart = date.newInstance(dtToday.year()-1, startMonth, startDay);
-            dtEnd = date.newInstance(dtToday.year(), endMonth, endDay);
+	        //seasonal address is contained within a single year
+	        if (startMonth <= endMonth && startDay <= endDay) {
```
davidhabib: I think it is incorrect to also be requiring startDay <= endDay.  your new logic will fail if the startdate is June 15, and the enddate is September 1.

src/classes/ADDR_Addresses_TDTM.cls
```diff
@@ -366,31 +366,44 @@ public class ADDR_Addresses_TDTM extends TDTM_Runnable {
     /*******************************************************************************************************
     * @description utility to return the current StartDate or EndDate of a Seasonal Address
     * @param addr the Address to look at
-    * @param fStart whether to return the StartDate or EndDate
+    * @param returnStartDate whether to return the StartDate or EndDate
     * @return Date
     */ 
-    private static Date dtStartOrEndOfAddrSeasonal(Address__c addr, boolean fStart) {    
+    private static Date getStartOrEndDateOfAddrSeasonal(Address__c addr, boolean returnStartDate) {
```
davidhabib: I don't think renaming my methods and variables is in the best interest of sharing code, just because you prefer your naming conventions.  I try to always respect the existing naming and syntax conventions of any shared code I am modifying.

dextermilo: Except when people use Hungarian notation in any form, then it should be allowed...  ;)

src/classes/ADDR_Addresses_TDTM.cls
ceiroa: Thanks for catching this. I made this change because the last check in testSeasonalAddrOverlapLogic was failing. However, what I should have written is what I'm going to commit in just a sec (together with a change that would have caught the issue you mention).

ceiroa:
@davidhabib Did my last commit on this pull request make sense? Do you feel it properly addresses the issue without any negative side effects? Thanks.

PR #144, opened 2016-01-12
namespacePrefix attribute can be defined only in Base component and i…

ceiroa:
…t'll properly inherited.


ceiroa:
@daveperry Ready for review. This is just a small change that takes advantage of inheritance in lightning components.

PR #145, opened 2016-01-15
Feature/9110 reciprocal relationships

ceiroa:
Fixes #9110: incorrect relationship explanation.

### DIFF
--- a/src/classes/REL_Relationships_TDTM.cls
+++ b/src/classes/REL_Relationships_TDTM.cls
@@ -287,7 +287,7 @@ public class REL_Relationships_TDTM extends TDTM_Runnable {
 
                 Contact relatedContact = cMap.get(r.Contact__c);
             
-                if (rlMap.containsKey(r.Type__c)){                                      
+                if (rlMap.containsKey(r.Type__c)){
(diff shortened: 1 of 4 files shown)


ceiroa:
@daveperry, this pull request is ready to be reviewed and merged. Other than some minor lowercase to uppercase changes, them main changes are:

- Removing the getAll() custom settings calls and replacing them with calls to the Facade
- Caching of list settings in the Facade
- Fixing default custom setting value that was causing relationship explanation to be incorrect
- Adding default relationship lookups to be inserted on install

@njjc, this includes all we were talking about during the scrum. You may also have to create the getReciprocalSettings and getAutoCreateSettings, though in your case they don't need to be aura-enabled since your settings page is not in lightning.

njjc:
Thanks @ceiroa, I was wondering about your reasoning behind using a query instead of getAll, since when we add new fields we'll have to remember to add them to the query. Also, why the 50k limit?

ceiroa:
No good reason, really. Initially I wrote it like that and I was using it just to call it from the settings page. Then I gave it the limit just because it was flagged as warning in the sec review. Using getAll would be totally fine, I think.

njjc:
Got it, thx!

PR #147, opened 2016-01-26
Parameter passed from JS to Apex had the wrong name.

ceiroa:
# Warning

# Info
- Properly saving changes to Affiliation Mappings in Settings (9194).


ceiroa:
@daveperry In theory this could be merged as well, but the status is not green because of a "default" build that apparently is running. Let's wait until @jlantz clarifies what's going on with the build jobs before merging.
screen shot 2016-01-26 at 2 39 44 pm

ceiroa:
@jlantz ~ is it ok to merge this pull request now?

PR #148, opened 2016-01-29
Feature/label change

ceiroa:
# Warning

# Info
- Changing label from Department to Academic Program in Course Enrollment, for ease of use.
- Providing object descriptions.
- Enabling reports on custom objects.
- Fixing bug in seasonal address overlap detection.

### DIFF
--- a/src/classes/ADDR_Addresses_TDTM.cls
+++ b/src/classes/ADDR_Addresses_TDTM.cls
@@ -105,7 +105,7 @@ public class ADDR_Addresses_TDTM extends TDTM_Runnable {
 	            if (triggerAction == TDTM_Runnable.Action.BeforeInsert) {
 	               // when an address is marked Default, we update its latest date fields
 	               if (addr.Default_Address__c) { 
-	                   addr.Latest_Start_Date__c = system.today();
+	                   addr.Latest_Start_Date__c = System.today();
 	                   addr.Latest_End_Date__c = null;
 	               }
 	            }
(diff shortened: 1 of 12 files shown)


ceiroa:
@jlantz ~The branch status is green after I pushed to the remote, even though the job has not finished running as of right now.

ceiroa:
@davidhabib You may want to take a look at the last commit. As a result of a failing test, I fixed the logic of the overlap detection method (dtStartOrEndOfAddrSeasonal in NPSP), and added some more tests. I think this problem might also exist in NPSP.

ceiroa:
@daveperry ~ This is ready for review and merge. Thanks!

davidhabib:
@ceiroa I looked at your changes to getStartOrEndDateOfAddrSeasonal, and I believe you have broken the logic at line 252 in refreshCurrentHHAddress().  Your change always returns the start and end dates of a seasonal address that overlaps years to be the previous year.  Ie, if start was 10/1 and end is 2/1, and let's say on 12/1/2015 we were running the logic to see if we should use the seasonal address (which we should since it is in the 10/1 - 2/1 range).  But your change will return start = 10/1/2014 and end = 2/1/2015, so given its 12/1/2015, you won't apply the seasonal address.

ceiroa:
@davidhabib ~ Ok, I see this issue. But the problem was that the test testSeasonalAddrOverlapLogic in ADDR_Addresses_TEST had stared failing at https://github.com/SalesforceFoundation/HEDAP/blob/master/src/classes/ADDR_Addresses_TEST.cls#L1823 because the first address was from Feb 1st to Feb 10th of 2016, and the second one from Feb 11th to Feb 1st of 2017. Any thoughts on how to solve it without breaking refreshCurrentHHAddress?

davidhabib:
the problem got introduced when isSeasonalOverlap() was changed from the NPSP logic to use getStartOrEndDateOfAddrSeasonal().  the problem is getStartOrEndDateOfAddrSeasonal() is dependent on the current date (due to the reasons I explained previously).  but isSeasonalOverlap() should not include the year component if the seasonal addresses don't explicitly include year.  So I think you need to restore getStartOrEndDateOfAddrSeasonal() to what it was, and then redesign HEDAP's implementation of isSeasonalOverlap().

PR #150, opened 2016-02-05
Feature/9272 recursive update

ceiroa:
# Warning 

# Info
- Avoid error notification being sent when populating primary affiliation fields on Contact in edit mode (W-009272)

### DIFF
--- a/src/classes/AFFL_AccRecordType_TEST.cls
+++ b/src/classes/AFFL_AccRecordType_TEST.cls
@@ -58,12 +58,14 @@ public with sharing class AFFL_AccRecordType_TEST {
 
         Contact contact = new Contact(FirstName = 'Test', LastName = 'Testerson');
         insert contact;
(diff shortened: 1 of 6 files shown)


ceiroa:
@kbromer ~ The fix was really just wrapping 
    relatedContact.put(lookupFieldName, affl.Account__c);
inside the 
    if(CON_PrimaryAffls_TDTM.keyAfflLookupUpdated == false)
condition.

The interesting thing is that the contact record was being processed in the transaction, but was not inside our dmlWrapper. That's why an error was being thrown, even though we  were wrapping 
with that same condition.

Even though the fix was really easy it took me sometime to get all the tests working because initially I was trying to use a different static flag...

ceiroa:
This is ready to be reviewed, @daveperry. Thanks!

ceiroa:
Hey @daveperry, could you review and merge this? Thanks.

PR #170, opened 2016-02-23
Feature/169 affl contact change

ceiroa:
# Warning

# Info

# Issues
- Renaming HEDAP.app to HEDA.app
- Renaming package to HEDA
- Change auto-enrollment to Academic Program record type in settings. Also setting status and role.


cumulusci.properties
```diff
@@ -1,4 +1,4 @@
-cumulusci.package.name=HEDAP
+cumulusci.package.name=HEDA
```
ceiroa: Hey, @jlantz ~ This combined with the change to package.xml below should be ok, right? I figured I'd update it if it's feasible, now that we are starting from scratch.

PR #172, opened 2016-02-25
Changing master-detail field in Address to lookup.

ceiroa:
# Warning

# Info
- Changing master-detail field in Address to lookup.

# Issues


ceiroa:
@kbromer ~ This is the item we were discussing in chatter. After making the change the feature build passes without hiccups. Should we go ahead and merge it?

PR #207, opened 2016-03-17
Feature/205 acc contact del

### LINKED ISSUE #205: Do not allow deleting a Contact if it’s referenced by one or more Affiliations, Program Enrollments, or Course Enrollments
It'll be in its own TDTM class so those that don't want this restriction can disable it.

### LINKED ISSUE #206: Do not allow deleting an Account if it’s referenced by one or more Affiliations, Program Enrollments, or Course Enrollments
It'll be in its own TDTM class so those that don't want this restriction can disable it.

ceiroa:
# Warning

# Info
- Adding classes that will stop a user from deleting an Account or Contact if it has child Affiliation, Program Enrollment, or Course Enrollment records. This logic can be turned off by setting the Active field to false for the ACCT_CannotDelete_TDTM and CON_CannotDelete_TDTM entries in the Trigger Handler object.

# Issues
Fixes #205.
Fixes #206.

### DIFF
--- a/ApexDocContent/Accounts.htm
+++ b/ApexDocContent/Accounts.htm
@@ -11,6 +11,7 @@
         when a Contact is created. Administrative and Household Accounts are recommended as the preferred Account Model.</p>
             
 	<ul>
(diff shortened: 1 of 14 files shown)


ceiroa:
Ready for review, @daveperry.

PR #210, opened 2016-03-17
Renaming name field in Affiliation and Relationship.

### LINKED ISSUE #208: Consistency naming Relationship and Affiliation name field
For affiliations we call the Id field “Affiliation Name” and for relationships we call the Id field “Relationship Number” and they are both auto number. Let's be consistent in how we do this. If I could change everything I would call it a “Key” and not an Id, Name or Number.

ceiroa:
# Warning

# Info

# Issues
Fixes #208.


ceiroa:
Ready for review and merge, @daveperry.

PR #214, opened 2016-03-22
Removing unnecessary system field from Account. Looking instead at the account's record type.

### LINKED ISSUE #212: Try removing SYSTEM_AccountType__c field in Account
It's possible we can just look at the Account record type.

ceiroa:
# Warning

# Info
- Field SYSTEM_AccountType__c in Account can now be safely deleted from the user's org.

# Issues
Fixes #212.

### DIFF
--- a/src/classes/ACCT_IndividualAccounts_TDTM.cls
+++ b/src/classes/ACCT_IndividualAccounts_TDTM.cls
@@ -41,6 +41,10 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
     */ 
     private static Hierarchy_Settings__c ContactsSettings;
     
+    private static ID orgRecTypeID = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get('Business Organization').getRecordTypeId();
+    private static ID householdRecTypeID = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get('Household Account').getRecordTypeId();
+    private static ID adminRecTypeID = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get('Administrative').getRecordTypeId();
+    
     /*******************************************************************************************************
     * @description a set of languages that require different naming conventions
     */ 
@@ -91,7 +95,7 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
            
            // requery to get correct Account values (and all the other fields we will look at)
            string strSoql = 'select id,AccountId,' +
-                'Account.SYSTEM_AccountType__c,' +
+                'Account.RecordType.Name,' +
                 'Account.Primary_Contact__c,Account.Name,' +
                 'firstname, lastname, OwnerId, Salutation,' +
                 'MailingStreet, MailingCity, MailingState, MailingPostalCode, MailingCountry, MailingLatitude, MailingLongitude, ' + 
@@ -132,8 +136,8 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
                     }
 
 	                // HHAccount renaming needs to be called to set various HH naming fields like greeting.
-	                if (c.AccountId != null && (c.Account.SYSTEM_AccountType__c == CAO_Constants.HH_ACCOUNT_PROCESSOR
-	                || c.Account.SYSTEM_AccountType__c == CAO_Constants.ADM_ACCOUNT_PROCESSOR))
+	                if (c.AccountId != null && (c.Account.RecordType.Name == CAO_Constants.HH_ACCOUNT_PROCESSOR
+	                || c.Account.RecordType.Name == CAO_Constants.ADM_ACCOUNT_PROCESSOR))
 	                   listAccountIdHHToRename.add(c.AccountId);    
 	            }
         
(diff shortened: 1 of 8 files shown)


ceiroa:
Hey @kbromer, could you please review and merge this one? Thanks.

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -41,6 +41,10 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
     */ 
     private static Hierarchy_Settings__c ContactsSettings;
     
+    private static ID orgRecTypeID = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get('Business Organization').getRecordTypeId();
```
kbromer: These recordtypes aren't necessarily hardcoded, are they? What happens if the user's recordtypes are something other than the default?

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -346,10 +350,10 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
                 ADDR_Addresses_TDTM.copyAddressStdSObj(c, 'Mailing', a, 'Billing');
                 ADDR_Addresses_TDTM.copyAddressStdSObj(c, 'Other', a, 'Shipping');
 
-                if(CAO_Constants.isHHAccountModel()) {
-                    a.SYSTEM_AccountType__c = CAO_Constants.HH_ACCOUNT_PROCESSOR;   
-                } else if(CAO_Constants.isAdmAccountModel()) {
-                    a.SYSTEM_AccountType__c = CAO_Constants.ADM_ACCOUNT_PROCESSOR;
+                if(CAO_Constants.isHHAccountModel() && householdRecTypeID != null) {
```
kbromer: they'd just get the default?

src/classes/ADDR_Addresses_TDTM.cls
```diff
@@ -290,7 +290,7 @@ public class ADDR_Addresses_TDTM extends TDTM_Runnable {
         set<Id> setHHId =  mapAccIdAddr.keySet();
         list<Contact> listCon = [select Id, is_Address_Override__c, Current_Address__c, AccountId 
             from Contact where 
-                (Account.SYSTEM_AccountType__c = :CAO_Constants.HH_ACCOUNT_PROCESSOR or Account.SYSTEM_AccountType__c = :CAO_Constants.ADM_ACCOUNT_PROCESSOR)
+                (Account.RecordType.Name = :CAO_Constants.HH_ACCOUNT_PROCESSOR or Account.RecordType.Name = :CAO_Constants.ADM_ACCOUNT_PROCESSOR)
```
kbromer: You know, if we were smart, we'd be storing this as the id instead of the name... am I crazy for thinking that's safer?

src/classes/ADDR_Addresses_TEST.cls
```diff
@@ -61,7 +61,17 @@ private with sharing class ADDR_Addresses_TEST {
...
     * @description The list of created test Addresses.
     */
-    private static list<Address__c> listAddrT;           
+    private static list<Address__c> listAddrT;
+    
+    private static ID orgRecTypeID;
+    private static ID householdRecTypeID;
+    private static ID adminRecTypeID;
+    
+    public static void recTypesSetup() {
```
kbromer: same as above - can these be hardcoded?

src/classes/ACCT_IndividualAccounts_TDTM.cls
ceiroa: Yes, they should get the default for the profile.

src/classes/ADDR_Addresses_TDTM.cls
ceiroa: Yeah, I was thinking about this earlier... This doesn't support other languages or changing the account record type label. So it's not very robust. But if we were to change it, we would need some kind of configuration. Maybe asking what is the ID of the record type that you consider Administrative, and the record type that you consider Household. It feels like a bigger issue than the one we are dealing with here, though...

src/classes/ADDR_Addresses_TEST.cls
ceiroa: We don't include any Account record types in the package, so we just query to see if they are there. If the record types aren't there the tests will fail, which is why at some point I proposed included those record types that are used in tests to be included in the package. But we decided not to do that.

ceiroa: Here is the record types issue in AA ("never" status): https://foundation--agf.na6.visual.force.com/apex/adm_userstorydetail?id=a2x800000001i8T, and the doc where we discussed this: https://docs.google.com/document/d/1IYYfMHt-5eN6uspDaG9MGHDB_PqNSVFuA8JrnHV009Q/edit. I'm open to revisiting this and trying to come up with a better design, if we think it's worth it.

kbromer: Yeah, I'm not going to hold this up for that, but I do think there's a better way to handle it. We're using a pattern from 2007 right now.  I don't think including the types is the right answer though - if we're storing the relevant type in a setting somewhere, we can retrieve it that - I can't recall, is that in there?

src/classes/ADDR_Addresses_TDTM.cls
kbromer: Agreed - lets look at it as I mentioned below

Issue #223, opened 2016-03-30
De-duplicate Household and Administrative tests after #216 changes


ceiroa:
Many of our tests are duplicate, to cover both the Household and the Administrative Account models. Since we are now going to allow any piece of functionality to be available to any record type, we should not have duplicate tests and instead have tests that verify any functionality works on any record type.

ceiroa:
**lurch: attach W-009783

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000003QepJAAS" target="_blank">W-009783</a>

mrbelvedere:
Included in production release version 1.15

Issue #222, opened 2016-03-30
Use setting to determine which Accounts need renaming


ceiroa:
Currently the method needsHHAccountRename in ACCT_IndividualAccounts_TDTM is not called from anywhere. We either remove it or we allow the user to determine, through a setting, which type of accounts will be renamed.

ceiroa:
**lurch: attach W-009781

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000003Qep9AAC" target="_blank">W-009781</a>

ceiroa:
@jvarese ~ I'm not going to open a ticket for this one in AA yet, as we don't know what we are going to do with it at this point...

jlantz:
Included in beta release version 1.10 (Beta 3) and higher

mrbelvedere:
Included in production release version 1.10

Issue #221, opened 2016-03-30
Use setting to determine if Account owner should be changed


ceiroa:
Method "updateOwners" in ACCT_IndividualAccounts_TDTM updates the Account owner if it's a Household or Administrative Account. It should be reading from a setting instead why types of Accounts need this.

ceiroa:
**lurch: attach W-009780

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000003Qep4AAC" target="_blank">W-009780</a>

kporray1:
This issue was resolved in December 2019.

Issue #220, opened 2016-03-30
Use setting to determine which type of accounts support seasonal addresses


ceiroa:
For this and similar functionality we should probably have a method that gets an account record type ID and returns true or false depending on whether functionality is active for that type. Something like: 

public static Boolean isSeasonalAddrActiveForRecType(ID recTypeID)

ceiroa:
**lurch: attach W-009782

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000003QepEAAS" target="_blank">W-009782</a>

jlantz:
Included in beta release version 1.8 (Beta 5) and higher

mrbelvedere:
Included in production release version 1.8

PR #226, opened 2016-03-30
Feature/216 acc rec types change

### LINKED ISSUE #216: Allow for Account record type labels to be changed
Right now, if the label of the Administrative of the Household Account record types get changed, all the related functionality will not work. This means they are not translatable into other languages.

We should instead just ask the user for the ID of the record type that he wants to use when an Contact is created without a parent Account. Or (even better), display all the Account record types that exist and let them choose. This avoids them having to deal with IDs.

We can also asks for the record types for which they want addresses to work. Or should we just support Address management in all record types? I don't see a reason why not to...

ceiroa:
# Warning

# Info
- Account record type labels can now be changed (and translated) without affecting functionality. Internally we are dealing only with the record type IDs.
- We are now displaying a drop-down with all available Account record types under HEDA Settings > System > Account Processor.

# Issues
Fixes #216.


ceiroa:
@kbromer ~ There are a number of issues related to this (#219 #220 #221 #222 #223), but I'm putting in this pull request first because if we were to have one single pull request with all these issues together it'd be really hard to debug.

src/aura/STG_CMP_System/STG_CMP_System.cmp
```diff
@@ -2,17 +2,24 @@
     <aura:handler name="init" value="{!this}" action="{!c.init}"/>
     
     <aura:handler event="c:STG_EVT_Edit" action="{!c.toggleIsView}"/>
+    <aura:handler event="c:STG_EVT_Save" action="{!c.saveAccRecType}"/>
     
-    <aura:attribute name="namespacePrefix" type="String" />
+    <aura:attribute name="accRecTypes" type="Map" />
```
kbromer: I assume these attribs are calling getAccountRecordTypes from somewhere in the base controller for this component?

src/aura/STG_CMP_System/STG_CMP_System.cmp
```diff
@@ -30,7 +37,7 @@
             <aura:if isTrue="{!v.isView}">
                 <ui:outputCheckbox value="{!v.hierarchySettings.Store_Errors_On__c}"/>
             <aura:set attribute="else">
-                <ui:inputCheckbox value="{!v.hierarchySettings.Store_Errors_On__c}" label="Store Erros On" labelClass="slds-hide"/>
+                <ui:inputCheckbox aura:id="storeErrors" value="{!v.hierarchySettings.Store_Errors_On__c}" label="Store Erros On" labelClass="slds-hide"/>
```
kbromer: what's the value in referencing this from the controller instead of directly like the view-based version above?

src/aura/STG_CMP_Tabs/STG_CMP_Tabs.cmp
```diff
@@ -32,7 +35,8 @@
              <c:STG_CMP_Addr hierarchySettings="{!v.hierarchySettings}" isView="{!v.isView}" namespacePrefix="{!v.namespacePrefix}" />
          </div>
          <div aura:id="systemTabContent" class="site-code--content slds-scrollable--x slds-tabs__content" role="tabpanel">
-             <c:STG_CMP_System hierarchySettings="{!v.hierarchySettings}" isView="{!v.isView}" namespacePrefix="{!v.namespacePrefix}" />
+             <c:STG_CMP_System hierarchySettings="{!v.hierarchySettings}" accRecTypes="{!v.accRecTypes}" accRecTypeId="{!v.accRecTypeId}" 
+             accRecTypeName="{!v.accRecTypeName}" isView="{!v.isView}" namespacePrefix="{!v.namespacePrefix}" />
```
kbromer: help me understand why we need to carry the namespace prefix here - the component doesn't recognize it?

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -37,12 +37,9 @@
...
     /*******************************************************************************************************
-    * @description the current Contact Settings object
-    */ 
-    private static Hierarchy_Settings__c ContactsSettings;
-    
-    private static ID householdRecTypeID = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get('Household Account').getRecordTypeId();
-    private static ID adminRecTypeID = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get('Administrative').getRecordTypeId();
+    * @description The selected Account record type.
+    */     
+    private static ID defaultRecTypeID = UTIL_CustomSettingsFacade.getSettings().Account_Processor__c;
```
kbromer: Is this pulling orgwide defaults or profile/user based?

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -120,32 +115,25 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
                 	
                 	// HH Account should be created if:
```
kbromer: Might want to update this comment, its not necessarily the HH Acocunt, right? or is it?

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -120,32 +115,25 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
...
                     //    1. If the contact is connected to an Account
-                    //    2. The Contact is flagged as an individual
-                    //    3. we are in the one-to-one or HH Account model
                     if (c.AccountId != null) { 
                         // contacts are connected to Accounts, make the connection in the other direction
                         mapAccountIdContactId.put(c.AccountId, c.Id);
-                    }
-
-	                // HHAccount renaming needs to be called to set various HH naming fields like greeting.
-	                if (c.AccountId != null && (c.Account.RecordType.Name == CAO_Constants.HH_ACCOUNT_PROCESSOR
```
kbromer: Where's the naming happening otherwise?

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -414,19 +387,20 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
...
             for (Account thisAccount : accountsOnContacts) {
                 //if the legacy or current type field show it's not a HH Account record, don't delete the account
-                if (thisAccount.RecordType.Name == CAO_Constants.HH_ACCOUNT_PROCESSOR || 
-                thisAccount.RecordType.Name == CAO_Constants.ADM_ACCOUNT_PROCESSOR) {
+                //TODO: this should be looking at a setting to determine which types can be deleted (W-009779)
+                if (thisAccount.RecordTypeId == UTIL_Describe.getHhAccRecTypeID() || 
+                thisAccount.RecordTypeId == UTIL_Describe.getAdminAccRecTypeID()) {
                     
                     // if there are no Opportunities for this Account, add it for deletion
                     if (thisAccount.Opportunities.size() == 0 && thisAccount.Contacts.size() == 0) {
```
kbromer: I'm not sure we need the Opportunities restriction on this either any longer. Typically it was done to prevent cascade deletion in scenarios where the contact still had donations, but in reality, there's enough other dependent information, that I'm thinknig we might need to just add another issue to determine what the correct cascade delete behavior we want to enable it

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -509,4 +484,8 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
     public static String strNameAdmAccountForContact(Contact con) {
         return con.LastName + ' ' + system.label.DefaultAdminName;
     }
+    
+    public static String strNameAccountForContact(Contact con) {
```
kbromer: Where is this being called at? Did I miss a reference somewhere?

src/classes/UTIL_Describe.cls
```diff
@@ -370,4 +370,108 @@ public class UTIL_Describe {
...
+     * @description Returns a map of active, user-available RecordType IDs for a given SObjectType, keyed
+     * by either the RecordType's Name or the Developer Name, depending on the value of the second param. 
+     * @param objectName Name of the object to retrieve record types for.
+     * @param devName True if we want the key of the returned map to be the record type's developer name.
+     * False if we want it to be the friendly name.
+     * @return Map<String, Id> Map of Name-to-ID or DeveloperName-to-ID (depending on the second paramater), 
+     * of available record types for object specified. 
+     */
+    private static Map<String, Id> getRecTypesMap(String objectName, Boolean devName) {      
+        if(mapRecordTypes == null) {
```
kbromer: Is there ever a scenario we'd want to force intelligent retrieval of the recordtypes at runtime - an appending to the map? Long running transactions, or anything where the state is maintained?

src/classes/ADDR_Addresses_TEST.cls
```diff
@@ -68,9 +68,9 @@ private with sharing class ADDR_Addresses_TEST {
     private static ID adminRecTypeID;
     
     public static void recTypesSetup() {
-        orgRecTypeID = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get('Business Organization').getRecordTypeId();
-        householdRecTypeID = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get('Household Account').getRecordTypeId();
-        adminRecTypeID = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get('Administrative').getRecordTypeId();
+        orgRecTypeID = UTIL_Describe.getBizAccRecTypeID();
```
kbromer: Are we predefining this? Providing a setting for base organizations?

src/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -91,7 +91,7 @@ public without sharing class UTIL_CustomSettingsFacade {
      * Private methods that will configure the default org-level custom settings.
      **/
     private static void configSettings(Hierarchy_Settings__c hs) {
-		hs.Account_Processor__c = 'Administrative';
+		hs.Account_Processor__c = UTIL_Describe.getRecTypesMapByDevName('Account').get('Administrative');
```
kbromer: So, what happens if the recordtype name is something other that Administrative? Administrativo?

src/classes/UTIL_Describe.cls
```diff
@@ -370,4 +370,108 @@ public class UTIL_Describe {
...
+            mapRecordTypes = new Map<String, Id>();
+        
+            // Get the Describe Result
+            fillMapsForObject(objectName);
+            Schema.DescribeSObjectResult obj = objectDescribes.get(objectName);
+        
+            // Obtain ALL Active Record Types for the given SObjectType objectType
+            String soql = 'SELECT Id, Name, DeveloperName FROM RecordType ' + 'WHERE SObjectType = \''
```
kbromer: Would this support retrieval of the recordtypes for a managed object? I assume not...

src/classes/UTIL_Describe.cls
```diff
@@ -370,4 +370,108 @@ public class UTIL_Describe {
...
+        
+            // Obtain the RecordTypeInfos for this SObjectType objectType
+            Map<Id,Schema.RecordTypeInfo> recordTypeInfos = obj.getRecordTypeInfosByID();
+        
+            for (SObject rt : results) { 
+                // Check RecordType IS available for the running user
+                if (!Test.isRunningTest()) {
+                    if (recordTypeInfos.get(rt.Id).isAvailable()) {
+                        if(devName) {
+                            mapRecordTypes.put(String.valueOf(rt.get('DeveloperName')), rt.Id);
```
kbromer: Sorry, i'm not sure i'm following this - why are you pulling 'DeveloperName' here?

src/labels/CustomLabels.labels
```diff
@@ -26,13 +26,21 @@
     </labels>
     <labels>
         <fullName>DefaultHouseholdName</fullName>
-        <categories>Households, Household Naming</categories>
+        <categories>Account Naming</categories>
         <language>en_US</language>
         <protected>false</protected>
         <shortDescription>Default Household Name</shortDescription>
```
kbromer: Change the 'shortDescription' as well

src/aura/STG_CMP_System/STG_CMP_System.cmp
ceiroa: Those attributes are being passed from the container component: https://github.com/SalesforceFoundation/HEDAP/blob/feature/216-acc-rec-types-change/src/aura/STG_CMP_Tabs/STG_CMP_Tabs.cmp#L38

src/aura/STG_CMP_System/STG_CMP_System.cmp
ceiroa: Not sure I follow. What line are we comparing line 40 with?

src/aura/STG_CMP_Tabs/STG_CMP_Tabs.cmp
ceiroa: The main reason is that you need the namespace when displaying labels, like in https://github.com/SalesforceFoundation/HEDAP/blob/feature/216-acc-rec-types-change/src/aura/STG_CMP_System/STG_CMP_System.cmp#L27. There is no way to choose it dynamically.

src/classes/ACCT_IndividualAccounts_TDTM.cls
ceiroa: The getSettings method in the Façade always tries to get the user/profile ones first: https://github.com/SalesforceFoundation/HEDAP/blob/feature/216-acc-rec-types-change/src/classes/UTIL_CustomSettingsFacade.cls#L64

src/classes/ACCT_IndividualAccounts_TDTM.cls
ceiroa: We don't yet have custom naming (we have an open work item for that). Standard account naming happens in https://github.com/SalesforceFoundation/HEDAP/blob/feature/216-acc-rec-types-change/src/classes/ACCT_IndividualAccounts_TDTM.cls#L480-L500

src/classes/ACCT_IndividualAccounts_TDTM.cls
ceiroa: We have this open issue: https://github.com/SalesforceFoundation/HEDAP/issues/219. Do you think specifying the types is enough, or do we need a "Delete acc only if no child opps exist" setting?

src/classes/ADDR_Addresses_TEST.cls
ceiroa: We provide a number of account record types, including one for business organizations, in an unmanaged bundle that the installed puts in before the package.

src/classes/UTIL_CustomSettingsFacade.cls
ceiroa: This is looking at the Developer Name, which is the same even if the label gets changed. Also, configSettings is only called if no Hierarchy Settings exist. The assumption is that this is done only once, after installation. If our Administrative record type gets deleted, hopefully they will select another one (maybe one they created) in the Account Processor field in settings.

src/classes/UTIL_Describe.cls
ceiroa: Not sure I follow. Are you thinking of record types that are added or removed while the transaction that called this method is still running? It seems like a rare scenario to me. I don't think record types are added or removed very often. And right now the only ones that call these methods are the install script and the settings pages...

src/classes/UTIL_Describe.cls
ceiroa: Why not? Dynamic queries have been supporting namespaces for a while.

src/classes/UTIL_Describe.cls
ceiroa: Because DeveloperName is the one that never changes. Name can be modified.

src/labels/CustomLabels.labels
ceiroa: That's actually the one for the household. I just categorized the three under Account Naming: https://github.com/SalesforceFoundation/HEDAP/blob/feature/216-acc-rec-types-change/src/labels/CustomLabels.labels#L19-L41

src/aura/STG_CMP_System/STG_CMP_System.cmp
kbromer: Nevermind, I think I was misunderstanding how the component is used...

src/aura/STG_CMP_Tabs/STG_CMP_Tabs.cmp
kbromer: Got it, thx

src/classes/ACCT_IndividualAccounts_TDTM.cls
kbromer: Ok, we probably need to doc this diff, something for Varese's backlog

src/classes/ACCT_IndividualAccounts_TDTM.cls
kbromer: A "No Child Opps Only" setting might be good, I'm not sure we'd want the behavior to be different across recordtypes - any good uses cases you're aware of to support that?

src/classes/UTIL_CustomSettingsFacade.cls
kbromer: Though this would bomb if they removed ours prior to hitting setup?

src/classes/UTIL_Describe.cls
kbromer: Ok, amkes sense

src/classes/UTIL_Describe.cls
kbromer: Sure, provided you're passing in the prefixed name i guess... probably ok

src/labels/CustomLabels.labels
kbromer: Got it, thx

src/classes/ACCT_IndividualAccounts_TDTM.cls
ceiroa: Not that I know of, but I was trying to standardize and allow any of this functionality (#219 #220 #221 #222) to be chosen for any of the existing record types.

src/classes/UTIL_CustomSettingsFacade.cls
ceiroa: It shouldn't... The get method will just return null if the record type doesn't exist. It would only blow up if UTIL_Describe.getRecTypesMapByDevName('Account') returns null, which shouldn't happen, afaik.

ceiroa:
@kbromer ~ I've tried addressing any comments. Let me know if there are any others. Happy to make any changes necessary. :-)

PR #230, opened 2016-04-08
Schedule job only if not scheduled.

ceiroa:
Scheduling job only if not scheduled. To avoid tests failing after the first install of the package in an org. Addresses failures we were seeing in the new Bamboo CI system. 

Changes in ADDR_Addresses_TEST.cls are just undoing some minor changes that had been done directly in master branch.


ceiroa:
@daveperry ~ could you please review? Thanks!

PR #233, opened 2016-04-20
Feature/219 accs delete setting

### LINKED ISSUE #219: Use setting to determine which type of accounts to delete if they have no children
Currently we delete them if they are Household or Administrative account and all children have been deleted, but this should be configurable.

### LINKED ISSUE #220: Use setting to determine which type of accounts support seasonal addresses

ceiroa:
# Warning

# Info
- Multi-addresses now supported in any Account record type existing in the org.
- Admin can how choose with type of Accounts will be automatically deleted if all its children Contacts have been deleted.

# Issues
Fixes #219
Fixes #220

### DIFF
--- a/src/aura/STG_CMP_Base/STG_CMP_BaseHelper.js
+++ b/src/aura/STG_CMP_Base/STG_CMP_BaseHelper.js
@@ -1,28 +1,13 @@
 ({
 	addPrefixListSettings : function(settings, namespacePrefix) {
-		if(namespacePrefix && namespacePrefix.length > 0) {
-			//We need to end up with an array of objects, because the settings param is an array of objects
-    		var settings_w_prefix = []; //create array
-    		//Add package prefix to each custom field
-    		for(var key in settings) { //Iterate over each row - key is just a numeric index here
-    			var obj = {}; //create object
-    			for(var key2 in settings[key]) { //Iterate over each field in each row - key2 is the actual field name
-    				if(key2.endsWith('__c')) { 
-	    				var key2_w_prefix = namespacePrefix + key2;
-	    				obj[key2_w_prefix] = settings[key][key2];
-	    			} else {
-	    				obj[key2] = settings[key][key2];
-	    			}
-    			}
-    			settings_w_prefix.push(obj);
-    		}
-    		return settings_w_prefix;
-		} else {
-			return settings;
-		}
+		return this.processPrefixListSettings(settings, namespacePrefix, this.addPrefix);
 	},
 	
 	removePrefixListSettings : function(settings, namespacePrefix) {
+		return this.processPrefixListSettings(settings, namespacePrefix, this.removePrefix);
+	},
+	
+	processPrefixListSettings : function(settings, namespacePrefix, prefixHandlingFunc) {
 		if(namespacePrefix && namespacePrefix.length > 0) {
 			//We need to end up with an array of objects, because the settings param is an array of objects
     		var settings_no_prefix = []; //create array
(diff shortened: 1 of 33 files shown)


ceiroa:
Hi @kbromer ~ There are a lot of changes that went into this pull request, including a new reusable lightning component, and some Javascript refactoring. Do you think you could review it for me? Thanks!

src/aura/STG_CMP_Base/STG_CMP_BaseHelper.js
```diff
@@ -81,6 +56,14 @@
 		}
 	},
 	
+	addPrefix : function(namespacePrefix, key) {
```
davidhabib: I assume the namespace prefix contains the dot in all this code?

src/aura/STG_CMP_System/STG_CMP_System.cmp
```diff
@@ -14,10 +13,10 @@
             <aura:if isTrue="{!v.isView}">
                 <ui:outputText aura:id="recTypeName" value="{!v.accRecTypeName}"/>
             <aura:set attribute="else">
-                <ui:inputSelect aura:id="accRecTypes">
+                <ui:inputSelect aura:id="accRecTypes" change="{!c.onSelectChange}">
                     <aura:iteration items="{!v.accRecTypes}" var="recType">
-                        <ui:inputSelectOption label="{!recType.devName}" text="{!recType.id}" 
-                            value="{!v.accRecTypeName == recType.devName ? true : false}" />
+                        <ui:inputSelectOption label="{!recType.name}" text="{!recType.id}" 
+                            value="{!v.accRecTypeId == recType.id ? true : false}" />
```
davidhabib: why the change from using recordtype Developer names to using recordtype Id's?

src/package.xml
```diff
@@ -214,15 +215,15 @@
...
         <members>Hierarchy_Settings__c.Account_Processor__c</members>
+        <members>Hierarchy_Settings__c.Accounts_to_Delete__c</members>
         <members>Hierarchy_Settings__c.Affl_ProgEnroll_Del_Status__c</members>
         <members>Hierarchy_Settings__c.Affl_ProgEnroll_Del__c</members>
         <members>Hierarchy_Settings__c.Allow_AutoCreated_Duplicates__c</members>
         <members>Hierarchy_Settings__c.Disable_Error_Handling__c</members>
         <members>Hierarchy_Settings__c.Enable_Debug__c</members>
         <members>Hierarchy_Settings__c.Error_Notifications_On__c</members>
         <members>Hierarchy_Settings__c.Error_Notifications_To__c</members>
-        <members>Hierarchy_Settings__c.Household_Adm_Acct_Addresses_Enabled__c</members>
```
davidhabib: can you delete these if they were already packaged?

src/aura/STG_CMP_Base/STG_CMP_BaseHelper.js
ceiroa: Actually, we are manipulating custom settings fields, and the namespace needs to have the double underscore. This is done in https://github.com/SalesforceFoundation/HEDAP/blob/master/src/pages/STG_Settings.page#L29.

src/aura/STG_CMP_System/STG_CMP_System.cmp
ceiroa: Just to standardize storing recordtype IDs in all settings. The change to using dev names was done recently (https://github.com/SalesforceFoundation/HEDAP/pull/226/), as an intermediate step.

src/package.xml
ceiroa: I think you can delete custom settings fields the same you can delete object fields.

ceiroa:
@davidhabib ~ This is ready for another look. I moved those checkboxes we talked about to the right, and since I was on it, I implemented the same pattern for all checkboxes across settings. The others didn't have labels, but this make them all have the slds look also in edit mode.

Issue #218, opened 2016-03-27
The Ability to Customize the Account Naming Setting


judisohn:
Hub thread: https://powerofus.force.com/0D58000002atMIl

Description from the existing item in the internal backlog that has now been closed as duplicate: "Allow users to, for example, add the Contact's student ID to the name of the Administrative Account."

judisohn:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000003Q59HAAS" target="_blank">W-009736</a>

jacebryan:
Closing this Issue as it is not related to an Idea.

jacebryan:
Closed #436 as duplicate . Reopening to continue discussion for PUB issue https://powerofus.force.com/hub-ideas/1158

jacebryan:
Power of US Idea Owner: [Jennifer Sanchez](https://powerofus.force.com/_ui/core/userprofile/UserProfilePage?u=00580000009z5kg&tab=sfdc.ProfilePlatformFeed)
[Power of US Idea](https://powerofus.force.com/hub-ideas/1158)

**Enhancement Request**: Whether the default HEDA account model is set to "Administrative" or "Households" it would be very useful to provide the ability to customize the account naming setting. Similar to NPSP, provide a setting within the HEDA settings to be able to specify the account name format. Here is the NPSP documentation for customizing household names: https://powerofus.force.com/articles/Resource/NPSP-Customize-Household-Names#!#nameformat

**Requirements**:

**Ability to customize the the "Administrative" Account name with the following formats:**

- {!LastName} Administrative Account:  The current option for Administrative Accounts.

- {!{!FirstName}} {!LastName} Administrative Account: Generates an Administrative Account name such as “John Smith Administrative Account”.

- {!LastName} ({!{!FirstName}}) Administrative Account: Generates an Administrative Account name such as “Smith (John) Administrative Account”.

- You can also select other and specify your own naming convention.

**Ability to customize the "Household" Account name with the following formats:**

- none: Does not generate a Household Name, leaving it up to the user to assign one.

- {!LastName} Household:  Generates a Household name such as “Smith Household” or “Smith and Jones Household,” depending on the number of Contacts in the Household.

- {!{!FirstName}} {!LastName} Household: Generates a Household name such as “John Smith Household” or “John and Jane Smith Household”.

[...]

PR #249, opened 2016-05-10
Removing unused method.

### LINKED ISSUE #222: Use setting to determine which Accounts need renaming

ceiroa:
# Warning

# Info

# Issues
Fixes #222.

### DIFF
--- a/src/classes/ACCT_IndividualAccounts_TDTM.cls
+++ b/src/classes/ACCT_IndividualAccounts_TDTM.cls
@@ -449,26 +449,7 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
             // add all accounts needing updating to our list
             dmlWrapper.objectsToUpdate.addAll((List<SObject>)listAccToUpdate);
         }
-    }
- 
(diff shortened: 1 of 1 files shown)


ceiroa:
Hi @kbromer ~ I saw you comment regarding this item in AA. Since I cannot reply there at the moment, I thought we could have the discussion here. 

The thing is that this method, the way it is right now, isn't really useful. For most account types it doesn't make sense to have the names of its contacts as part of the account name. And the method really isn't called from anywhere. There is no logic in the package for naming any type of accounts based on its children. So we were just thinking removing this method, and look into custom naming for all types of accounts when we tackle #218.

If this makes sense, please feel free to merge the request. Let me know otherwise. :-)

ceiroa:
Hey @kbromer, are you ok with this? Is it ok if @daveperry merges it?

kbromer:
All good after Dave reviews, we can always pull it in later-

PR #265, opened 2016-06-06
Creating user without role to avoid test failure if role isn't available in the org.

### LINKED ISSUE #262: Role-related test failures
Parent: #252

From Hub:

The "List has no rows" error was caused by the tests which queried for the 'COO' role on the org. Since my org doesn't have this role, it would return no records and the test would throw an exception. There is an assumption that this role will be present in the org. The workaround: Add the role 'COO' to the org. Suggested code fix: don't assume the 'COO' role is present in the org.

ceiroa:
# Warning

# Info

# Issues
Fixes #262.

### DIFF
--- a/src/classes/UTIL_UnitTestData_TEST.cls
+++ b/src/classes/UTIL_UnitTestData_TEST.cls
@@ -83,7 +83,6 @@ public class UTIL_UnitTestData_TEST {
     public static User CreateNewUserForTests(string strUsername) {
 
         Profile p = [SELECT Id FROM Profile WHERE Name='Standard User'];
-        UserRole r = [SELECT Id FROM UserRole WHERE Name='COO'];
         User u;	
         // to work around mixed DML errors during tests, you must
         // create user account in its own transaction.  got this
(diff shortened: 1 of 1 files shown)


ceiroa:
@kbromer ~ It turns out userRole isn't even a required field when creating a user, so I just removed it. Additional work done in https://github.com/SalesforceFoundation/Cumulus/blob/dev/src/classes/UTIL_UnitTestData_TEST.cls#L283 seems overkill, especially considering the method CreateNewUserForTests is called only from ERR_Handler_TEST.setupWithEmailNotifications. Does this seem fine to you?

ceiroa:
@kbromer ~ Just a followup. Are we good to merge this?

dospromptman:
Carlos, here is some context around why the NPSP code looks the way it does:

I opened this issue when I had NPSP tests failing in one of my dev orgs, because it did not have a 'COO' UserRole:

 https://github.com/SalesforceFoundation/Cumulus/issues/1806

That lead to this work item for NPSP and and its sister item in HEDA:

 https://foundation--agf.na6.visual.force.com/apex/adm_bugdetail?id=a2x800000001g1N
 https://foundation--agf.na6.visual.force.com/apex/adm_bugdetail?id=a2x800000001g9q

The fix to the NPSP version of this issue is in PR (ignore the part about adding the @isTest annotation-- that actually broke things because some methods were being used in production code): 

 https://github.com/SalesforceFoundation/Cumulus/pull/1808

 Note that the test user creation was happening in two locations in NPSP, and this PR refactors test user creation into one location, replacing the other instance with a call to `createNewInactiveUserForTests`, which calls `createNewUserForTests`.

 That is all background information.

 As far as I can tell, if you do not rely on the COO UserRole (or any UserRole) in any of your tests, and it is not required when creating a User, then omitting it when creating test users should be fine.

kbromer:
Omitting should be fine - provided we're not testing managed sharing behavior anywhere (which I don't believe we are since we're not actively doing that in either app)

ceiroa:
Thanks for taking a look, @dospromptman. 

@daveperry, could you please review and merge? Thanks.

PR #274, opened 2016-06-15
Renaming component that was causing push upgrade failures.

### LINKED ISSUE #273: Rename STG_CMP_Header
This component causes many failures when doing push upgrades. Every other component is fine. It forces us to re-run the push a number of times.

ceiroa:
# Warning

# Info

# Issues
Fixes #273.


davidhabib:
@ceiroa why was the previous name causing push upgrade failures?

ceiroa:
We don't know why this component was causing push upgrades failures. Core couldn't tell us either. This is just an attempt to make a change that just "might" help with the issue. If the component has a different name maybe it's successfully propagated to all pods. That assumes that the problem was that the component wasn't available in some pods. 

Since you are here, @davidhabib , could you please review and merge?

PR #299, opened 2016-08-24
Feature/267 tdtm custom objects

### LINKED ISSUE #267: Support custom objects in TDTM framework
Currently, HEDA includes triggers for a number of objects. In this case users that want to add TDTM classes to the framework don't need to create triggers, just the classes. However, users may want to create TDTM classes for objects for which we don't provide triggers, of for custom objects.

### LINKED ISSUE #272: Expose the TDTM default config as a global method
We do this with our global API in NPSP. We need to do the same so that users writing unit tests can integrated them with our code. The team from UW-Extension just asked about this.

Thread where we talked about how to do this in NPSP:
https://powerofus.force.com/_ui/core/chatter/groups/GroupProfilePage?g=0F980000000CjRe&fId=0D58000001xrlVh&s1oid=00D00000000hWLM&OpenCommentForEdit=1&s1nid=0DB800000004Fa7&emkind=chatterPostNotification&s1uid=00580000005vESn&emtm=1426798147229&fromEmail=1&s1ext=0

ceiroa:
# Critical Changes

# Changes
- **TDTM_Global_API class added a number of global methods and a global object**.

    - The `TDTM_Global_API.TdtmToken` object is an abstraction of the HEDA Trigger_Handler__c SObject and should be used instead of directly working with the Trigger_Handler__c SObject within tests.

    - The `getTdtmConfig` and `setTdtmConfig` methods allow code inside Salesforce to get the default TDTM configuration and set the TDTM configuration in their tests. This lets any user write tests that exercise all or part of the HEDA code together with their own TDTM code. The `getTdtmConfig` method returns a list of `TdtmToken` objects. The `setTdtmConfig` method accepts a list of `TdtmToken` objects.

    - The `run` method lets users create triggers for their own custom objects or for standard objects for which HEDA doesn't have a trigger. 

# Issues Closed
Fixes #267.
Fixes #272.


src/classes/ADDR_HouseholdAcc_TEST.cls
```diff
@@ -1698,6 +1698,7 @@ public with sharing class ADDR_HouseholdAcc_TEST {
         // insert new contacts to the HH without any addresses
         List<Contact> listCon = UTIL_UnitTestData_TEST.getMultipleTestContacts(2);
         for (Integer i = 0; i < 2; i++) {
+            listCon[i].LastName = 'ToAvoidDupeRule' + i;
```
ceiroa: Changes in this class are due to Duplicate Rules errors. Don't really have anything to do with TDTM...

src/classes/ADDR_Seasonal_UTIL.cls
```diff
@@ -256,6 +256,7 @@ public with sharing class ADDR_Seasonal_UTIL {
 
         TDTM_Runnable.Dmlwrapper dmlWrapper = new TDTM_Runnable.Dmlwrapper();
         ADDR_Addresses_TDTM.refreshParentAndSiblings(mapAccIdAddr, dmlWrapper, null);
-        TDTM_TriggerHandler.processDML(dmlWrapper);
+        Savepoint sp = Database.setSavepoint();
```
ceiroa: I'm attempting to simplify TDTM_TrigggerHandler. One of the things I'm doing is passing the savepoint as a parameter (I'm also removing the other savepoint, and instatiating TDTM_ObjectDataGateway inside the run method instead of passing it as a parameter every time - it was being passed as a parameter only because the original design had to alternatives for storing the TDTM config, one in custom setting and one in custom object).

src/classes/TDTM_Global_API.cls
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+* @author Salesforce.org
+* @date 2016
+* @group TDTM
+* @group-content ../../ApexDocContent/API.htm
+* @description Provides methods to call the Trigger Handler class, which in turns calls the classes that need to
+* run for each DML action on a set of records, and to retrieve the default HEDA TDTM configuration.
+*/
+global class TDTM_Global_API {
```
ceiroa: After some thinking, I believe having a separate class only for the TDTM-integration related methods is the most reasonable option. Additional global classes can be later created, if necessary.

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -197,20 +196,7 @@ public class TDTM_TriggerHandler {
...
-        processDMLWithRollback(dmlWrapper, true);
-    }
-    /*******************************************************************************************************
-    * @description Performs pending DML operations, on the records stored in the DMLWrapper that it gets as parameter.
-    * We want this method to be static so that it can be called from the future method in TDTM_Runnable.
-    * @param dmlWrapper The class that contains the records that need to have DML performed on.
-    * @param withRollback True if errors should cause rollback, false if not.
-    * @return Errors Object with list of Error objects
-    */
-    public static ERR_Handler.Errors processDMLWithRollback(TDTM_Runnable.DmlWrapper dmlWrapper, boolean withRollback) {
```
ceiroa: This method is being removed. It was originally created in NPSP just so the Household Addresses code could call it, with the second parameter set to false. Now that code is doing its own DML and error processing.

src/triggers/TDTM_Account.trigger
```diff
@@ -30,8 +30,6 @@
 trigger TDTM_Account on Account (after delete, after insert, after undelete, 
 after update, before delete, before insert, before update) {
 
-    TDTM_TriggerHandler handler = new TDTM_TriggerHandler();  
-    handler.run(Trigger.isBefore, Trigger.isAfter, Trigger.isInsert, Trigger.isUpdate, Trigger.isDelete, 
-        Trigger.isUnDelete, Trigger.new, Trigger.old, Schema.Sobjecttype.Account, 
-        new TDTM_ObjectDataGateway());
+    TDTM_Global_API.run(Trigger.isBefore, Trigger.isAfter, Trigger.isInsert, Trigger.isUpdate, Trigger.isDelete,
```
ceiroa: We might as well "drink our own champagne", and go through the TDTM_Global_API class. It will also make more sense for those that look at the codebase.

src/classes/ADDR_Addresses_TDTM.cls
```diff
@@ -225,7 +225,20 @@ public class ADDR_Addresses_TDTM extends TDTM_Runnable {
     private void processDml(DmlWrapper dmlWrapper, Map<Id, Address__c> mapParentIdAddr) {
         // we must process dmlWrapper ourselves, so we can extract out the appropriate errors that might occur updating contacts 
         // and parents, and associating those errors with the correct addresses.
-        ERR_Handler.Errors errors = TDTM_TriggerHandler.processDMLWithRollback(dmlWrapper, false);
```
ceiroa: This code does its own DML work, so we can remove the processDMLWithRollback method.

src/classes/TDTM_TriggerHandler.cls
```diff
@@ -197,20 +196,7 @@ public class TDTM_TriggerHandler {
...
-        //Create transaction
-        Savepoint sp = Database.setSavepoint();
```
ceiroa: Saving one DML operation (setting the savepoint), until we do the error handling re-work.

src/classes/TDTM_Global_API.cls
ceiroa: Any suggestions on how to name the class in case later we want to add methods that also work as REST endpoints to it? @kbromer @paulmalenke (See https://docs.google.com/document/d/1tUI0yGjhkrd2_1ELQlwEe3Gio0akLvNoYjB8kpEqAgQ/ for additional info.)

kbromer: I think that sounds right @ceiroa, especially since its unlikely we'll want/need to expose TDTM externally anyway (sorta defeats the point in some ways...)

src/classes/TDTM_Global_API.cls
```diff
@@ -0,0 +1,74 @@
...
+    
+     /*******************************************************************************************************
+    * @description Returns the default TDTM configuration. 
+    * @return List<Trigger_Handler__c> a list of trigger handler records with the default values. The records
+    *         have not yet been inserted in the database. You can manipulate them, if necessary, and insert 
+    *         them. If no Trigger_Handler__c records exist, the application will insert the defaults when any 
+    *         of the triggers run. 
+    */
+    global static List<Trigger_Handler__c> getDefaultTdtmConfig() {
+
```
ceiroa: @dospromptman raised a good point here, that we probably shouldn't be returning a specific SObject type. Otherwise are forever tied to it. What if later we want to store the configuration in a custom setting, for example? The challenge is that if we instead return a class we need to make that class and the methods in it global. We need to think about potential alternatives. @kbromer @paulmalenke

ApexDocContent/TDTM.htm
```diff
@@ -7,10 +7,8 @@
 	    <ul>
 	        <li>TDTM_DefaultConfig - stores the default TDTM configuration.</li>
 	        <li>TDTM_Filter - Allows running TDTM class only for those records that do not meet the filtering criteria.</li>
-	        <li>TDTM_ObjectDataGateway - class that queries custom object to retrieve classes to call for creation or modification
```
ceiroa: TDTM_ObjectDataGateway has been renamed TDTM_Config.

src/classes/STG_InstallScript.cls
```diff
@@ -42,29 +42,23 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
-    	 } else if(context.isUpgrade() || context.isPush()) {
-    	 	List<Trigger_Handler__c> defaultHandlers = TDTM_DefaultConfig.getDefaultRecords(); 
-            insertTdtmDefaults(defaultHandlers);
-    	 }
+    	    //Insert default TDTM config (not necessary, but better to be explicit).
+    	    List<TDTM_Global_API.TdtmToken> tokens = TDTM_Global_API.getTdtmConfig();
+            TDTM_Global_API.setTdtmConfig(tokens); 
     	 
-    	 //Replace the value in the "Account Processor" field with the ID of the record type that was there
-    	 if(context.previousVersion() != null && context.previousVersion().compareTo(new Version(1,7)) == 0) {
```
ceiroa: This chunk of code was version-specific and is not needed any more.

src/classes/STG_InstallScript.cls
```diff
@@ -130,40 +124,61 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+                                                       Filter_Value__c, User_Managed__c from Trigger_Handler__c where Class__c != null];
+        List<TDTM_Global_API.TdtmToken> existingTdtmConfig = new List<TDTM_Global_API.TdtmToken>();
+        for(Trigger_Handler__c handler : existingHandlers) {
+            existingTdtmConfig.add(new TDTM_Global_API.TdtmToken(handler.Class__c, handler.Object__c, handler.Trigger_Action__c, handler.Active__c, 
+                    Integer.valueOf(handler.Load_Order__c), handler.Asynchronous__c, handler.Filter_Field__c, handler.Filter_Value__c, handler.User_Managed__c));
+        }
 
-        List<Trigger_Handler__c> handlersToInsert = new List<Trigger_Handler__c>();
-        List<Trigger_Handler__c> handlersToUpdate = new List<Trigger_Handler__c>();
+        List<TDTM_Global_API.TdtmToken> newTdtmConfig = new List<TDTM_Global_API.TdtmToken>();
```
ceiroa: We are using tokens now, and replacing the entire configuration.

ceiroa:
@paulmalenke & @dospromptman 

This pull request is ready for review. Ideally, I'd spend an additional day making everything looks good. Since that's not going to be possible, I'd ask you to both review the pull request. If you feel strongly there are changes that need to be made, one of you could take ownership of the branch. Or you can wait until I get back, if that's not a huge blocker for Paul. :-)

These are the main goals of this branch:

- wrap database representation of trigger configuration in objects
- write global methods that expose trigger configuration and allow third-parties to create their own triggers that plug into TDTM (for objects that we don't have triggers for, including custom)
- keep using the trigger config cache. don't read from database when we don't need to
- make sure we only write trigger config to database when necessary
- make it easier for tests to insert a custom TDTM configuration, which could also be blank
- make sure post install scripts continues to properly update tdtm configuration

ceiroa:
@jlantz The feature job failed: https://salesforce-org.atlassian.net/builds/browse/HEDA-FB25-15/log But it doesn't seem related to my changes. It looks related to the changes @paulmalenke did around Courses...

ceiroa:
Ready to merge, @paulmalenke. Let me know if you have any additional questions.

src/classes/TDTM_Config.cls
```diff
@@ -0,0 +1,153 @@
...
+    /*******************************************************************************************************
+    * @description Sets the trigger handlers configuration with the data passed as parameter.
+    * @parameter tdtmTokens A list with the defaul trigger handlers configuration.
+    * @return void
+    *********************************************************************************************************/
+    public static void setTdtmConfig(List<TDTM_Global_API.TdtmToken> tdtmTokens) {
+        tdtmTokensCache = tdtmTokens;
+        
+        //clear existing TDTM configuration
+        delete [select Id from Trigger_Handler__c limit 50000];
```
ceiroa: @davidjray @xinnanlisf I think this is where the TDTM push bug that clears custom TDTM entries was introduced. Back in October...

ceiroa: Actually, upon taking a second look I don't think this is wrong. The post-install script is first retrieving all records that exist in the database. So it's a bit heavy-handed, but not clearly incorrect. We'll need to investigate further...

PR #304, opened 2016-08-30
Adding prefix to custom object when trying to get the record types.

paulmalenke:


paulmalenke:
@ceiroa This is good to merge with your review.

PR #316, opened 2016-09-13
Do Not Delete Unnecessary Affiliations and Course Enrollments.

### LINKED ISSUE #292: COFF_Affiliation_TDTM Deletes Unnecessary Affiliations and Enrollments on Removal of Faculty
When a Contact is set as Faculty on more than one Course Offering and the Faculty is cleared from one of the Course Offerings, the Faculty is removed as a Course Enrollment from all course offerings that have that Contact as Faculty. Even though the Course Enrollments are removed, that Contact still populates the Faculty field on all of the other Course Offerings. In the case that the original Faculty that created the Affiliation to the parent department is removed, the Affiliation is also removed even if there are other Course Offerings with that Faculty within that department that still warrant an Affiliation.

enrollafflbug

_note: screencast shows Course Enrollments as Course Connections_

paulmalenke:
# Critical Changes

# Changes

# Issues Closed
Fixes #292

### DIFF
--- a/src/classes/COFF_Affiliation_TDTM.cls
+++ b/src/classes/COFF_Affiliation_TDTM.cls
@@ -168,7 +168,7 @@ public class COFF_Affiliation_TDTM extends TDTM_Runnable {
                 for (integer i=0; i<oldlist.size(); i++) {
                     Course_Offering__c oldOff = (Course_Offering__c)oldlist[i];
 
-                    if(triggerAction == TDTM_Runnable.Action.AfterDelete) {
+                    if(triggerAction == TDTM_Runnable.Action.BeforeDelete) {
 
                         //Course Offering has been deleted
                         //     --> delete related Affiliation, if the faculty field had a value
@@ -176,6 +176,7 @@ public class COFF_Affiliation_TDTM extends TDTM_Runnable {
                         if(oldOff.Faculty__c != null) {
                             UTIL_Debug.debug('****Course Offering with faculty has been deleted');
                             facultyCleanupIDs.add(oldOff.Faculty__c);
+                            offsWithRemovedFaculty.add(oldOff.Id);
                         }
                     }
                 }
@@ -186,7 +187,7 @@ public class COFF_Affiliation_TDTM extends TDTM_Runnable {
                 if (UTIL_CustomSettingsFacade.courseConnectionsEnabled()) {
                     updateConnectionsOnFacultyAdd(offsWithNewFaculty, objectsToInsert, objectsToUpdate);
                 } else {
-                    createAffls(offsWithNewFaculty, newFacultyIDs, courseEnrllsToInsert);
+                    createOrLinkAffls(offsWithNewFaculty, newFacultyIDs, courseEnrllsToInsert);
                 }
                 dmlWrapper.objectsToInsert.addAll((List<SObject>)courseEnrllsToInsert);
             }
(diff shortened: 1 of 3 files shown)


paulmalenke:
@ceiroa This is ready for review and merge once the code freeze is over.

src/classes/COFF_Affiliation_TDTM.cls
```diff
@@ -471,20 +489,20 @@ public class COFF_Affiliation_TDTM extends TDTM_Runnable {
     }
 
    /*******************************************************************************************************
-    * @description Determines whether an Affiliation to the Department the Course Offering belongs to already
-    * exists for each faculty member. Public only so it can be called from the tests.
+    * @description Determines whether an Affiliation to the Account the Course Offering belongs to already
```
ceiroa: I won't hold the pull request for this, but just fyi, the term "Department" was being used on purpose for clarity, as the lookup from Course to Account is called "Department".

src/classes/COFF_Affiliation_TDTM.cls
```diff
@@ -471,20 +489,20 @@ public class COFF_Affiliation_TDTM extends TDTM_Runnable {
...
+    * exists for each faculty member. Will return null if no Affiliation exists.
+    * @return Affiliation__c Affiliation to the Account or null
     ********************************************************************************************************/
-    public Boolean afflToDptExists(Course_Offering__c offering, Map<ID, List<Affiliation__c>> facultyIDtoAffls) {
+    @testVisible
+    private Affiliation__c affiliationToAccount(Course_Offering__c offering, Map<ID, List<Affiliation__c>> facultyIDtoAffls) {
         List<Affiliation__c> facultyMemberAffls = facultyIDtoAffls.get(offering.Faculty__c);
```
ceiroa: It's not clear to me why change the return value from Boolean to Affiliation__c. It doesn't seem like anything is done with the returned affiliation in any of the places this method is called.

src/classes/COFF_Affiliation_TDTM.cls
```diff
@@ -168,14 +168,15 @@ public class COFF_Affiliation_TDTM extends TDTM_Runnable {
...
-                    if(triggerAction == TDTM_Runnable.Action.AfterDelete) {
+                    if(triggerAction == TDTM_Runnable.Action.BeforeDelete) {
 
                         //Course Offering has been deleted
                         //     --> delete related Affiliation, if the faculty field had a value
                         //     --> delete the Course Enrollment, if the faculty field had a value
                         if(oldOff.Faculty__c != null) {
                             UTIL_Debug.debug('****Course Offering with faculty has been deleted');
                             facultyCleanupIDs.add(oldOff.Faculty__c);
+                            offsWithRemovedFaculty.add(oldOff.Id);
```
ceiroa: Just to confirm, we are changing from AfterDelete to BeforeDelete because the record's ID field is null in AfterDelete. Correct?

src/classes/COFF_Affiliation_TDTM.cls
paulmalenke: Line 249 uses the returned affiliation to attach it to an Enrollment.

src/classes/COFF_Affiliation_TDTM.cls
paulmalenke: I kept the variable name when the method is used the same but changed the method to make it more generic.

src/classes/COFF_Affiliation_TDTM.cls
paulmalenke: That is correct.

src/classes/COFF_Affiliation_TDTM.cls
ceiroa: I see it now. Thanks for the clarification.

src/classes/COFF_Affiliation_TDTM.cls
ceiroa: Clarification: ID field is not null here, but querying for these IDs won't return the expected records in after delete.

PR #325, opened 2016-10-13
Winter '17 SLDS Fix

### LINKED ISSUE #321: Settings Account Record Types Checkbox Array Broken in Winter 17
The Accounts and Contacts settings page displays two separate arrays of Account Record Types and the alignment of those components is broken in Winter 17. Attaching screenshot example.
heda account checkboxes view

paulmalenke:
# Critical Changes

# Changes

# Issues Closed
Fixes #321


davidhabib:
I know we moved NPSP to use private scoped SLDS not too long ago, but I don't recall us putting in the extends="ltng:outAppUnstyled" tag.  And I seem to recall some recent issues have been making us reconsider our decision in NPSP to not use the system supplied SLDS.  @jjbennett do you recall where we stand on this?

paulmalenke:
@davidhabib The overall plan is to use the provided SLDS but HEDA Settings is still back on a 0.8 version of SLDS. The `ltng:outAppUnstyled` is to opt-out of the provided SLDS version that Winter '17 provides automatically. When we get more time I would like to convert to use provided SLDS but everything breaks when we upgrade so opt-out is the best option right now.

jjbennett:
@davidhabib we will be removing the scoped slds and using the built in slds after Winter 17 has been pushed to all orgs. Our other option would have been to opt out but opting in also allows us to easily use some of the new lightning base components.

davidhabib:
thanks @paulmalenke and @jjbennett.  Those plans both sound good!

paulmalenke:
Winter '17 is live so this is good to go. Build now passes.

PR #331, opened 2016-10-19
check for null

### LINKED ISSUE #329: AFFL_AccRecordType_TDTM not checking for null
If an Account is changed from a null record type to a non-null record type and doesn't have a primary affiliation mapping for that record type, keyAfflFieldNewType (line 77) is equal to null. This causes an exception when it is added to the query on line 155. This can be fixed by changing line 87 to "} else if (keyAfflFieldNewType != null) {".

kyleschmid:
# Critical Changes

# Changes

# Issues Closed
Fixes #329.


ceiroa:
Thank you @kyleschmid!

PR #334, opened 2016-10-27
Catching dml exception when trying to delete record handlers, in case…

### LINKED ISSUE #333: ENTITY_IS_DELETED error on Informatica integration
From https://powerofus.force.com/_ui/core/userprofile/UserProfilePage?u=0058000000BMwyH&tab=sfdc.ProfilePlatformFeed&fId=0D580000032FgmD.

After version 1.19 was pushed to production on 27/10/16, user reported getting the following error from an existing Informatica integration that was previously working correctly:

Error Type: System.DmlException
Error Date: 2016-10-27 11:50:48
Message: "Delete failed. First exception on row 0 with id a2R4E0000001L5UUAU; first error: ENTITY_IS_DELETED, entity is deleted: []"
 
Stack Trace:
Stacktrace available at the time of saving the record - Class.hed.TDTM_Config.setTdtmConfig: line 82, column 1
Class.hed.TDTM_Config.getClassesToCallForObject: line 124, column 1
Class.hed.TDTM_TriggerHandler.run: line 75, column 1
Class.hed.TDTM_Global_API.run: line 60, column 1
Trigger.hed.TDTM_Account: line 33, column 1

### LINKED ISSUE #335: Non-default trigger handlers being deleted
In version 1.19, anytime you create a trigger handler and then perform DML on an object that calls a trigger, that trigger handler is deleted because it is not one of the default trigger handlers in TDTM_DefaultConfig.getDefaultRecords.

ceiroa:
… there's race condition.

# Critical Changes

# Changes

# Issues Closed
Fixes #333.
Fixes #335.

### DIFF
--- a/src/classes/TDTM_Config.cls
+++ b/src/classes/TDTM_Config.cls
@@ -79,7 +79,11 @@ public class TDTM_Config {
         tdtmTokensCache = tdtmTokens;
         
         //clear existing TDTM configuration
-        delete [select Id from Trigger_Handler__c limit 50000];
+        try {
+            delete [select Id from Trigger_Handler__c limit 50000];
+        } catch(System.DmlException e) {
+            UTIL_Debug.debug('***Could not delete default config');
+        }
         
         //Persisting new configuration to database
         List<Trigger_Handler__c> handlers = new List<Trigger_Handler__c>();
--- a/src/classes/TDTM_Config.cls
+++ b/src/classes/TDTM_Config.cls
@@ -60,16 +60,13 @@ public class TDTM_Config {
             return tdtmTokensCache;
         }
         List<TDTM_Global_API.TdtmToken> tokens = new List<TDTM_Global_API.TdtmToken>();
-        List<Trigger_Handler__c> tdtmConfig;
-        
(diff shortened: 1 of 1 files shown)


ceiroa:
@paulmalenke @kbromer @dospromptman 
What do you guys think? I have verified it successfully fixes the issue (I was able to duplicate the problem with the script in the issue's comments). Now one of the requests gets and catches the exception, and the other one doesn't. This can be seen in the Debug Console logs.

ceiroa:
Also added fix for #335. I don't think we can create a unit test for it though, as we would need pre-existing data in the trigger handler table (and we don't want to make the test seealldata=true).

ceiroa:
@kbromer ~ I made some more changes this morning that I'm really happy with. 

Catching the exception when two external calls were coming in simultaneously was not the ideal approach. We are now instead setting the cache every time getTdtmConfig is called. 

The Trigger_Handler__c table is cleared only when setTdtmConfig is explicitly called. This won't happen every time TDTM_TriggerHandler is called.

@paulmalenke and I reviewed this, and we think it's good to go.

PR #340, opened 2016-11-04
Add macOS specific ignores.

paulmalenke:
# Critical Changes

# Changes

# Issues Closed


paulmalenke:
All of these should be in a global gitignore but adding them just in case.

PR #341, opened 2016-11-04
Referencing BillingCountryCode and MailingCountryCode was causing com…

ceiroa:
…pilation errors in orgs with StateCountryPicklist disabled.

Also, one of the tests was checking a value in Contact when it should be checking a value in Account.

We noticed this when the master build failed (https://salesforce-org.atlassian.net/builds/browse/HEDA-MAS-83/log). Just fyi, @kyleschmid.

### DIFF
--- a/src/classes/ADDR_Contact_TEST.cls
+++ b/src/classes/ADDR_Contact_TEST.cls
@@ -90,8 +90,8 @@ public with sharing class ADDR_Contact_TEST {
         System.assertEquals(null, acc.BillingStreet);
         System.assertEquals(null, acc.BillingState);
         System.assertEquals(null, acc.BillingPostalCode);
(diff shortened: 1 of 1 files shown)


ceiroa:
@paulmalenke We'll need to merge this before the one you opened with ignore changes, as that branch fails due to the issue being fixed here.

kyleschmid:
Ah. I figured the short-circuiting when checking to see if picklists were enabled would take care of any runtime errors, but compilation errors are a different story.

kyleschmid:
It looks like this change will still cause a test failure in orgs with a default country set (we have ours set to United States). Might be easier to take this assert out altogether.

kyleschmid:
Never mind. Ignore my last comment. If state/country picklists aren't enabled then there won't be a default country to begin with...

PR #346, opened 2016-11-30
Setting up CumlusCI2

davidjray:
# Critical Changes

# Changes
Added cumulusci.yml for CumulusCI 2

# Issues Closed

PR #347, opened 2016-11-30
Improve Address Error Messages

### LINKED ISSUE #343: Update multi-addresses error message
When a user tries to create an Address record from the Addresses related list in Contact or Account, and multi-addresses is not enabled for Contacts or for that Account record type, respectively, an error is displayed in the UI. The text of this error should be updated to: [Docs to provide text]

davidjray:
# Critical Changes

# Changes
Separate error messages for Account and Contact when trying to create a child address without Multi-Addresses enabled.

# Issues Closed
Fixes #343.

### DIFF
--- a/src/classes/ADDR_Addresses_TDTM.cls
+++ b/src/classes/ADDR_Addresses_TDTM.cls
@@ -1,10 +1,10 @@
 /*
     Copyright (c) 2014, Salesforce.org
     All rights reserved.
-    
+
     Redistribution and use in source and binary forms, with or without
     modification, are permitted provided that the following conditions are met:
-    
+
     * Redistributions of source code must retain the above copyright
       notice, this list of conditions and the following disclaimer.
     * Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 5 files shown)


davidjray:
This pull request also includes the CumulusCI 2 yml file because I needed it for this work as well. That makes #346 redundant.

src/labels/CustomLabels.labels
```diff
@@ -134,7 +134,15 @@
...
         <protected>true</protected>
         <shortDescription>addrNotEnabled</shortDescription>
-        <value>Address management not enabled for this object type in settings.</value>
+        <value>The ability to add multiple addresses is disabled for this Account record type. To enable it, go to HEDA Settings, click on Accounts and Contacts, and select the appropriate checkbox next to Account Types with Multi-Addresses Enabled.</value>
+    </labels>
+    <labels>
+        <fullName>addrNotEnabledContact</fullName>
+        <categories>Address Management</categories>
+        <language>en_US</language>
+        <protected>true</protected>
```
paulmalenke: @ceiroa I cannot remember if we said we would move forward with true or false here. Did we document that somewhere?

src/package.xml
```diff
@@ -318,6 +318,7 @@
         <members>RelationshipsLookupDescription</members>
         <members>addrHhAdmAccountOnly</members>
         <members>addrNotEnabled</members>
+        <members>addrNotEnabledContact</members>
```
paulmalenke: I believe this is the only line that should be committed for package.xml. The extra lines are not needed.

src/classes/ADDR_Addresses_TDTM.cls
```diff
@@ -1,10 +1,10 @@
 /*
     Copyright (c) 2014, Salesforce.org
     All rights reserved.
-
```
paulmalenke: Loving all the whitespace deletion!

src/labels/CustomLabels.labels
ceiroa: They should be protected unless there's a good reason for them not to be (if we wanted users to be able to change them). The reason why it's better for us to make them protected is that protected labels can be deleted from the package if we don't need them any more.

ceiroa:
I might be missing something, but I don't see where the new error message is being used...

paulmalenke:
@ceiroa You might have to "Load diff" for src/classes/ADDR_Addresses_TDTM.cls. The diff is large so it does not show it unless you click the link in the Files changed.

ceiroa:
@paulmalenke You are right, it wasn't loading it. But now that I see it I still think it's only using one of the error labels...

Issue #270, opened 2016-06-14
Have settings page check the profile permissions on save


ceiroa:
So only sysadmins can save changes to the settings page. Non sysadmins shouldn't even have access to the tab, but we should double-check that the user's profile has the permissions, just in case sysadmin forgot to hide tab from all users.

ceiroa:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000003qFMdAAM" target="_blank">W-010654</a>

kporray1:
This issue was fixed in Jan 2017 and will be closed.

PR #375, opened 2017-01-19
Check User Profile permissions before loading settings page

davidjray:
# Critical Changes

# Changes
- Only allows User Profiles with the "Customize Application" setting to access the HEDA settings page.

# Issues Closed
- Implements #270

### DIFF
--- a/src/classes/UTIL_CustomSettingsFacade.cls
+++ b/src/classes/UTIL_CustomSettingsFacade.cls
@@ -52,7 +52,7 @@ public without sharing class UTIL_CustomSettingsFacade {
     static List<Relationship_Lookup__c> relationshipLookup;
 
     /*******************************************************************************************************
-    * @description Constructor checks 
+    * @description Constructor, sets boolean value to determine is User can access Settings page 
     * 
     */    
     public UTIL_CustomSettingsFacade() {
--- a/src/classes/UTIL_CustomSettingsFacade.cls
+++ b/src/classes/UTIL_CustomSettingsFacade.cls
@@ -56,9 +56,8 @@ public without sharing class UTIL_CustomSettingsFacade {
     * 
     */    
     public UTIL_CustomSettingsFacade() {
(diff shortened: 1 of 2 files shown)


davidjray:
image

src/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -50,6 +52,16 @@ public without sharing class UTIL_CustomSettingsFacade {
     static List<Relationship_Lookup__c> relationshipLookup;
 
     /*******************************************************************************************************
+    * @description Constructor, sets boolean value to determine is User can access Settings page 
+    * 
+    */    
+    public UTIL_CustomSettingsFacade() {
+        User user = [SELECT ProfileId FROM User where Id = :UserInfo.getUserId()]; 
+        Profile profile = [SELECT PermissionsCustomizeApplication FROM Profile Where Id = :user.ProfileId];
```
ceiroa: Can we combine these two queries into one?

ceiroa: Also, shouldn't this be the pull request in which we create a separate class to be the component's controller and move those methods that don't belong out of UTIL_CustomSettingsFacade?

src/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -37,6 +37,8 @@
 */
 public without sharing class UTIL_CustomSettingsFacade {
 
+    public Boolean hasPerm {get;set;}
```
ceiroa: Does this need to be public for the component to be able to read it? If not, better make it private.

davidjray: It does need to be public since the VF page uses it.

ceiroa:
@davidjray Also ready to be merged?

davidjray:
@ceiroa, yes ready to be merged.

PR #386, opened 2017-01-31
Seasonal Address Test Updates

davidjray:
# Critical Changes

# Changes
- Seasonal Address test adjustments
# Issues Closed

### DIFF
--- a/src/classes/ADDR_Seasonal_TEST.cls
+++ b/src/classes/ADDR_Seasonal_TEST.cls
@@ -744,28 +744,32 @@ private with sharing class ADDR_Seasonal_TEST {
         UTIL_UnitTestData_TEST.ContactsWithAddrs contactsAddrs = UTIL_UnitTestData_TEST.createTestContactsAddrs(2);
         List<Address__c> conAddrs = contactsAddrs.addrs;
 
-        // create additional Account addresses, with seasonal information to expire today
+        // create additional Account addresses, with seasonal information to expire tomorrow
         List<Address__c> accAddrs2 = UTIL_UnitTestData_TEST.getMultipleTestAddresses(2);
         for (Integer i = 0; i < 2; i++) {
             accAddrs2[i].Parent_Account__c = accsAddrs.accs[i].Id;
             accAddrs2[i].Default_Address__c = false;
-            accAddrs2[i].Seasonal_Start_Month__c = String.valueOf(system.today().addMonths(-1).month());
-            accAddrs2[i].Seasonal_Start_Day__c = '28';
-            accAddrs2[i].Seasonal_End_Month__c = String.valueOf(system.today().month());
+            accAddrs2[i].Seasonal_Start_Month__c = String.valueOf(system.today().addDays(-31).month());
+            accAddrs2[i].Seasonal_Start_Day__c = String.valueOf(system.today().addDays(-31).day());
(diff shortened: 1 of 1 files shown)


src/classes/ADDR_Seasonal_TEST.cls
```diff
@@ -744,28 +744,32 @@ private with sharing class ADDR_Seasonal_TEST {
...
             accAddrs2[i].Default_Address__c = false;
-            accAddrs2[i].Seasonal_Start_Month__c = String.valueOf(system.today().addMonths(-1).month());
-            accAddrs2[i].Seasonal_Start_Day__c = '28';
-            accAddrs2[i].Seasonal_End_Month__c = String.valueOf(system.today().month());
+            accAddrs2[i].Seasonal_Start_Month__c = String.valueOf(system.today().addDays(-31).month());
+            accAddrs2[i].Seasonal_Start_Day__c = String.valueOf(system.today().addDays(-31).day());
+            accAddrs2[i].Seasonal_Start_Year__c = system.today().addDays(-31).year();
+            accAddrs2[i].Seasonal_End_Month__c = String.valueOf(system.today().addDays(1).month());
             accAddrs2[i].Seasonal_End_Day__c = String.valueOf(system.today().addDays(1).day());
+            accAddrs2[i].Seasonal_End_Year__c = system.today().addDays(1).year();
```
ceiroa: We should not use Seasonal_Start_Year__c and Seasonal_End_Year__c, as this changes the intent of the test. We have other tests that include the year, but those are for addresses that aren't seasonal (in the sense that they don't repeat every year).

src/classes/ADDR_Seasonal_TEST.cls
```diff
@@ -744,28 +744,32 @@ private with sharing class ADDR_Seasonal_TEST {
...
             conAddrs2[i].Default_Address__c = false;
-            conAddrs2[i].Seasonal_Start_Month__c = String.valueOf(system.today().addMonths(-1).month());
-            conAddrs2[i].Seasonal_Start_Day__c = '28';
-            conAddrs2[i].Seasonal_End_Month__c = String.valueOf(system.today().month());
+            conAddrs2[i].Seasonal_Start_Month__c = String.valueOf(system.today().addDays(-31).month());
+            conAddrs2[i].Seasonal_Start_Day__c = String.valueOf(system.today().addDays(-31).day());
+            conAddrs2[i].Seasonal_Start_Year__c = system.today().addDays(-31).year();
+            conAddrs2[i].Seasonal_End_Month__c = String.valueOf(system.today().addDays(1).month());
             conAddrs2[i].Seasonal_End_Day__c = String.valueOf(system.today().addDays(1).day());
+            conAddrs2[i].Seasonal_End_Year__c = system.today().addDays(1).year();
```
ceiroa: We should not use Seasonal_Start_Year__c and Seasonal_End_Year__c, as this changes the intent of the test. We have other tests that include the year, but those are for addresses that aren't seasonal (in the sense that they don't repeat every year).

src/classes/ADDR_Seasonal_TEST.cls
davidjray: Good call. I wanted to make sure the year was not part of the problem.

src/classes/ADDR_Seasonal_TEST.cls
davidjray: Good call. I wanted to make sure the year was not part of the problem.

PR #395, opened 2017-03-06
Clean Code

davidjray:
removed System.debug call

# Critical Changes

# Changes

# Issues Closed


davidjray:
Didn't create this as a feature branch, it is a single line being removed that should not have been there in the first place.

PR #396, opened 2017-03-07
Feature/email enhancements

### LINKED ISSUE #344: Investigate adding Personal Email field and Personal option in Preferred drop-down

### LINKED ISSUE #371: Custom created email fields do not populate the standard email field once set as preferred in HEDA
Currently after creating a custom email field in HEDA, when you manually add the custom email field to the preferred picklist and select the custom email field as preferred, the standard email field is not updated. 

We need the behavior to be when you select any email from the preferred picklist (including custom), that the email will be copied into the standard email field.

### LINKED ISSUE #372: In HEDA, the standard email field can be overwritten with a null value if the preferred email setting points to an empty email field
The "Email" field should never be overwritten with a null value. If there is a value in the "Email" field, but there is no address, say, in the "Work Email" field, then, if the "Preferred Email" dropdown is set to "Work" we should not copy the null value from "Work Email" to "Email" (as HEDA currently does); rather, we should display an error "Work Email is null and therefore cannot be preferred"

### LINKED ISSUE #373: Preferred email setting should be required if any email address exists for the contact
In HEDA today, if you add an email to the contact, but don't specify any preferred email, then the standard email field will not be populated (assuming the contact's email field wasn't already populated before installation of HEDA). What we'd like HEDA to do is to make the preferred email setting required if and only if there is an email present for a given contact. This must take into account not just the standard HEDA email fields of "University, Work, and Alternate," but also take into account any custom email fields that the admin may have created for their contact.

### LINKED ISSUE #374: HEDA should preserve the standard email for a contact by copying that email into the HEDA Alternate Email field
Today, if a contact exists prior to the installation of HEDA and that contact has an email address, the salesforce standard email address is hidden from view in the HEDA layout. Moreover, if you then enter in one of the HEDA email address fields: University, Work, Alternate, and set one of them as the preferred email, then that email will overwrite the standard email address.

What we should do is when HEDA is installed, we should check for the presence of an email in the standard email field for each existing contact. If the standard email field is populated, then we should copy the email to the Alternate email field and set the Alternate email field as the preferred email. This will preserve the standard email and also make it the preferred email by default. The admin can change their preferred to another email manually if desired.

davidjray:
# Critical Changes

# Changes
- Enhancements to how the Preferred Email field is parsed for Contacts. System Administrators can now add a custom email field and add that field as an option to the Preferred Email field pick-list. 
- Preferred Email is now required when any email fields are populated on a Contact.
- Preferred Email field cannot be set to an email field with a blank value.
- On new installations of HEDA if there is a value in the standard Email field it will be copied to the Alternate email field and Alternate will be set as the preferred email.

# Issues Closed
Fixes #371 
Fixes #372 
Fixes #373 
Fixes #374 
Fixes #344

### DIFF
--- a/src/classes/CON_Email.cls
+++ b/src/classes/CON_Email.cls
@@ -1,0 +1,132 @@
+/*
+    Copyright (c) 2014 Salesforce.org
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of Salesforce.org nor the names of
+      its contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+
+    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+    FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+    INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+/**
+* @author Salesforce.org
+* @date 2016
+* @group Contacts
+* @group-content ../../ApexDocContent/Contacts.htm
+* @description A Class to handle Contact Email related Items
+*
+*/
+public class CON_Email {
+
+    /*******************************************************************************************************
+    * @description Sets the Email field depending on the value of Preferred Email
+    * @param Contact the contact to update
+    */
+	public static void processPreferredEmail(Contact contact, List<Contact> oldlist) {
+
+		Boolean isUpdate =  (oldlist != null) ? true : false;
+
+	    // Store default email field mappings to preferred email pick list.
+	    Map<String,String> preferredEmailMappings = new Map<String,String>{   
+	        'University' => 'UniversityEmail__c',
+	        'Work' => 'WorkEmail__c',
+	        'Alternate' => 'AlternateEmail__c'
+	    };
+
+	    Map<String,String> valuedEmailMap = emailFieldsWithValue(contact);
+
+        if( valuedEmailMap.size() > 0 ) {
+
+            // Enforce preferred email field
+            if(String.isBlank(contact.Preferred_Email__c)){
+                contact.addError( Label.PreferredEmailRequiredError );
+            }
+
+            // Get the Api name that matches the value in Preferred Email
+            String preferredApiField = ( preferredEmailMappings.containsKey(contact.Preferred_Email__c) ) ? preferredEmailMappings.get(contact.Preferred_Email__c) : valuedEmailMap.get(contact.Preferred_Email__c);
+
+            String prefEmailVal = (String)contact.get(preferredApiField);
+
+            if (  String.isNotBlank( prefEmailVal ) ) { 
+                contact.Email = prefEmailVal;
+            } else {
+                contact.addError(Label.PreferredEmailMatchNotNull);
+            }      
+        
+        } else {
+
+            // Cannot have preferred email set if there are no emails present
+            if( !String.isBlank(contact.Preferred_Email__c)) {
+                contact.addError(Label.PreferredEmailMatchNotNull);                    
+            }
+
+            // oldlist is NULL onInsert so we check to make sure this logic only applies on updates.
+			if( isUpdate && String.isNotBlank(contact.Email) ) {
+	            Map<ID, Contact> oldmap = new Map<ID, Contact>( (List<Contact>)oldlist); 
+
+	            // Get the old email values - if any - to compare.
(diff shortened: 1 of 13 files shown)


davidjray:
We want to include some additional documentation to support these changes. Before merging, a few of us will meet to plan that out.

davidjray:
Re: my comment above, documentation is being written that will be released near the production release of these features, so we can move forward with this PR.

davidjray:
I think this PR affects #344 because this actually makes the addition of other items in the preferred email field configurable.

wongmike:
We actually decided against doing #344 but this PR does affect it because apparently it wasn't working properly if someone added their own personal email field.

davidjray:
Thanks @wongmike ! - I'll update the PR to reflect this Fixes #344 indirectly.

src/classes/CON_Preferred_TDTM.cls
```diff
@@ -36,42 +36,36 @@
...
+        if((triggerAction == TDTM_Runnable.Action.BeforeInsert || triggerAction == TDTM_Runnable.Action.BeforeUpdate) && newlist != null) {
+
+            for(SObject so : newlist) {
+                Contact contact = (Contact)so;
+
+                // Process the prefered Email
+                CON_Email.processPreferredEmail(contact, oldlist);
+                
+                // Process the Preferred Phone
+                if(!String.isBlank(contact.PreferredPhone__c)) {
```
ceiroa: I think you just made whitespace changes in this section, but we should open a work item to make this translatable. It has the same problem mentioned above, where the dropdown could be modified and this would stop working. Maybe we can tackle both issues at the same time.

src/classes/CON_Email.cls
```diff
@@ -0,0 +1,132 @@
...
+
+        if( valuedEmailMap.size() > 0 ) {
+
+            // Enforce preferred email field
+            if(String.isBlank(contact.Preferred_Email__c)){
+                contact.addError( Label.PreferredEmailRequiredError );
+            }
+
+            // Get the Api name that matches the value in Preferred Email
+            String preferredApiField = ( preferredEmailMappings.containsKey(contact.Preferred_Email__c) ) ? preferredEmailMappings.get(contact.Preferred_Email__c) : valuedEmailMap.get(contact.Preferred_Email__c);
```
ceiroa: Oh, I think I understand how. We first look at our custom email fields. If the preferred isn't one of them, we look at any other field of Email type, correct?

Maybe the easiest to make sure the dropdown values are translatable would be to avoid the mapping altogether?

src/classes/CON_Email.cls
```diff
@@ -0,0 +1,132 @@
...
+            }
+
+            // oldlist is NULL onInsert so we check to make sure this logic only applies on updates.
+			if( isUpdate && String.isNotBlank(contact.Email) ) {
+	            Map<ID, Contact> oldmap = new Map<ID, Contact>( (List<Contact>)oldlist); 
+
+	            // Get the old email values - if any - to compare.
+	            Map<String,String> oldValuedEmails = emailFieldsWithValue( oldmap.get(contact.Id) );
+
+	            // If this is an update and the contact contained previous emails we want to clear the estandard email field
```
ceiroa: Sorry, I'm not totally clear why we are doing this. Maybe a reference to the work item in the comment would help?

src/classes/CON_Email.cls
```diff
@@ -0,0 +1,132 @@
...
+	                contact.Email = null; 
+	            }
+	        } else if(String.isNotBlank(contact.Email)) {
+	        	copyStdEmailToAlternate(contact);
+	        }
+        } 
+	}
+
+    /*******************************************************************************************************
+    * @description Copys the value of the standard Email field to the Alternate Email field if Email has a value and no other emails.
```
ceiroa: Typo :-) (copys)

src/classes/CON_Email.cls
```diff
@@ -0,0 +1,132 @@
...
+
+	            // Get the old email values - if any - to compare.
+	            Map<String,String> oldValuedEmails = emailFieldsWithValue( oldmap.get(contact.Id) );
+
+	            // If this is an update and the contact contained previous emails we want to clear the estandard email field
+	            if (oldValuedEmails.size() > 0) {
+	                contact.Email = null; 
+	            }
+	        } else if(String.isNotBlank(contact.Email)) {
+	        	copyStdEmailToAlternate(contact);
```
ceiroa: Should this be happening on every insert an update or only when HEDA is installed for the first time?

src/classes/CON_Email.cls
```diff
@@ -0,0 +1,132 @@
...
+        
+        } else {
+
+            // Cannot have preferred email set if there are no emails present
+            if( !String.isBlank(contact.Preferred_Email__c)) {
+                contact.addError(Label.PreferredEmailMatchNotNull);                    
+            }
+
+            // oldlist is NULL onInsert so we check to make sure this logic only applies on updates.
+			if( isUpdate && String.isNotBlank(contact.Email) ) {
```
davidjray: We are not using this class in the delete context, only insert/update. However, I agree checking to make sure contact isn't null will make the method a little more reusable/future proof

src/classes/CON_Email.cls
davidjray: Yes, each update/insert to catch external system integrations for inserts.

src/classes/CON_Email_TEST.cls
```diff
@@ -0,0 +1,113 @@
...
+		Test.startTest();
+	    CON_Email_BATCH contbatch = new CON_Email_BATCH();
+	    Database.executeBatch( contbatch );
+		Test.stopTest();
+
+		// After batch run assert
+		List<Contact> contacts = [SELECT Id, Name, Email, AlternateEmail__c FROM Contact WHERE Email != null];
+
+		for(Contact c : contacts) {
+			System.assertEquals( c.AlternateEmail__c , c.Email );
```
davidjray: Really good catch. I misread what the UTIL_UnitTestData_TEST was doing.

src/classes/CON_Email.cls
```diff
@@ -0,0 +1,132 @@
...
+* @group Contacts
+* @group-content ../../ApexDocContent/Contacts.htm
+* @description A Class to handle Contact Email related Items
+*
+*/
+public class CON_Email {
+
+    /*******************************************************************************************************
+    * @description Sets the Email field depending on the value of Preferred Email
+    * @param Contact the contact to update
```
davidjray: Updated with second parameter

src/classes/CON_Preferred_TDTM.cls
davidjray: Issue created #400

src/classes/CON_Email.cls
```diff
@@ -0,0 +1,145 @@
...
+
+        // No need to run if we are in a delete context
+        if(contact == null) {
+            return;
+        }
+
+		Boolean isUpdateOrDelete = (oldlist != null) ? true : false;
+
+	    // Store default email field mappings to preferred email pick list.
+	    Map<String,String> preferredEmailMappings = new Map<String,String>{
```
ceiroa: Could we just take this mapping out completely? Would that directly make the preferred picklist translatable?

ceiroa:
Thanks for implementing all the changes, David. I just have one question about the email mappings. Everything else looks good.

src/classes/CON_Email.cls
davidjray: As I see it, we cannot take out the mappings because the values available in the Preferred Email field are not directly related to the actual custom fields. We could possibly use `string.startsWith(prefix)` but it is not a perfect 1:1 mapping of field labels and preferred email options.

PR #399, opened 2017-03-14
Feature/error handling

### LINKED ISSUE #284: Error Handling Modification Implementation
Based on conversations on Tuesday 6/28, its been determined we need to modify and narrow the scope of the error handling framework included in the HEDA. A specification outlining a proposed set of error handling use-cases and architectures should be the outcome of this story.

### LINKED ISSUE #332: User receives HEDA error message when Merging Contacts using DemandTools with Salesforce Merge Disabled
We believe this to be a false positive. If so, we should avoid sending this error notification. (We should probably work on this at the same time we redo error handling #284).

davidjray:
# Critical Changes

# Changes
- Error Handling has been modified to allow exceptions messages to display to users. Previously exceptions were caught, an error log was stored, and an email was sent to an admin or other configured user. With these changes, errors that happen during record processing will be exposed to the user or other callers in the system.
- Common error messages will be caught and re-written to be more clear, including required field errors and custom validation rule errors.
- Transactions that happen without a user present including future calls, and scheduled APEX etc will be monitored with a scheduled job. When errors occur they will be logged and an Email will be sent to the configured user.
# Issues Closed
Fixes #284 
Fixes #332


src/classes/ERR_ExceptionHandler_TEST.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>36.0</apiVersion>
```
ceiroa: Update API version to 39?

ceiroa:
I don't mean to put pressure on you @davidjray, or anything like that, but let me summarize what we need in order to move ahead with this PR:

1. Clarification on removal of re-entrant flags
2. Branch that contains all changes from this branch  plus code that would help to test these changes
3. Instructions on how to test using the aforementioned branch (please add instructions as comment in PR or reference them from PR)

Thank you!

davidjray:
No pressure @ceiroa, these are small changes, but they have a large impact. As I am investigating the removal of the reentrant flags, I am believing more and more we have a true recursion problem that was silently being caught.

ceiroa:
@davidjray Is this ready for review?

davidjray:
@ceiroa here is my response to your points from your previous comment.

1. Clarification on removal of re-entrant flags 
    - This was done because I was running into errors while testing. After further investigation, the cause of the errors in the updated unit tests was due to self-referencing Affiliation Triggers. The affiliation trigger was performing DML on Contacts in the before trigger context. This was happening in both AFFL_ContactChange_TDTM and AFFL_AccChange_TDTM. This caused the errors. I switched the updates of the related objects to the After Update context, and that has resolved the problems.
2. Branch that contains all changes from this branch  plus code that would help to test these changes 
    - This has been pushed as the branch [TEST-error-handling](https://github.com/SalesforceFoundation/HEDAP/compare/TEST-error-handling)
3. Instructions on how to test using the aforementioned branch (please add instructions as comment in PR or reference them from PR) 
    - I will share a QUIP doc containing the instructions

I am continuing to do a sanity check on the remaining unit tests and I will notify you when this is ready for review.

davidjray:
@ceiroa take a look now, this should be ready for review. Thanks!

src/classes/STG_InstallScript.cls
```diff
@@ -133,7 +138,7 @@ global without sharing class STG_InstallScript implements InstallHandler {
     * @return void
     */
     @TestVisible 
-    private static void updateDefaultTdtmConfig(List<TDTM_Global_API.TdtmToken> defaultTdtmConfig) {
+    global static void updateDefaultTdtmConfig(List<TDTM_Global_API.TdtmToken> defaultTdtmConfig) {
```
ceiroa: Any specific reason why this method has been made global?

ceiroa:
Great job, @davidjray. I had a question about a method in the install script being changed to global. But since most other methods in that class are already global (in most cases to be able to call them from the anonymous window), I think it's ok.

davidjray:
@ceiroa I made it global while I was developing this and testing the during development. I decided to leave it so it can be called from the anonymous apex window. Much like other methods in the class.

PR #408, opened 2017-04-08
Preferred Email enhancements and configuration

### LINKED ISSUE #406: HEDA error received as part of 1.30 release
This error is sent via email when a custom Preferred Email picklist value does not match the custom email field's "label". Note: the error does not prevent the record from being created.

For example:

- Custom email field Label: Custom Email Field
- Custom email field Label: FAFSA Email
- Preferred Email picklist value: No match

I do not receive an error email when I use the Preferred Email picklist value of "Custom Email Field" or "FAFSA"

Per @wongmike I am opening this up as a bug. @davidjray - this bug may be associated to the following 1.30 release change but I need you to confirm: **_Enhanced how the Preferred Email field controls email addresses for Contacts. Admins can now add custom email fields and add those fields as options in the Preferred Email field picklist._**

### LINKED ISSUE #407: Make Preferred Email enhancements introduced in 1.30 optional
Due to the feedback received from the community we will make the Preferred Email enforcements optional.

davidjray:
# Critical Changes
- Added an option in the Accounts and Contacts tab of HEDA Settings to not require that Contacts have a Preferred Email. While Preferred Email is required by default, you can disable the requirement and we do not enforce the requirement for orgs upgrading to HEDA 1.31 from an earlier version.
# Changes
- Added a batch process in HEDA Settings to clean email data for Contacts after the Preferred Email enhancements take effect. Run the batch cleanup to update Contacts' emails and prevent errors for Contacts who didn't previously have a Preferred Email specified. 

# Issues Closed
Fixes #406 
Fixes #407


src/classes/CON_Email.cls
```diff
@@ -43,103 +43,184 @@ public class CON_Email {
...
+                    // Get the first item from the list either one or many
+                    contact.Email = emailFields.valuedFields[0].value;
+                    contact.Preferred_Email__c = emailFields.valuedFields[0].prefLabel;
+                }
             }
 
-            // Get the Api name that matches the value in Preferred Email
-            String preferredApiField = ( preferredEmailMappings.containsKey(contact.Preferred_Email__c) ) ? preferredEmailMappings.get(contact.Preferred_Email__c) : valuedEmailMap.get(contact.Preferred_Email__c);
+            // Enforce preferred email field unless validation is disabled
+            if(validatePreferred && String.isBlank(contact.Preferred_Email__c)){
```
ceiroa: This will run for both Batch and non-Batch context. That's the intention, right?

src/aura/STG_CMP_Addr/STG_CMP_Addr.cmp
```diff
@@ -10,6 +10,49 @@
...
+            <aura:set attribute="else">
+                <ui:outputText value="{!$Label.c.stgHelpContactPreferredEmail}" class="slds-text-body--small" />
+            </aura:set>
+            </aura:if>
+        </div>
+
+        <div class="slds-col slds-size--1-of-2 slds-p-right--xx-large">
+            <h2>Preferred Email Data Clean Up:</h2>
+            <div class="slds-text-body--small">
+                <p>You can run this data clean up process if the Preferred Email enforcement has been disabled for some time, or if the Preferred Email enforcement is being turned on for the first time.</p>
```
davidjray: @lianeyee Do you mind looking over the info here. We are trying to give a short explanation of what this batch process does.

src/labels/CustomLabels.labels
```diff
@@ -89,6 +89,14 @@
         <value>Male</value>
     </labels>
     <labels>
+        <fullName>PreferredEmailMatchMustExist</fullName>
+        <categories>Contacts</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>PreferredEmailMatchMustExist</shortDescription>
+        <value>The value selected in Preferred Email does not match any email fields. Values in Preferred Email must match a custom email field label.</value>
```
davidjray: @lianeyee an error message to review

src/labels/CustomLabels.labels
```diff
@@ -329,6 +337,14 @@
         <value>Select this option to enable multiple address management for Contacts.</value>
     </labels>
     <labels>
+        <fullName>stgHelpContactPreferredEmail</fullName>
+        <categories>Settings</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>stgHelpContactPreferredEmail</shortDescription>
+        <value>Select this box to disable the enforcement of the Preferred Email field for Contacts.</value>
```
davidjray: @lianeyee More test to review: This is the help text displayed in the HEDA settings

davidjray:
Screenshot of new setting and button placement: 
screen shot 2017-04-10 at 2 44 46 pm

@lianeyee

src/classes/CON_Email_TEST.cls
```diff
@@ -106,40 +112,78 @@ private class CON_Email_TEST {
 
 	// Test the batch Class
 	@isTest static void testContactEmailBatch() {
+
+		UTIL_CustomSettingsFacade.getSettingsForTests(
+            new Hierarchy_Settings__c(Disable_Preferred_Email_Enforcement__c = false)
+            );
+
+        UTIL_CustomSettingsFacade.getSettings();
```
ceiroa: Why are we calling UTIL_CustomSettingsFacade.getSettingsForTests and then UTIL_CustomSettingsFacade.getSettings()? Wouldn't the second call override the first one? Also, the Setup method is being called. Do we care about that?

src/classes/STG_InstallScript.cls
```diff
@@ -250,4 +252,20 @@ global without sharing class STG_InstallScript implements InstallHandler {
         }
         return null;
     }
+
+    /*******************************************************************************************************
+    * @description Disables Preferred Email handling when upgrading from 1.30
+    * @return void
+    */
+    global static void disablePreferredEmailEnforcement(Version previousVersion) {
```
ceiroa: Since Disable_Preferred_Email_Enforcement__c  is a new field, and the default value is false, do we even need this method?

src/classes/CON_Email.cls
```diff
@@ -43,103 +43,184 @@ public class CON_Email {
...
+        CON_EmailFieldList emailFields = new CON_EmailFieldList(contact);
 
-	    Map<String,String> valuedEmailMap = getEmailFieldsWithValue(contact);
+        if( emailFields.valuedCount() > 0 ) {
 
-        if( valuedEmailMap.size() > 0 ) {
+            // Check the presence of a special cases when running a batch process.
+            // This will prevent some validation errors below to allow batches to complete successfully.
+            if( validatePreferred && System.isBatch() ) {
+                if( !emailFields.valueExists(contact.Email) && String.isBlank(contact.Preferred_Email__c)) {
```
davidjray: Great question, it is not necessary to check, because te next check overrides the preferred email, but to make it clearer what is happening I agree we should add the check for the blank email here.

src/classes/CON_Email.cls
davidjray: Yes!

src/classes/STG_InstallScript.cls
davidjray: Yes, because we are actually setting it to true here, for existing Orgs upgrading. Remember disabled == true for the setting. I know that seems backwards, but it kinda makes sense for me since all new Orgs will have it unchecked.

ceiroa:
@davidjray I finally tried to investigate why I hadn't received the email confirming the batch process had run successfully. It seems like it hadn't. 

screen shot 2017-04-10 at 5 26 37 pm

According to https://salesforce.stackexchange.com/questions/81238/sobject-row-does-not-allow-errors-in-salesforce-before-trigger that error has to do with trying to use addError on a record that isn't part of the trigger. I'm afraid we are going to need to investigate this a bit...

src/aura/STG_CMP_Addr/STG_CMP_Addr.cmp
lianeyee: @davidjray I suggest some changes to the heading and body text.
<h2>Preferred Email Data Cleanup:</h2>
<p>Preferred Email is a required field for Contacts. If you're upgrading from an earlier version of HEDA that didn't enforce this requirement, run this batch cleanup process to make sure that all Contacts have a Preferred Email specified.</p>

src/classes/STG_InstallScript.cls
ceiroa: Ugh, yes. The "disable" meaning the validation is enabled got me. Thanks for the clarification.

src/labels/CustomLabels.labels
lianeyee: "Values in Preferred Email must match a custom email field label." @davidjray I'm not sure what this means. How would this error get triggered? Why do we explicitly say a "custom" email field label, as opposed to either one of the regular email fields or a custom one?

src/labels/CustomLabels.labels
lianeyee: @davidjray Could we say this more simply?
<value>Do not require Contacts to have a Preferred Email.</value>

lianeyee: Also, for the button text, it should say "Run Cleanup" (cleanup as one word).

lianeyee:
@davidjray I also edited the release notes. Take a look when you get a chance.

davidjray:
Looks great @lianeyee !

ceiroa:
@davidjray Ok, I tested the batch job again (after disabling state/country picklists) and it worked properly (yippee!). However, the message in settings says "An email will be sent at the completion of the batch." I did not receive an email, and I'm having a hard time locating the code that is supposed to be sending it...

src/labels/CustomLabels.labels
davidjray: This error happens when an admin adds a value to the "Preferred Email" field, but there is not actually a corresponding custom email with a matching label. When Preferred Email is selected we match the labels to the custom fields added by an administrator. If there is a match, the matching custom email fields value is copied to the standard Email field on the contact.

src/aura/STG_CMP_Addr/STG_CMP_Addr.cmp
davidjray: For existing Orgs with HEDA installed, we are actually pushing the changes with the validation disabled. So the purpose of this batch is to offer a way for current users of HEDA to clean their data if they choose to enable this feature.

ceiroa:
@davidjray It all looks good. I didn't receive the email, but I see that the code is there (and it looks correct). So maybe it's just something with my org. @chrismcmahon, please make sure you tests all cases (it's a complicated PR), and specially make sure you get the email confirmation after running the bath process (from the settings page). Thank you.

ceiroa:
@davidjray I got the email! It seems like my initial deployment of the latest changes had initially failed and I hadn't realized. I've finally been able to successfully deploy the changes, and received the email after the batch process ended.

davidjray:
That is great new @ceiroa !

src/labels/CustomLabels.labels
lianeyee: I see, I think. I was confused about where this error gets displayed. Would this work? "We can't find the email specified in Preferred Email. Make sure that you're entering the field label of an existing custom email field." @davidjray

lianeyee: Sure, I'm OK with handling as a separate WI for the next release. Would you or Dave mind creating it? Thanks, Carlos.

davidjray: @lianeyee since we are working one some updates for the upcoming release, I can work this in. Thank you!

lianeyee: Fantastic!

lianeyee:
@davidjray Can we tweak the informational text in src/aura/STG_CMP_Addr/STG_CMP_Addr.cmp? The current wording is a bit vague. I can't tell what the batch job is going to do. I'd like to say specifically that it copies the email from Standard to Alternate and sets Alternate as Preferred. We might need to further wordsmith this bit together. LMK what you think.

<p>Run this data cleanup process if you're enabling the Preferred Email requirement for the first time or after a period of not enforcing the requirement.</p>

<p>The cleanup sets a Preferred Email for existing Contacts who don't have a Preferred Email specified. We copy the Contact's email from the standard Email field to the Alternate Email field and set that field as the Preferred Email.</p>

PR #410, opened 2017-04-12
Feature/email enhancements

davidjray:
# Critical Changes

# Changes
- Allow partial completion of the Preferred Email batch process on HEDA installation.
# Issues Closed


src/classes/CON_Email.cls
```diff
@@ -114,24 +114,30 @@ public class CON_Email {
...
-                // we do this because when a user deletes all email values in the HEDA contact without clearing
-                // the value if email, there would be an email address that the user tried to delete.
-	            if (oldEmailFields.valuedCount() > 0) {
-	                contact.Email = null;
-	            }
-	        } else if(String.isNotBlank(contact.Email)) {
-	        	copyStdEmailToAlternate(contact);
-	        }
+            // We are not going to format data for orgs that are not enforcing the preferred email field
+            if(validatePreferred && String.isNotBlank(contact.Email)) {
```
davidjray: @ceiroa I made an improvement to this section. While testing the batch, I noticed we needed to treat Inserts and Updates the same unless the records were deleted.

Also, I made sure the data copy only happens if the validation is active. We don't want to reformat the data in an org on an upgrade unless they have specified.

PR #423, opened 2017-05-15
Fix for account null pointer exception

### LINKED ISSUE #422: NullPointerException when trying to change Account record type
When you try to change the record type of an account you get a NPE:

screen shot 2017-05-12 at 3 25 23 pm

xinnanlisf:
# Critical Changes

# Changes
Improved error message and handling for errors caused by an affiliation mapping referring to an invalid Account record type. For more information about affiliation mappings, see the HEDA documentation: https://powerofus.force.com/articles/Resource/HEDA-Create-Accounts-and-Affiliations#!#mappings

# Issues Closed
Fixes #422

### DIFF
--- a/src/classes/AFFL_MultiRecordTypeMapper.cls
+++ b/src/classes/AFFL_MultiRecordTypeMapper.cls
@@ -145,17 +145,19 @@ public with sharing class AFFL_MultiRecordTypeMapper {
         			return fieldName;
         		}
             }else{
-                //TODO: Need Liane's review on the error msg
                 //Setup null pointer error message
                 String[] params = new String[]{
                     String.valueOf(recTypeName)
                 };
-                String nullPointerMsg = String.format(label.afflNullPointerError, params);
-                Trigger.new[0].addError(nullPointerMsg);
+                String nullPointerMsg = String.format(Label.afflNullPointerError, params);
+                throw new AfflNullPointerException(nullPointerMsg);
             }
 		}
 		return null;
     }
+    
+    public class AfflNullPointerException extends Exception {}
+
     
     /*******************************************************************************************************
     * @description Tells caller if Account record type needs Program Enrollment auto-creation.
--- a/src/classes/AFFL_MultiRecordTypeMapper.cls
+++ b/src/classes/AFFL_MultiRecordTypeMapper.cls
@@ -136,12 +136,23 @@ public with sharing class AFFL_MultiRecordTypeMapper {
     ********************************************************************************************************/
 	public String getKeyAfflFieldByAccRecordType(ID recordTypeId) {
 		for(String recTypeName : mapAccRecTypeToContactLabel.keySet()) {
-    		ID rcId = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get(recTypeName).getRecordTypeId();
-    		if(rcId == recordTypeId) {
-    			String fieldLabel = mapAccRecTypeToContactLabel.get(recTypeName);
-    			String fieldName = contactLabelNames.get(fieldLabel);
-    			return fieldName;
-    		}
+            if (Schema.Sobjecttype.Account.getRecordTypeInfosByName().get(recTypeName) != null)
+            {
+        		ID rcId = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get(recTypeName).getRecordTypeId();
+        		if(rcId == recordTypeId) {
+        			String fieldLabel = mapAccRecTypeToContactLabel.get(recTypeName);
+        			String fieldName = contactLabelNames.get(fieldLabel);
+        			return fieldName;
+        		}
+            }else{
+                //TODO: Need Liane's review on the error msg
+                //Setup null pointer error message
+                String[] params = new String[]{
+                    String.valueOf(recTypeName)
+                };
+                String nullPointerMsg = String.format(label.afflNullPointerError, params);
+                Trigger.new[0].addError(nullPointerMsg);
+            }
 		}
 		return null;
     }
--- a/src/classes/AFFL_MultiRecordTypeMapper.cls
+++ b/src/classes/AFFL_MultiRecordTypeMapper.cls
@@ -147,7 +147,7 @@ public with sharing class AFFL_MultiRecordTypeMapper {
             }else{
                 //Setup null pointer error message
                 String[] params = new String[]{
-                    String.valueOf(recTypeName)
+                    recTypeName
                 };
                 String nullPointerMsg = String.format(Label.afflNullPointerError, params);
                 throw new AfflNullPointerException(nullPointerMsg);
--- a/src/classes/ERR_ExceptionHandler.cls
+++ b/src/classes/ERR_ExceptionHandler.cls
@@ -63,7 +63,8 @@ public with sharing class ERR_ExceptionHandler {
                 message = EXCEPTION_MESSAGES.get(STATUS) + message.substringBetween(SUBSTRINGS.get(STATUS)[0], SUBSTRINGS.get(STATUS)[1]);
                 ex.setMessage(message);
                 break;
-            }else if ((typeName.contains(STATUS))){
+            }else if ((typeName.contains(STATUS)))
+            {
                 ex.setMessage(message);
                 break;                
             }
--- a/src/classes/ERR_ExceptionHandler.cls
+++ b/src/classes/ERR_ExceptionHandler.cls
@@ -58,7 +58,6 @@ public with sharing class ERR_ExceptionHandler {
     public static void beautifyExceptionMessage(Exception ex) {
         String message = ex.getMessage();
         String typeName = ex.getTypeName();
-        system.debug(typeName);
         for (String STATUS : EXCEPTION_MESSAGES.keySet()) {
             if (message.contains(STATUS)) {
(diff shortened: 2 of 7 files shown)


ceiroa:
@davidjray Would you mind taking a first pass at this, including PR text content (Liane's mention should not be in "Changes"), code review, checking out the code, deploying it, and verifying it works as expected? Thank you!

lianeyee:
@xinnanlisf Thanks for the ping! For the label, I suggest: "The {0} Account record type doesn't exist. Contact your admin for help."

xinnanlisf:
@lianeyee 

Thank you for your feedback!

I think the first suggestion makes sense. I will go ahead and make the change. 

For the affiliation mapping thing, can we say: "Contact your admin to check if affiliation mappings and account record types are set up correctly." 

Completely open to suggestions. :)

lianeyee:
Good, let's mention affiliation mappings, then. How about condensing everything into a single label? "The {0} Account record type doesn't exist. Ask your admin to check record types in affiliation mappings."

lianeyee:
@jacebryan Care to take a crack at clarifying the issue description?

src/labels/CustomLabels.labels
```diff
@@ -201,6 +201,14 @@
         <value>Address records should only be created as children of Account or Contact.</value>
     </labels>
     <labels>
+        <fullName>afflNullPointerError</fullName>
+        <categories>Error</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>afflNullPointerError</shortDescription>
+        <value>One of the Account Record Types: {0} does not exist. Please contact Admins.</value>
```
davidjray: Go ahead and place Liane's updates here.

davidjray:
@xinnanlisf As also mentioned, you'll want to make sure you "@" mention people in comments, since our release notes are generated from the PR notes.

jacebryan:
@lianeyee - I think your label is good. I am wondering if there is a way to link to the Release Notes into the error label displayed? Or perhaps a link tied directly to the Github issue? So an admin knows where to go for additional help after being contacted by their customer?

lianeyee:
@jacebryan I wouldn't link to the release notes from an error message in the app (not a great experience to link a general audience end user to GitHub). It makes more sense is to include this change in the release notes' Changes section, and include a link to this existing doc: https://powerofus.force.com/articles/Resource/HEDA-Create-Accounts-and-Affiliations

@ceiroa I'm curious why you commented above that this item shouldn't be mentioned in release notes?

davidjray:
@lianeyee , he just meant we shouldn't use "@" mentions in the release notes section.

lianeyee:
@xinnanlisf I didn't want to overwrite your Changes text without letting you know first. I'll make this edit if you're OK with it: "Improved error message and handling for errors caused by an affiliation mapping referring to an invalid Account record type. For more information about affiliation mappings, see the HEDA documentation: https://powerofus.force.com/articles/Resource/HEDA-Create-Accounts-and-Affiliations#!#mappings"

FYI @jacebryan

ceiroa:
@lianeyee Sorry, I didn't mean the item shouldn't be included (though it's debatable whether a label deserves its own line, considering it's part of the issue itself). I just meant that the string "@lianeyee please review the error message" shouldn't be in there, or it will end up in the release notes. 

Also, I agree we probably don't want links to documentation or release notes in error messages.

xinnanlisf:
@lianeyee 

Yea, feel free to make the changes.

FYI, I have removed all the @ mentions.

xinnanlisf:
@davidjray 

Thank you for the feedbacks!

I will go ahead and make the changes.

src/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -136,12 +136,23 @@ public with sharing class AFFL_MultiRecordTypeMapper {
...
+        			return fieldName;
+        		}
+            }else{
+                //TODO: Need Liane's review on the error msg
+                //Setup null pointer error message
+                String[] params = new String[]{
+                    String.valueOf(recTypeName)
+                };
+                String nullPointerMsg = String.format(label.afflNullPointerError, params);
+                Trigger.new[0].addError(nullPointerMsg);
```
xinnanlisf: @davidjray 

I have been thinking about how to bulkify this. It seems that we can't show which record type is creating an error if bulkify unless we let it return a map. Maybe I am wrong, any thoughts on this?

Also, @ceiroa

xinnanlisf:
> The {0} Account record type doesn't exist. Ask your admin to check record types in affiliation mappings.

@lianeyee 

It seems that I can not put an apostrophe in the custom label value. Does 'does not exist' work too?

lianeyee:
@xinnanlisf About displaying apostrophes: try escaping it using a double apostrophe? i.e., doesn''t

xinnanlisf:
> @xinnanlisf About displaying apostrophes: try escaping it using a double apostrophe? i.e., doesn''t

@lianeyee 

I just figured as well thus I deleted my comment. Thank you for the insight thou:)

lianeyee:
Ah, no wonder I couldn't find it. I'm still not that used to Git. =) Glad you figured it out.

src/classes/AFFL_MultiRecordTypeMapper.cls
davidjray: @xinnanlisf I think the best course of action here is to throw an exception since technically the error isn't due to a particular object, but related to a configuration issue in the Org.

xinnanlisf: Great point, @davidjray. Just to clarify, do you mean we use throw exception instead of addError? However It is still in the loop, right? Do you suggest to move the error handling out of the method?

davidjray: Yes, I would throw an exception instead of `addError()`. 

You are still in the loop as part of the TDTM class, however, as I reviewed this thinking through how I would approach it I realized the problem is with the configuration and not with any particular object. 

If "bulkified", every object in the batch would show the same error, and because of that, it is not an "object" issue, but an issue with the configuration. With all that in mind, if I were working on this, I would throw an exception, halting the entire transaction inside this same method. 

Since this is part of an overall transaction managed by TDTM, the TDTM Trigger Handler will deal with error handling.

xinnanlisf:
Changes committed.

@lianeyee could you review the new error message?

@davidjray could you review the code changes?

Many thanks!

lianeyee:
@xinnanlisf The text looks good. Thanks! I'm not sure if that's how we typically handle the double apostrophe. I'll let @davidjray comment on that convention.

davidjray:
Great job @xinnanlisf! I was able to deploy this and test it successfully in a new scratch org. 

@lianeyee you can approve the changed, encoding the apostrophes in fine by me.

xinnanlisf:
Thank you @lianeyee @davidjray :) Let me know what next step is.

davidjray:
@xinnanlisf Next step is @ceiroa will review and then it will get merged.

Usually, there will only be one reviewer, but this being the first PR we had a few extra.

src/classes/ERR_ExceptionHandler.cls
```diff
@@ -56,11 +57,16 @@ public with sharing class ERR_ExceptionHandler {
     */
     public static void beautifyExceptionMessage(Exception ex) {
         String message = ex.getMessage();
+        String typeName = ex.getTypeName();
```
ceiroa: Just for my understanding, is it that in this case the identifier for the exception is in the type instead of the message?

xinnanlisf: Yes, it is identifying the type of the exception to display the exception message

src/classes/ERR_ExceptionHandler.cls
```diff
@@ -56,11 +57,16 @@ public with sharing class ERR_ExceptionHandler {
...
         for (String STATUS : EXCEPTION_MESSAGES.keySet()) {
             if (message.contains(STATUS)) {
                 message = EXCEPTION_MESSAGES.get(STATUS) + message.substringBetween(SUBSTRINGS.get(STATUS)[0], SUBSTRINGS.get(STATUS)[1]);
                 ex.setMessage(message);
                 break;
+            }else if ((typeName.contains(STATUS)))
+            {
```
xinnanlisf: Sure I will do that!

src/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -136,16 +136,29 @@ public with sharing class AFFL_MultiRecordTypeMapper {
...
+        		ID rcId = Schema.Sobjecttype.Account.getRecordTypeInfosByName().get(recTypeName).getRecordTypeId();
+        		if(rcId == recordTypeId) {
+        			String fieldLabel = mapAccRecTypeToContactLabel.get(recTypeName);
+        			String fieldName = contactLabelNames.get(fieldLabel);
+        			return fieldName;
+        		}
+            }else{
+                //Setup null pointer error message
+                String[] params = new String[]{
+                    String.valueOf(recTypeName)
```
ceiroa: Sorry, just one more thing. `recTypeName` is already a string. Do we really need the `String.valueOf` call?

xinnanlisf: That's a good catch. I will remove it.

PR #426, opened 2017-05-17
Set preferred email automatically when only a single email is present on Contact

### LINKED ISSUE #428: Auto set Preferred Email when only one custom email field is present
As part of the Preferred Email field enforcement, we should auto set Preferred Email and the Standard Email field when a single custom email field has a value.

This allows a better user experience and better compatibility when converting leads.

davidjray:
# Critical Changes

# Changes
- When a Contact has only a single email address, we set that value as the Preferred Email and copy the value to the standard Email field. This feature helps prevent errors during lead conversion.

# Issues Closed
Fixes #428

### DIFF
--- a/src/classes/CON_Email_TEST.cls
+++ b/src/classes/CON_Email_TEST.cls
@@ -55,7 +55,7 @@ private class CON_Email_TEST {
         insert listCon;
 	}
 
-	@isTest static void tedLeadConversion() {
+	@isTest static void testLeadConversion() {
 		// create a Lead
 		Lead tLead = new Lead(
 			FirstName='Joshua',
--- a/src/classes/CON_Email_TEST.cls
+++ b/src/classes/CON_Email_TEST.cls
@@ -55,6 +55,34 @@ private class CON_Email_TEST {
         insert listCon;
 	}
 
+	@isTest static void tedLeadConversion() {
+		// create a Lead
+		Lead tLead = new Lead(
+			FirstName='Joshua',
+			LastName='McTesterman',
+			Email='mctesterman@mctesersite.com',
+			Company='Test',
+			Status='Inquiry'
+		);
+
+		insert tLead;
+
+		Database.LeadConvert lc = new Database.LeadConvert();
+		lc.setLeadId(tLead.Id);
+
+		LeadStatus convertStatus = [SELECT Id, MasterLabel FROM LeadStatus WHERE IsConverted=true LIMIT 1];
+		lc.setConvertedStatus(convertStatus.MasterLabel);
+
+		Test.startTest();		
+		Database.LeadConvertResult lcr = Database.convertLead(lc);
+		Test.stopTest();
+
+		String contId = lcr.getContactId();
+		Contact c = [SELECT Id, Name, Email FROM Contact WHERE Id =:contId LIMIT 1];
+
+		System.assertEquals(c.Email,tLead.Email );
+	}
+
 	@isTest static void testSingleEmailSmartSet() {
 
 		Contact newCont = new Contact(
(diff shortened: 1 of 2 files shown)


davidjray:
@lianeyee Do you mind double checking my release notes?

lianeyee:
@davidjray Checked and edited. Feel free to tweak further if needed. Thanks!

davidjray:
Perfect @lianeyee.

davidjray:
@ceiroa let's hold off on merging this until we get sign off from @wongmike on the feature.

wongmike:
@davidjray I'll set up a call later today to review. Thanks.

wongmike:
spoke to @davidjray. No questions about this PR. Thanks.

ceiroa:
Hey @davidjray, any tests for @chrismcmahon and me on how to test this? Thanks!

ceiroa:
@davidjray Does the user still get an error during Lead Conversion if multiple email fields have a value, or is this not a plausible scenario? If it is, should we just choose the first available email (or random selected one) or do we prefer not to interfere?

src/classes/CON_Email_TEST.cls
```diff
@@ -55,6 +55,24 @@ private class CON_Email_TEST {
...
+	@isTest static void testSingleEmailSmartSet() {
+
+		Contact newCont = new Contact(
+			FirstName = 'JohnnyTest',
+			LastName = 'JohnnyTest',
+			WorkEmail__c = 'workit@test.com'
+		);
+
+		Test.startTest();
+        insert newCont;
```
ceiroa: Is the process exactly the same for converting a lead? Should we consider testing also using the Database.convertLead method, since that was the original issue?

davidjray: Since we cannot directly duplicate the scenario of converting a lead this is the next best thing. Since we are inserting a new Contact with only one email field having a value this is the same process that a lead goes through when being converted.

davidjray:
@ceiroa regarding multiple emails, this PR does not address other scenarios. I have created a new Work Item for additional investigation regarding Lead Conversion.

src/classes/CON_Email_TEST.cls
```diff
@@ -55,6 +55,52 @@ private class CON_Email_TEST {
         insert listCon;
 	}
 
+	@isTest static void tedLeadConversion() {
```
davidjray: @ceiroa is this the type of test you were asking about?

src/classes/CON_Email_TEST.cls
```diff
@@ -55,6 +55,52 @@ private class CON_Email_TEST {
         insert listCon;
 	}
 
+	@isTest static void testLeadConversion() {
```
davidjray: @ceiroa is this the type of test you were asking about?

lianeyee:
@davidjray I think this is a new scenario that we should add to https://powerofus.force.com/kA080000000HB9X?srPos=3&srKp=ka0&lang=en_US. What do you think?

New bullet: "If a Contact has only a single email address of record--and it isn't currently the standard Email value--HEDA copies that address to the standard Email field and sets the Preferred Email to the standard Email."

davidjray:
@lianeyee That is a great idea. I suggest a small update since technically the preferred email field is set to the field that had the value. Also, this only happens if the preferred email field enforcement is active. If they disable it this behavior doesn't happen.

Here is an example. University Email has a value of "ssmith@univ.edu". Then the standard Email is set to  "ssmith@univ.edu"  and Preferred Email is set to "University".

lianeyee:
Thanks, @davidjray. That clarification is helpful. I updated the release notes, too. Please take a look.

PR #434, opened 2017-05-19
Added tasks to run install script methods

davidjray:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -34,16 +34,6 @@ tasks:
                 STG_InstallScript.insertRelationshipLookups();
                 TDTM_Global_API.setTdtmConfig( TDTM_Global_API.getTdtmConfig() );
 
-    execute_install_apex_managed:
-        description: Runs most of the install script methods from STG_InstallScript class
-        class_path: cumulusci.tasks.anonymous_apex.AnonymousApexTask
-        options:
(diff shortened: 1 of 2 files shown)


cumulusci.yml
```diff
@@ -25,7 +23,27 @@ tasks:
...
+                STG_InstallScript.insertMappings();
+                STG_InstallScript.insertRelationshipLookups();
+                TDTM_Global_API.setTdtmConfig( TDTM_Global_API.getTdtmConfig() );
+
+    execute_install_apex_managed:
+        description: Runs most of the install script methods from STG_InstallScript class
+        class_path: cumulusci.tasks.anonymous_apex.AnonymousApexTask
+        options:
+            apex: >
+                hed.UTIL_CustomSettingsFacade.getSettings();
```
cdcarter: UTIL_CustomSettingsFacade isn't global, so it can't be accessed in managed installations. are the below method calls still meaningful if you don't call getSettings first? I don't know the install script well.

davidjray: Great catch, I didn't notice the failure, so I must have assumed a successful test. I'll look into this.

davidjray:
@cdcarter got this fixed up.

PR #442, opened 2017-06-13
Trigger Handler Updates

davidjray:
Added Owned by Namespace field to Trigger Handler so HEDA and other packages can isolate their Trigger Handlers for the purpose of default management. Also added the Owned by Namespace field to the TDTM_Token class.

Implemented hashCode() and equals() methods for the TDTM_Token custom type of the TDTM_Global_API class. The enables the use of sets helping to optimize the default configuration comparison algorithm. Added a handlerId field to the Token class to track when a handler could be updated instead of inserted into the database.

The setTdtmConfig() method of the TDTM_Global_API class was also updated with a new signature to accepts a List of tokens as well as a String to represent the "Namespace" of fields

Created TDTM_Manager class. This class will be instantiated when updating or installing new default TDTM triggers. Upon instantiation, the TDTM_Manager will contain the existing configuration and you can pass in a Set of tokens to compare against the current configuration to update/insert/delete. Also added a simple method on TDTM_Manager which accepts a map of Trigger Handlers and returns a TDTM_Token set.

Finally, the install script was modified to target upgrades to 1.36 so the Owned by Namespace field can be utilized moving forward for HEDA managed trigger handlers.

All these updates were important because if a developer of a package wanted to write their own Trigger Handlers to utilizeTDTM they would also have to write the code to retrieve then update and save them. 

TL;DR; - The way HEDA internally manages Trigger Handlers has been modified. These modifications have been exposed via the TDTM_Global_API class so others using TDTM can utilize the same management for their own Trigger Handlers. 

# Critical Changes

# Changes
- Enhanced the TDTM framework to prevent unintentional changes to Trigger Handlers used by other, non-HEDA packages. Use the new Owned by Namespace field on the Trigger Handler custom object to identify the package that manages a specific handler. Refer to the TDTM_Global_API documentation for more information: http://developer.salesforce.org/HEDAP/ApexDocumentation/TDTM_Global_API.html

# Issues Closed

### DIFF
--- a/src/classes/TDTM_Config.cls
+++ b/src/classes/TDTM_Config.cls
@@ -33,8 +33,7 @@
  * @date 2014
  * @group TDTM
  * @group-content ../../ApexDocContent/TDTM.htm
- * @description Class that queries custom object to retrieve classes to call for creation or modification
- * of a record of a specific object type.
+ * @description A Class to retreive and interact with the TDTM configurtaion
  */
 public class TDTM_Config {
 
@@ -71,7 +70,7 @@ public class TDTM_Config {
 
         for(Trigger_Handler__c handler : tdtmConfig) {
             tokens.add(new TDTM_Global_API.TdtmToken(handler.Class__c, handler.Object__c, handler.Trigger_Action__c,
-                    Integer.valueOf(handler.Load_Order__c), handler.Active__c, handler.Asynchronous__c,
+                    handler.Load_Order__c, handler.Active__c, handler.Asynchronous__c,
                     handler.Filter_Field__c, handler.Filter_Value__c, handler.User_Managed__c, handler.Owned_by_Namespace__c));
         }
 
@@ -91,7 +90,7 @@ public class TDTM_Config {
         List<Trigger_Handler__c> existingHandlers = TDTM_DefaultConfig.getDefaultRecords();
         for(Trigger_Handler__c handler : existingHandlers) {
             tokens.add(
-                new TDTM_Global_API.TdtmToken(handler.Class__c, handler.Object__c, handler.Trigger_Action__c, Integer.valueOf(handler.Load_Order__c), handler.Active__c, handler.Asynchronous__c, handler.Filter_Field__c, handler.Filter_Value__c, handler.User_Managed__c, handler.Owned_by_Namespace__c )
+                new TDTM_Global_API.TdtmToken(handler.Class__c, handler.Object__c, handler.Trigger_Action__c, handler.Load_Order__c, handler.Active__c, handler.Asynchronous__c, handler.Filter_Field__c, handler.Filter_Value__c, handler.User_Managed__c, handler.Owned_by_Namespace__c )
             );
         }
 
@@ -108,8 +107,8 @@ public class TDTM_Config {
         tdtmTokensCache = tdtmTokens;
 
         TDTM_Manager manager = new TDTM_Manager(namespace);
-        Set<TDTM_Global_API.TdtmToken> tSet =  new Set<TDTM_Global_API.TdtmToken>(tdtmTokens);
-        manager.updateDefaultTdtmConfig( tSet );
+        Set<TDTM_Global_API.TdtmToken> tokenSet =  new Set<TDTM_Global_API.TdtmToken>(tdtmTokens);
+        manager.updateDefaultTdtmConfig( tokenSet );
     }
 
     /*******************************************************************************************************
--- a/src/classes/TDTM_Config.cls
+++ b/src/classes/TDTM_Config.cls
@@ -33,7 +33,7 @@
  * @date 2014
  * @group TDTM
  * @group-content ../../ApexDocContent/TDTM.htm
- * @description A Class to retreive and interact with the TDTM configurtaion
+ * @description A Class to retreive, cache and interact with the Org wide TDTM configurtaion
  */
 public class TDTM_Config {
 
@@ -84,6 +84,9 @@ public class TDTM_Config {
     * @return List<TDTM_Global_API.TdtmToken> List of trigger handler tokens with the default values.
     *********************************************************************************************************/
     public static List<TDTM_Global_API.TdtmToken> getDefaultTdtmConfig() {
+        if(tdtmTokensCache != null) {
+            return tdtmTokensCache;
+        }
 
         List<TDTM_Global_API.TdtmToken> tokens = new List<TDTM_Global_API.TdtmToken>();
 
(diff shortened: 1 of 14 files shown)


davidjray:
**Testing Instructions:**
1. Deploy Metadata to new Org
2. Run updated code inside Install Script lines [54-56] (https://github.com/SalesforceFoundation/HEDAP/blob/feature/dr-trigger-handler-manage/src/classes/STG_InstallScript.cls#L54-L56)
3. Verify trigger handlers were added
4. Make an arbitrary change to handler definitions in TDTM_DefaultConfig. You could even delete some entries for testing.
5. Run code from lines [76-78] (https://github.com/SalesforceFoundation/HEDAP/blob/feature/dr-trigger-handler-manage/src/classes/STG_InstallScript.cls#L76-L78)
6. Verify the updates you made before were persisted to the existing handlers in the DB.

Other exploratory testing should be done using the techniques above to verify functionality.

davidjray:
@lianeyee take a peek at my TLDR in the opening comment. It sums up what is happening fairly well.

I am not sure of the best place to document/communicate these changes or how they should be placed in the release notes. 

@jacebryan

lianeyee:
@davidjray @jacebryan I've not looked into TDTM much before today. Much to my (pleasant) surprise, it appears we have this doc about it:
http://developer.salesforce.org/HEDAP/ApexDocumentation/TDTM_Global_API.html

Anyone know how this gets updated and by whom?

lianeyee:
Carlos says the docs are auto-generated from ApexDocs annotations in the code. If annotations can handle describing the changes, then for release notes, ca we say something like:

Enhanced the TDTM framework to prevent unintentional changes to Trigger Handlers used by other, non-HEDA packages. Use the new Owned by Namespace field on the Trigger Handler custom object to identify the package that manages a specific handler. Refer to the TDTM_Global_API documentation for more information: http://developer.salesforce.org/HEDAP/ApexDocumentation/TDTM_Global_API.html

src/classes/TDTM_Config.cls
```diff
@@ -37,52 +37,52 @@
...
 	        tdtmConfig = TDTM_DefaultConfig.getDefaultRecords();
         }
-        
+
         for(Trigger_Handler__c handler : tdtmConfig) {
-            tokens.add(new TDTM_Global_API.TdtmToken(handler.Class__c, handler.Object__c, handler.Trigger_Action__c, 
-                    Integer.valueOf(handler.Load_Order__c), handler.Active__c, handler.Asynchronous__c, 
-                    handler.Filter_Field__c, handler.Filter_Value__c, handler.User_Managed__c));
+            tokens.add(new TDTM_Global_API.TdtmToken(handler.Class__c, handler.Object__c, handler.Trigger_Action__c,
+                    Integer.valueOf(handler.Load_Order__c), handler.Active__c, handler.Asynchronous__c,
```
davidjray: Good catch. In my experience, a Decimal would accept an integer as a value which is why the tests passed. I updated the unit test to sprinkle in some floats into the configuration to make sure they are accepted and work as values.

src/classes/TDTM_Config.cls
```diff
@@ -91,37 +91,27 @@ public class TDTM_Config {
...
-                    Trigger_Action__c = token.actions, Active__c = token.active, Load_Order__c = token.loadOrder, 
-                    Asynchronous__c = token.async, Filter_Field__c = token.filterField, Filter_Value__c = token.filterValue,
-                    User_Managed__c = token.userManaged);
-            handlers.add(handler);
-        }
-        UTIL_Debug.debug('****Handlers to insert: ' + JSON.serializePretty(handlers));
-        insert handlers;
+
+        TDTM_Manager manager = new TDTM_Manager(namespace);
+        Set<TDTM_Global_API.TdtmToken> tSet =  new Set<TDTM_Global_API.TdtmToken>(tdtmTokens);
```
davidjray: Correct.

src/classes/TDTM_Manager.cls
```diff
@@ -0,0 +1,164 @@
...
+			        // Using Class and Target Object fields we check if the new set contains a matching token
+					TDTM_Global_API.TdtmToken match = existingToken.getmatch( newTokenSet );
+
+					// Explicitly skip over User Managed tokens
+			        if(existingToken.userManaged && match != null ) {
+			        	// Skip if User Managed
+			        	newTokenSet.remove(match);
+			        
+			        // Check if the match is different, if so we'll update the current token
+			        } else if( !existingToken.userManaged && match != null && !existingToken.equals(match) ) {
```
davidjray: You got it.

src/classes/TDTM_Manager_TEST.cls
```diff
@@ -0,0 +1,270 @@
...
+		TDTM_Manager updateManager = new TDTM_Manager('hed');
+
+		Test.startTest();
+		// Note: updateDefaultTdtmConfig will empty defaults
+		updateManager.updateDefaultTdtmConfig( newConfig );
+		Test.stopTest();
+
+		for( Trigger_Handler__c h : getDbHanders().values() ){
+			// Verify Change
+			if(h.Class__c == 'AFFL_Affiliations_TDTM'){
```
davidjray: Yes, and expanded a little to cover a few additional scenarios.

src/classes/TDTM_Config.cls
ceiroa: Not only they need to be accepted, but the need to work as expected. The decimal point shouldn't be dropped when casting to integer. That is, if we have two TDTM classes on Contact with load orders of 1.1 and 1.2, the former should run first and the latter second. We don't want them to internally become both load order 1, or the order in which they run will not be deterministic.

davidjray: Agreed. I made sure the unit tests include a float and the validation checks the value with decimal places is preserved.

davidjray:
@ceiroa the build failures were due to the "setTdtmConfig()" method I removed from the TDTM_Config class. The test class that was failing relied on the tokenCache to disable the trigger. The old setTdtmConfig() method overwrote the cache with the incoming token list. We shouldn't do this anymore, of course, because we now accept configurations from multiple namespaces. So basically, I de-coupled the setting of the config from the TDTM_Config class. 

This is ready to re-review.

ceiroa:
Looks great, @davidjray. I think the new conceptual separation between TDTM_Config, TDTM_Manager, and TDTM_Global_API is very clean now. I just have one question (in the comments), and we should be good to go after that.

davidjray:
When namespace is null the TDTM_Manager will search for Trigger handlers where Namespace is null, basically null becomes the namespace. I am taking advantage of this to configure the HEDA Trigger Handlers on when Orgs are upgraded to include these changes.

PR #444, opened 2017-06-16
Feature/acct too many rows

### LINKED ISSUE #425: In ACCT_IndividualAccounts_TDTM line 248, Too many query rows: 50001 when account has >50000 contacts
There is a line of code that is checking for a new primary contact (line 248) in:
https://github.com/SalesforceFoundation/HEDAP/blob/02b9f382bccb8d56d9e7145a0c61a3bc88f97d41/src/classes/ACCT_IndividualAccounts_TDTM.cls
~~~~
        list<Contact> listCon = [select Id, AccountId, Account.Primary_Contact__c from Contact where AccountId in :listAccountId];
~~~~
For some context, here is the method and some surrounding code:
~~~
 private void resetPrimaryContactForAccount(list<Id> listAccountId, DmlWrapper dmlWrapper) {
    
        // for each account, we need a set of their contacts.
        map<Id, set<Id>> mapAccountIdSetContacts = new map<Id, set<Id>>();
        
        list<Contact> listCon = [select Id, AccountId, Account.Primary_Contact__c from Contact where AccountId in :listAccountId];
        for (Contact con : listCon) {
            set<Id> setContactId = mapAccountIdSetContacts.get(con.AccountId);
            if (setContactId == null) { 
                setContactId = new set<Id>();
                mapAccountIdSetContacts.put(con.AccountId, setContactId);
            }
            setContactId.add(con.Id);
        }
~~~

This line queries the account for all of its contacts.  When that account has > 50000 contacts, as they sometimes due when using a "bucket" account model, this throws an exception. Too many query rows: 50001

Please consider limiting the query so that less contacts are returned.  

Currently, I have a workaround using the filters to exclude this bucket account from the trigger, but that likely wont last forever.  

Thanks!

xinnanlisf:
# Critical Changes

# Changes
Enhanced how ACCT_IndividualAccounts_TDTM.cls queries for an Account's Contacts, to prevent an error when the Account has a large number of Contacts.

# Issues Closed
Fixes #425

### DIFF
--- a/src/classes/ACCT_IndividualAccounts_TDTM.cls
+++ b/src/classes/ACCT_IndividualAccounts_TDTM.cls
@@ -294,18 +294,6 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
         // now do the update
         dmlWrapper.objectsToUpdate.addAll((List<SObject>)listAccToUpdate);
     }
-        
-    /*******************************************************************************************************
-    * @description returns the first Id from the specified set
-    * @param setId a set of Id's
-    * @return Id.  null if the set is empty.
-    */ 
-    private Id firstIdFromSet(set<Id> setId) {
-        for (ID id : setId) {
-            return Id;
-        }
-        return null;
-    }
 
     /*******************************************************************************************************
     * @description Inserts a new Account for an Individual Contact
--- a/src/classes/ACCT_IndividualAccounts_TDTM.cls
+++ b/src/classes/ACCT_IndividualAccounts_TDTM.cls
@@ -241,40 +241,56 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
     * @return void
     */ 
     private void resetPrimaryContactForAccount(list<Id> listAccountId, DmlWrapper dmlWrapper) {
-    
-        // for each account, we need a set of their contacts.
-        map<Id, set<Id>> mapAccountIdSetContacts = new map<Id, set<Id>>();       
-        for (list<Contact> listCon : [select Id, 
-                                            AccountId, 
-                                            Account.Primary_Contact__c 
-                                    from Contact 
-                                    where AccountId in :listAccountId]) {
-            for (Contact con : listCon){
-                set<Id> setContactId = mapAccountIdSetContacts.get(con.AccountId);
-                if (setContactId == null) { 
-                    setContactId = new set<Id>();
-                    mapAccountIdSetContacts.put(con.AccountId, setContactId);
+        Set<Id> setPrimaryContactIds = new Set<Id>();
+        Set<Id> setAccToResetIds = new Set<Id>();    
+        List<Account> listAccToUpdate = new List<Account>();
+
+        //Build a set of primary contact ids based on give account ids
+        List<Account> listAccount = [SELECT Primary_Contact__c,
+                                            (SELECT id
+                                            FROM Contacts
+                                            LIMIT 1)
+                                    FROM Account
+                                    WHERE Id in :listAccountId];
+        for (Account acc : listAccount){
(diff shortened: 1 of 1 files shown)


xinnanlisf:
@lianeyee 
Could you review the release notes, please? 

@davidjray 
Could you do a final review on the code, please? I have combined those two queries. Feel free to test it in your environment.

Please let me know if any questions or concerns. Thank you, both! :)

lianeyee:
@xinnanlisf The reference to the Individual Bucket model is new to me. I know we have that concept in NPSP but didn't realize that it's available in HEDA also? We probably don't need to mention it in the release notes.

That issue aside, it'd be helpful to specify when the error occurred. How about something like: Querying for an Account's Contacts no longer causes an error when an Account has more than 50,000 Contacts.

xinnanlisf:
@lianeyee 

Thank you for your reply.

I agree that we don't need to put bucket model in the release model. 

For the other one, querying account with more than 50000 contacts will still hit the limit and cause an error. We are fixing the use case when the user updates the contact and the system needs to check if the account's primary contact is still its child contact. If not, a new child contact will become that related account's primary contact.

Let me know if this makes sense or you need further explanation. :)

lianeyee:
@xinnanlisf Thanks. That's clearer, though I'm still not understanding where/when the error would be encountered. 

So before, we'd attempt to check whether the update to the contact meant that we needed to update the account's primary contact. What was it that failed? The attempt to check? The attempt to update the account's primary contact? The attempt to save the update to the contact?

xinnanlisf:
@lianeyee 

Unfortunately, we don't have many details about how to reproduce this issue.

However, I managed to recreate this issue by removing an account from a contact detail page.

lianeyee:
OK, it sounds like we don't need to be that specific then. I've edited the release note above. @xinnanlisf @jacebryan Please check when you get a chance.

jacebryan:
Looks good to me @lianeyee . Thank you!

xinnanlisf:
Thank you @lianeyee and @jacebryan!

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -241,31 +241,56 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
...
+            setAccToResetIds.addAll(listAccountId);
         }
-        
-        // now go thru each account and update their primary contact if the current one isn't valid
-        list<Account> listAcc = [select Id, Primary_Contact__c from Account where Id in :listAccountId];
-        list<Account> listAccToUpdate = new list<Account>();
-        for (Account acc : listAcc) {
-            set<Id> setContactId = mapAccountIdSetContacts.get(acc.Id);
-            if (setContactId != null && !setContactId.contains(acc.Primary_Contact__c)) {
-                acc.Primary_Contact__c = firstIdFromSet(setContactId);
```
davidjray: Since this method is not being used anymore should remove it?

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -241,31 +241,56 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
...
+
+        //if setAccToResetIds is empty, skip dml and return
+        if (setAccToResetIds.size() == 0)
+        {
+            return;
+        }
+
+        //for each account need to be reset 
+        //query one child contact and use it as primary contact
+        for (Account acc : listAccount)
```
davidjray: Good job here, I like the fact you re-used listAccount.

src/classes/ACCT_IndividualAccounts_TDTM.cls
xinnanlisf: Hey David, does red highlight mean it is removed :)

PR #446, opened 2017-06-23
Update Global API Class

davidjray:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/src/classes/TDTM_Global_API.cls
+++ b/src/classes/TDTM_Global_API.cls
@@ -202,6 +202,7 @@ global class TDTM_Global_API {
             this.targetObject = targetObject;
             this.actions = actions;
             this.loadOrderDec = loadOrder;
+            this.loadOrder = Integer.valueOf(loadOrder);
             this.active = true;
             this.async = false;
             this.userManaged = false;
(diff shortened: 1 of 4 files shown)


src/classes/TDTM_Global_API.cls
```diff
@@ -194,7 +201,7 @@ global class TDTM_Global_API {
             this.className = className;
             this.targetObject = targetObject;
             this.actions = actions;
-            this.loadOrder = loadOrder;
```
ceiroa: Don't we still need to populate the old property (loadOrder) for backwards compatibility? (In case the was any code outside HEDA retrieving loadOrder).

src/classes/TDTM_Global_API.cls
```diff
@@ -235,7 +243,7 @@ global class TDTM_Global_API {
             this.targetObject = targetObject;
             this.actions = actions;
             this.active = active;
-            this.loadOrder = loadOrder;
```
ceiroa: Same here. I think we also need to populate the old value for external code querying getting a value. (For example, if the real value is 1.2, they'll get 1. But better than getting null.)

src/classes/TDTM_Global_API.cls
davidjray: Makes sense. I'll change it,

davidjray:
Alrighty @ceiroa, ready to re-review!

PR #448, opened 2017-06-23
Fix ADDR_Contact_TDTM chunking error

### LINKED ISSUE #447: ADDR_Contact_TDTM too many chunks error
The propagateAddrInfo and createAddrsFromContacts methods of ADDR_Contact_TDTM throw a "System.TypeException: Cannot have more than 10 chunks in a single operation. Please rearrange the data to reduce chunking." in some cases. This is because contacts/accounts are being added to the objectsToUpdate list in an unsorted manner. I will submit a pull request.

kyleschmid:
Fixes a bug that was causing a too many chunks exception.

# Critical Changes

# Changes

# Issues Closed
Fixes #447.

### DIFF
--- a/src/classes/ADDR_Contact_TDTM.cls
+++ b/src/classes/ADDR_Contact_TDTM.cls
@@ -345,8 +345,10 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
         }
         //explicitly doing the update, because if we put the records in DmlWrapper we get a "duplicate id in list" error due to
         //the method updateAccountPrimaryContact in ACCT_IndividualAccounts_TDTM also putting them there.
-        update accountsToUpdate;
-        update contactsToUpdate;
+        List<SObject> objectsToUpdate = new List<SObject>();
+        objectsToUpdate.addAll((List<SObject>)accountsToUpdate);
+        objectsToUpdate.addAll((List<SObject>)contactsToUpdate);
+        update objectsToUpdate;
     }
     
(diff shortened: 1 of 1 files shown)


src/classes/ADDR_Contact_TDTM.cls
```diff
@@ -326,25 +327,26 @@ public class ADDR_Contact_TDTM extends TDTM_Runnable {
...
                 originalContact.Current_Address__c = acc.Current_Address__c;
-                objectsToUpdate.add(originalContact);
+                contactsToUpdate.add(originalContact);
             }
         }
         //explicitly doing the update, because if we put the records in DmlWrapper we get a "duplicate id in list" error due to
         //the method updateAccountPrimaryContact in ACCT_IndividualAccounts_TDTM also putting them there.
-        update objectsToUpdate;   
+        update accountsToUpdate;
+        update contactsToUpdate;
```
ceiroa: The PR makes a lot of sense, @kyleschmid. Good catch. And thank you for your contribution. 

Now that we have accounts and contacts in two separate lists, would you mind combining them into one for the update? The order should be preserved, and switching from one object type to another should occur only once. Thanks!

kyleschmid: Will do

ceiroa:
I see the new PR, but just fyi, we could have continued working with this one. The latest commit became part of it too. Either way works, but it's a little less work for you. :-)

PR #456, opened 2017-07-04
Update TDTM token sorting

### LINKED ISSUE #455: Admin Account is not created when a Contact is created
Introduced from recent TDTM updates. 

TDTM_Config.getClassesToCallForObject() is sorting the tokens via a FOR loop with an integer key. Now that we are using DECIMALS for the load order field we cannot access the tokens via the for loop integer key.

davidjray:
# Critical Changes

# Changes

# Issues Closed
Fixes #455

### DIFF
--- a/src/classes/TDTM_Config.cls
+++ b/src/classes/TDTM_Config.cls
@@ -66,7 +66,7 @@ public class TDTM_Config {
         // Getting the default configuration only if there is no data in the Trigger Handler object. Otherwise
         // we would delete customizations and Trigger Handlers entries that aren't in the default configuration.
         if(tdtmConfig.size() == 0) {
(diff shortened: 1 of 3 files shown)


davidjray:
Since we recently switched to Decimals for the sort order of TDTM Trigger Handlers the sorting mechanism was broken. I implemented the comparable interface so the tokens could be easily sorted. I updated TDTM_DefaultConfig to include a decimal value for the 'ACCT_IndividualAccounts_TDTM' class trigger handler, this will actually cause a failure in the unit tests moving forward.  

To test:
1. Deploy master to a new scratch org and create a new contact. You will see the Admin Account is not created.
2. Deploy this branch, and create another contact. You will see the admin account is now created.

paulmalenke:
Deployed and verified fix in a scratch org.

PR #460, opened 2017-07-12
Feature/update address type

### LINKED ISSUE #394: Fix ADDR_Addresses_TDTM.cls to update Primary Address Type on Contact
From PowerofUS Hub Ideas Exchange: https://powerofus.force.com/hub-ideas/1166
[ADDR_Addresses_TDTM.zip](https://github.com/SalesforceFoundation/HEDAP/files/790975/ADDR_Addresses_TDTM.zip)

I have been setting up a new org that has the HEDA model installed. I noticed that when an address is added to the Address table (from the Contact record) and it is marked as the Default address, that address (street, city, state, zip and country) is updated on the Contact record itself in the Mailing address fields, BUT the Primary Address Type field is not updated to reflect the address type of the new address. I've looked at the code and I think this is happening:

The ADDR_Addresses_UTL.cls has code called copyAddressAddrSObj that is called to copy Address Fields from an Address object to a Contact or Account. When it is called, it's supposed to have some parameters passed into it.

public static void copyAddressAddrSObj(Address__c addr, SObject sobjDst, string strFieldPrefix, string strFieldAddrType, 
TDTM_Runnable.DmlWrapper dmlWrapper)

You'll see that the second to last parameter is a string field called strFieldAddrType. This is "an optional Address Type field on sobjDst to copy to" according to the comments for this code. The sobjDst is the name of the object that will be updated (Contact or Account).

Then inside this code where it is actually updating the Contact (or Account) record, there is this...
if (strFieldAddrType != null)
sobjDst.put(strFieldAddrType, addr.Address_Type__c);

So if there is a value for strFieldAddrType, then the the Address type field (specified in strFieldAddrType) will be updated with the value of Address_Type__c from the Address record.

This function is called in the ADDR_Addresses_TDTM.cls in the section called updateParentsAddresses. Inside updateParentsAddresses there is a line that calls the copyAddressAddrSObj. This is line 416 in the code.

ADDR_Addresses_UTIL.copyAddressAddrSObj(addr, contact, 'Mailing', null, dmlWrapper);

I believe this line needs to be changed to:

ADDR_Addresses_UTIL.copyAddressAddrSObj(addr, contact, 'Mailing', 'Primary_Address_Type__c', dmlWrapper);

xinnanlisf:
# Critical Changes

# Changes
When you add an address for a Contact and indicate that it's the Default Address, we now update the Contact's Primary Address Type field based on the Address Type of the newly added address (Home, Work, etc.).
# Issues Closed
Fixes #394


xinnanlisf:
@lianeyee 

Could you review the release note, please? Let me know if any questions. :)

lianeyee:
Thanks, @xinnanlisf. I'd like to see a bit more detail in the release note, to be clearer about the context for when this happens, and to specify the fields involved (since there are a few). Address stuff is typically confusing for customers, so I tend to be more explicit. Moving the last part to the beginning ("When adding . . .) helps set the context. Does my proposed revision work?

xinnanlisf:
@lianeyee that makes a lot of sense. Thank you for the insights!

PR #463, opened 2017-07-13
Prevent duplicate trigger handlers and increase unit test scenarios

### LINKED ISSUE #461: Possible duplicate trigger handlers created upon package upgrade after 1.37
In certain cases, trigger handlers could be deleted or duplicated when HEDA is upgraded.

The recent push of HEDA 1.37 may have deleted users custom trigger handlers when "User Managed" wasn't checked. 

We should update all Trigger handlers in the HEDA scope to use the "Owned by namespace" field before processing so HEDA Handlers are targeted specifically in future updates, and Users custom trigger handlers are not touched.

Also, future upgrades to HEDA 1.38 and above could create duplicate Trigger Handlers when Admins set the  "user Managed" field to true.

davidjray:
With this update, HEDA will update the Trigger handlers it owns to populate the Owned by Namespace field. This will isolate the HEDA Trigger Handlers from others. This will prevent users custom trigger handlers from being deleted on HEDA upgrades in the future.

Since the TDTM_Global_API includes a method that does not require a namespace the usage of Owned_by_Namespace__c is the preferred method to keep trigger handlers from being deleted.

# Critical Changes

# Changes

# Issues Closed
Fixes #461


ceiroa:
Do we need something like what you have in the ticket added to the Changes section:

"There are now two ways to prevent deletion of trigger handlers. You can use the User_Managed__c field or the Owned_by_Namespace__c field. The usage of Owned_by_Namespace__c is the preferred method."

@lianeyee

davidjray:
I was thinking about it, and technically those changes we part of the previous PR, I was just clarifying that point in my comment on the issue. This PR focuses mostly on preventing the duplicates, editing the HEDA trigger handlers namespace field, and covering addition cases in testing.

ceiroa:
Let's add the Owned by Namespace field to the "All - Main Fields" list view:
screen shot 2017-07-14 at 2 22 12 pm

Also, let's move that field to the right side on the details view, to have the same number of fields in each column:
screen shot 2017-07-14 at 2 23 50 pm

Thanks!

src/classes/STG_InstallScript_TEST.cls
```diff
@@ -107,13 +119,89 @@ public with sharing class STG_InstallScript_TEST {
...
+                // Verify Admin overrides
+                System.assertEquals(false, h.Active__c);
+                System.assertEquals(true, h.User_Managed__c);
+            }
+        }
+
+        Test.startTest();
+
+        //Run the install script
+        Test.testInstall(new STG_InstallScript(), new Version(1,38), true);
```
ceiroa: Is it really useful to run the install script twice in this test? The new code in the install script runs for any version lower than 1.38. What about running the install script just once, but also adding the verification of the custom handler? It would make the test a little easier to understand.

davidjray:
@ceiroa I added those changes!

src/classes/STG_InstallScript_TEST.cls
```diff
@@ -96,7 +97,18 @@ public with sharing class STG_InstallScript_TEST {
...
+        // Simulate a user adding their own handler, should not need User_Managed__c to be true
+        Trigger_Handler__c newHandler = new Trigger_Handler__c(
+            Active__c = true,
+            Asynchronous__c = false,
+            Class__c = 'Contact_Update',
+            Load_Order__c = 0.2,
+            Object__c = 'Contact',
+            Trigger_Action__c = 'BeforeInsert;BeforeUpdate'
+        );
+        insert newHandler;
```
davidjray: They are held in memory but not persisted to the database.

src/classes/STG_InstallScript_TEST.cls
davidjray: I am trying to simulate the actual process an Org will go through with this update. I am trying to test for future upgrades as well.

src/classes/STG_InstallScript_TEST.cls
```diff
@@ -107,13 +119,89 @@ public with sharing class STG_InstallScript_TEST {
...
+        Test.stopTest();
+
+        List<Trigger_Handler__c> tdtmHandlers2 = [SELECT Id, Name, Owned_by_Namespace__c, Active__c, User_Managed__c, Class__c FROM Trigger_Handler__c];
+
+        // Verify the number of default records didn't change.
+        System.assertEquals(TDTM_DefaultConfig.getDefaultRecords().size(), tdtmHandlers2.size()-1);
+
+        for(Trigger_Handler__c h: tdtmHandlers2){
+            if(h.Class__c != 'Contact_Update') {
+                // Unfortunantly the user overwrote them all, so we cannot edit it to add the namespace field
```
davidjray: Good catch

src/classes/TDTM_Manager.cls
```diff
@@ -40,6 +40,9 @@ public class TDTM_Manager {
 	public Map<Id,Trigger_Handler__c> currentHandlers {get;set;}
```
davidjray: Agreed, will do.

src/classes/STG_InstallScript_TEST.cls
davidjray: I added an assert to verify this.

lianeyee:
@davidjray I see that you and @ceiroa  considered whether to mention this change in Changes (and decided not to). Given that this PR adds Owned by Namespace to the Trigger Handlers list view, and that trigger handlers have been on people's radar since 1.37, I'm wondering we should reconsider and mention it after all?

davidjray:
I don't think it will hurt to add something @lianeyee  stating it was added to the layout and list view, so current Orgs will see they'll want to add it. I all for adding to the release notes!

lianeyee:
Thanks @davidjray. Please review the newly added Changes. We'll need Chris to regenerate the RNs.

davidjray:
Looks perfect @lianeyee!

PR #477, opened 2017-08-24
Feature/xinnanli/localize settings page

xinnanlisf:
# Critical Changes

# Changes
Replaced all labels in HEDA Settings with custom labels to facilitate language translations. 
# Issues Closed

### DIFF
--- a/src/labels/CustomLabels.labels
+++ b/src/labels/CustomLabels.labels
@@ -580,7 +580,7 @@
         <fullName>stgDefaultFacultyTypeTitle</fullName>
         <categories>Settings</categories>
         <language>en_US</language>
-        <protected>true</protected>
+        <protected>false</protected>
         <shortDescription>stgDefaultFacultyTypeTitle</shortDescription>
         <value>Default Faculty Record Type:</value>
     </labels>
@@ -588,7 +588,7 @@
         <fullName>stgDefaultStudentTypeTitle</fullName>
         <categories>Settings</categories>
         <language>en_US</language>
-        <protected>true</protected>
+        <protected>false</protected>
         <shortDescription>stgDefaultStudentTypeTitle</shortDescription>
         <value>Default Active Student Record Type:</value>
     </labels>
@@ -612,7 +612,7 @@
         <fullName>stgDisableErrorHandlingTitle</fullName>
         <categories>Settings</categories>
         <language>en_US</language>
-        <protected>true</protected>
+        <protected>false</protected>
         <shortDescription>stgDisableErrorHandlingTitle</shortDescription>
         <value>Disable Error Handling:</value>
     </labels>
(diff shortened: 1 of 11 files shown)


xinnanlisf:
@smarovich 
Hi, Sarah, would you mind reviewing this pull request, please? Thanks!

lianeyee:
@xinnanlisf I'm tempted to say something like "Replaced all labels in HEDA Settings with custom labels to facilitate language translations." However, I'm not sure that it's accurate to say "all labels." Are there any exceptions for certain strings?

xinnanlisf:
@lianeyee 
Sorry for the confusions. No, we are replacing all the hard coded strings in the settings page to support future implementation of translations.

And I think your release note looks great!

Let me know if it makes sense! :D

lianeyee:
Thanks for confirming, @xinnanlisf. You got us started with a great draft note. I've updated Changes.

src/labels/CustomLabels.labels
```diff
@@ -329,18 +665,26 @@
...
+        <protected>true</protected>
+        <shortDescription>stgHelpAccoutsDeletedIfChildContactsDeleted</shortDescription>
+        <value>Record Types of those Accounts that will be automatically deleted if all their children Contacts have been deleted.</value>
+    </labels>
+    <labels>
         <fullName>stgHelpAddressAccRecType</fullName>
         <categories>Settings</categories>
         <language>en_US</language>
-        <protected>false</protected>
+        <protected>true</protected>
```
davidjray: I am curious, why are the labels being changed to protected?

xinnanlisf: @davidjray 
Basically I just manually changed them in the xml file. Looks like I accidentally changed existing custom labels as well. Should I change them back?

davidjray: I'm not sure. I think we should have a standard we follow. @ceiroa what are your thoughts?

ceiroa:
New labels should be created protected so we can delete them later if we want to. If they aren't protected they cannot ever be removed from the package, as there might be code in the org that references them. Unfortunately the packaging org won't let you change a non-protected label to protected, for the same reason.

src/labels/CustomLabels.labels
```diff
@@ -284,35 +580,67 @@
         <fullName>stgDefaultFacultyTypeTitle</fullName>
         <categories>Settings</categories>
         <language>en_US</language>
-        <protected>false</protected>
```
davidjray: I agree with @ceiroa, that it is a good idea to make all new labels protected. Unfortunately, we cannot change existing labels due to packaging restrictions. So we'll need to change these back.

xinnanlisf: @davidjray 

I see I will change those back, thanks!

xinnanlisf:
@davidjray should be good now!

fyi, @ceiroa

davidjray:
@sreesureka and @chrismcmahon this PR is ready for QA!

davidjray:
@chrismcmahon I see you were assigned. The best way to test this is to complete some exploratory testing in the HEDA settings page. Visiting each tab and making some arbitrary changes.

chrismcmahon:
Checked HEDA Settings, all the text looks correct and Settings save properly

Issue #465, opened 2017-07-18
TDTM Filtering not working correctly when using Filter Field of Boolean type


jacebryan:
**Issue identified by** [Brian Petersen (Queen's School of Business) ](https://powerofus.force.com/0058000000ABNmU) on July 13, 2017
**PUB Thread**: https://powerofus.force.com/_ui/core/chatter/groups/GroupProfilePage?g=0F980000000D6uj&fId=0D58000003VpuaU

**Issue**: TDTM Filtering not working correctly when using Filter Field of Boolean typeValue

**Steps to reproduce**:

1. Create a custom Apex Class that updates a Contact field value if some condition is met.
- Example: IF Military Service == True then add "Please Contact VA Services" in the Military Background field

2. Create a new Trigger Handler that references the following:
- Class = class created in Step 1
- Object = Contact
- Trigger Action = Before Insert; Before Update
- Filter Field = ex: hed__Deceased__c
- Filter Value = ex: 'True' OR 'true'

3. Test the Class and Trigger Handler by Creating a new Contact
- Create a new Contact and set the value of hed_Military_Service__c to TRUE
- Save the record
- The hed_Military_Background__c field now has the value of "Please contact VA Services"

_Note_: The Trigger Handler indicates that unless the Filter Field (i.e. hed_Deceased__C) and Filter Value (i.e True or true) are met then the Class referenced will not be called.

**The class should not be called unless the Trigger Handler condition is TRUE.**

davidjray:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007O6ofAAC" target="_blank">W-019578</a>

ceiroa:
@weizsforg Would you mind adding some details about what you found to be the actual problem? Thank you!

weizsforg:
I did a investigation in one of the HEDA trial org and I found the following: 
1. CheckBox (HEDA field): **doesn't work**
2. Picklist (HEDA field):**doesn't work** 
3. Text (HEDA field):**doesn't work** 
4. CheckBox (HEDA field without namespace: For example user put Military_Background__c but not hed__Military_Background__c): **works**
5. CheckBox (user's custom field):**doesn't work** 
6. Picklist (user's custom field):**doesn't work** 
7. Text (user's custom field):**doesn't work** 
8. Text (standard, like LastName): **works**
9. Checkbox (standard, like Email Opt Out): **works**

In summary:
[...]

PR #502, opened 2017-10-19
TDTM Filtering not working correctly when using Filter Field of Boolean type

weizsforg:
# Critical Changes

# Changes
Fixed an issue that prevented the Trigger Handler from filtering properly by field values in Filter Field and Filter Value. 

# Issues Closed
Fixes: #465 

# New Metadata

# Deleted Metadata

# Testing Notes
For this PR, we need to test the following scenarios for the 'Filter Field' setting:
	1.	Standard Field (such as LastName)
	2.	Custom field in HEDA package (such as Deceased__c)
	3.	Custom field from unmanaged object (such as user's custom field MyField__c)
	4.	Custom field from another package (such as third party field tfp__TFCheckBox__c)
        5.     Relationship field (such as Account.Name, Current_Address__r.Parent_Account__r.RecordType.Name)

Recommended Steps to Test this PR:
1.	Deploy the HEDA code base to a Developer org and generate a Beta package. For QA, following the normal release process to get a Beta org
2.	Generate another Beta package with a different namespace to simulate third party package (or use this link to install one https://login.salesforce.com/packaging/installPackage.apexp?p0=04t1I000001mTMo)
3.	When both HEDA and third party package get installed, make sure the fields you need are in the object's page layout
4.	Create a simple Apex class that do a simple update of the record. For example, update Contact's "Military Background" to "Please Contact VA Services"

```
global class MyTDTM extends hed.TDTM_Runnable {
    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist, 
        hed.TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
		if(newlist != null) {
		    for (integer i=0; i<newlist.size(); i++) {
		         Contact c = (Contact)newlist[i];
		         c.hed__Military_Background__c = 'Please Contact VA Services';
		    }
		}
		return new DmlWrapper();
	}
}
```

5.	Create a Trigger Handler record for the Apex class that you created before. 
                For example: 
                Class = class created in Step 4 
                Object = Contact
                 Trigger Action = Before Insert; Before Update 
                 Filter Field = LastName     
                 Filter Value = aaa

6.	Confirmed that the Contact with last name = "aaa" will not have Military Background = "Please Contact VA Services" after you try to Save the Contact.
7.	Now, update the Filter Field and Field Value to test all 4 scenarios mentioned above with different field type (Checkbox, Picklist, Text). Note: if you install the package with the link in step2, you should have three fields on Contact available for you (tfp__TFCheckBox__c, tfp__TFPicklist__c, tfp__TFText__c).

Example Scenarios to Test this PR:
Basically this new proposal include two parts. 
1. We will implement a validation when user enter the “Filter Field” in Trigger Handler. The validation is to prevent user enter the field with namespace. 
2. We will also fix the filter logic to let user use their own field. 

Say, we have those fields on Contact object in an org:
Standard Field: LastName
HEDA Field: Deceased__c
Non HEDA Field: MyField__c 
Field from another package: tfp__TFCheckBox__c
Non HEDA Object: MyObject__c

(Filter Field Validation):
LastName : Valid
Deceased__c: Valid
hed__Deceased__c: NOT Valid
MyField__c: Valid
tfp__TFCheckBox__c: Valid
MyObject__r.MyObjectField__c Valid

(Actual Filter logic):
LastName : Works
Deceased__c: Works
hed__Deceased__c: Doesn't Work
MyField__c: Works:
tfp__TFCheckBox__c: Works
MyObject__r.MyObjectField__c Works

* Special case: Create a field in the org that has the same api name as HEDA field. For example, create a new field call Deceased in the org (the API name should be Deceased__c). Put Deceased__c into the Filter Field in the Trigger Handler. You should see an error message after you click Save. Check if the message make sense and following the message to Save the record again. 

Please note that the scenarios here only cover the checkbox field type. Please try different field type (Checkbox, Picklist, Text) as mentioned in step 7.


weizsforg:
Hi @lianeyee, can you please review this PR? Let me know if I you need something.

lianeyee:
Hi @weizsforg. I edited the notes slightly. Please check. Thanks!

ceiroa:
Hi @weizsforg. Thanks for making me a reviewer in this PR. I will review the code, and then I'll assign it to @xinnanlisf for actual deployment and testing. Would you mind adding a "Testing Notes" sections with details on how to test these changes? I have added this section to the PR template, to match the SAL template, but the PR hasn't been merged yet (https://github.com/SalesforceFoundation/HEDAP/pull/503). Thanks!

weizsforg:
Hi @ceiroa , Thanks. I will add the "Testing Notes" section in the comment.

src/classes/THAN_Filter_TDTM.cls
```diff
@@ -61,12 +61,16 @@ public class THAN_Filter_TDTM extends TDTM_Runnable {
...
-		                        String dynamicQuery = 'select ' + filterField + ' from ' + obj + ' limit 1';
-		                        List<SObject> records =  Database.query(dynamicQuery);
-		                    } catch(QueryException e) {
-		                        handler.addError(Label.InvalidFilter + handler.Class__c);
-		                    }
+                            try {
+                                //Note: We should use schema class to check if the field is valid for an object since dynamic query will ignore namespace automatically.
+                                SObjectField field = Schema.getGlobalDescribe().get(obj).getDescribe().fields.getMap().get(filterField);
+                                //Note: System will automatically remove HEDA namespace of client's custom field. For example, client create a custom field ABC__c, and put hed__ABC__c into the setting. The field value will not be null but ABC__c. We do not want user to accidentally add namespace into their own field.
+                                if(field == null || String.valueOf(field) != filterField) {
```
ceiroa: Thanks for the comment in line 67. I was going to ask you about the second condition of the if statement, but now I don't need to. :-)

ceiroa: Now that there's no query, do we still need the try-catch? No harm in having it, but it's possible nothing has now the potential of throwing an exception in this section, and we would avoid having the handler.addError line twice. Just a thought.

lianeyee:
Carlos' last comment makes me think that we can make the Critical Changes note clearer, @weizsforg. i.e., "With this release, check your existing Trigger Handlers that reference HEDA fields in the Filter Field. Verify that you've specified the full field name, including the HEDA namespace prefix (for example, hed__Deceased__c)."

src/classes/THAN_Filter_TDTM.cls
weizsforg: Yes, I think it is correct. I just did a little bit test, looks ok. So, I just removed it.

ceiroa:
Awesome. It looks all good to me now, @weizsforg. The only issue is the build failure. I see one build passed and one failed. Is that one failing the browser tests or something else?

weizsforg:
@ceiroa Yes, looks like that is a new Feature Test added just today. Not sure what it is. I already asked @cpolcyn about it. Maybe he knows.

cpolcyn:
@weizsforg yes we had a bug in mrbelvedereci that caused the namespaced version of the tests to fail. You can see in the most recent commit here that it has been fixed.

ceiroa:
@weizsforg I have removed myself from the PR Asignee field to avoid confusion. Please re-add me once the PR is ready. BTW, I think the notes need updating.

weizsforg:
Hi @lianeyee , there are some updates of the design for this PR.  One of the change is that we are going to add a new error message when user enter the invalid value in "Filter Field" in Trigger Handler record. 

The use case is that: 
User create a custom field Deceased__c and then put it into the "Filter Field"and try to save the Trigger Handler record. Our logic found that HEDA also has Deceased__c field. In this case, we will display an error on the page layout says:

“Looks like this field is duplicate with one of the HEDA field. Please note that system will use HEDA field as the Filter. If you would like to proceed, please add (ProceedDuplicate) after your Filter Field and click Save. For example: put Deceased__c(ProceedDuplicate) as Filter Field."

So, once user put (ProceedDuplicate) and click Save, system will save the record without throwing the error.  Please review the error message and the string we ask user to add behind the filter field string "(ProceedDuplicate)".  Please let me know if you have any question about it.

lianeyee:
Hi @weizsforg. See if this wording for the error message and the string make sense?

“The Filter Field value names a duplicate of a HEDA field. We will filter according to the value in the HEDA field. To proceed with saving this record, either specify a non-duplicate field or acknowledge that you accept this filtering logic by appending (DuplicateAcknowledged) after the field name. For example, if the duplicate field is Deceased__c, enter Deceased__c(DuplicateAcknowledged) as the Filter Field value."

weizsforg:
@lianeyee Thanks

weizsforg:
Hi @ceiroa , this PR is ready for review now. 

This PR is about fixing the Filter Validation and Filter Logic to let user use their own field or third party package field. It also includes the error message for the shadowing field issue in the filter. 

I did not put the Critical Changes in this PR since we are going to preserve all existing filter logic and just fix the Filter Logic for user's field and third party field. Please let me know if there is any concern about it.

src/classes/TDTM_Filter.cls
```diff
@@ -100,6 +100,28 @@ public class TDTM_Filter {
...
     	return null;
     }
+
+    /*******************************************************************************************************
+    * @description apply the correct namespace of the field
+    * @param filterField the name of the filter
+    * @return String the filter field with correct namespace based on existing filter logic
+    */
+    public string handleNamespace(string filterField) {
+        string theFilter = '';
```
weizsforg: upper case of the class

src/classes/THAN_Filter_TDTM.cls
```diff
@@ -61,12 +61,44 @@ public class THAN_Filter_TDTM extends TDTM_Runnable {
...
+
+                            //Get the ignoreDuplicateField flag and remove it from the Filter Field value
+                            if(ignoreDuplicateField) {
+                                filterField = filterField.replace(Label.duplicateFieldInFilter, '');
+                                handler.Filter_Field__c = handler.Filter_Field__c.replace(Label.duplicateFieldInFilter, '');
+                            }
+
+                            //System should't allow a Filter Field with HEDA namespace
+                            if(hasHEDAPrefix) {
+                                handler.addError(Label.InvalidFilter + handler.Class__c);
```
weizsforg: Use another message let user know they need to remove heda prefix

src/classes/THAN_Filter_TDTM.cls
```diff
@@ -61,12 +61,44 @@ public class THAN_Filter_TDTM extends TDTM_Runnable {
...
+                            } else {
+                                Boolean fieldExists = false;
+                                Boolean fieldExistsHEDA = false;
+
+                                //First, test if the field exists outside HEDA 
+                                fieldExists = Schema.getGlobalDescribe().get(obj).getDescribe().fields.getMap().keySet().contains(filterField.toLowerCase());
```
weizsforg: Limit of the describe call?

src/classes/THAN_Filter_TDTM.cls
```diff
@@ -61,12 +61,44 @@ public class THAN_Filter_TDTM extends TDTM_Runnable {
...
+
+                                //Second, add the HEDA namespace and test if the field exists in HEDA
+                                filterField = UTIL_Namespace.StrTokenNSPrefix(filterField);
+                                fieldExistsHEDA = Schema.getGlobalDescribe().get(obj).getDescribe().fields.getMap().keySet().contains(filterField.toLowerCase());
```
weizsforg: similar

weizsforg:
Hi @lianeyee . I think we need another error message for this PR.

The use case is that if user put a field as the "Filter Field" and the field has HEDA namespace. We need to tell them not to include the namespace in the field. So, the message will be something like: 

"The Filter field should not include HEDA namespace. Please remove HEDA namespace and try again."

lianeyee:
How's this, @weizsforg? "The Filter Field value can't include the HEDA namespace. Remove it from the field name."

weizsforg:
Looks great! @lianeyee

weizsforg:
Hi @ceiroa , I added a new Label to tell user remove the heda namespace when entering the Filter Field. Also, I switched the global describe to use UTIL_Describe in Filter Validation.  Please have a look.

src/classes/THAN_Filter_TDTM.cls
```diff
@@ -61,12 +61,47 @@ public class THAN_Filter_TDTM extends TDTM_Runnable {
...
+                            //4. Field from another package with namespace e.g. NS__Field__c
+
+                            String hedaPrefix = String.isBlank(UTIL_Namespace.getNamespace()) ? '' : (UTIL_Namespace.getNamespace() + '__');
+                            Boolean hasHEDAPrefix = filterField.startsWithIgnoreCase(hedaPrefix) ? true : false;
+                            Boolean ignoreDuplicateField = filterField.endsWithIgnoreCase(Label.duplicateFieldInFilter) ? true: false;
+
+                            //Get the ignoreDuplicateField flag and remove it from the Filter Field value
+                            if(ignoreDuplicateField) {
+                                filterField = filterField.replace(Label.duplicateFieldInFilter, '');
+                                handler.Filter_Field__c = handler.Filter_Field__c.replace(Label.duplicateFieldInFilter, '');
```
ceiroa: Do we need line 77? Are we using handler.Filter_Field__c anywhere below?

src/classes/THAN_Filter_TDTM.cls
```diff
@@ -61,12 +61,47 @@ public class THAN_Filter_TDTM extends TDTM_Runnable {
...
+                            }
+
+                            //System should't allow a Filter Field with HEDA namespace
+                            if(hasHEDAPrefix) {
+                                handler.addError(Label.InvalidFilter + handler.Class__c + '. ' + Label.fieldWithHEDANamespace);
+                            } else {
+                                Boolean fieldExists = false;
+                                Boolean fieldExistsHEDA = false;
```
ceiroa: Not a big deal, but these two variables could be set the first time they are used, without initializing them first (we could just remove lines 84 and 85).

src/classes/THAN_Filter_TDTM.cls
weizsforg: I think we need line 77 because this is the value that we are going to save to database. the filterField is just a variable in this method.

src/classes/THAN_Filter_TDTM.cls
weizsforg: right

src/classes/THAN_Filter_TDTM.cls
ceiroa: Ah right, because we are running this on Before actions that run on Trigger_Handler__c. Never mind. :-)

src/classes/THAN_Filter_TDTM.cls
```diff
@@ -61,12 +61,45 @@ public class THAN_Filter_TDTM extends TDTM_Runnable {
 	                    String filterField = String.escapeSingleQuotes(handler.Filter_Field__c);
 	                    String obj = String.escapeSingleQuotes(handler.Object__c);
 	                    if(!String.isBlank(filterField) && !String.isBlank(obj)) {
-		                    try {
-		                        String dynamicQuery = 'select ' + filterField + ' from ' + obj + ' limit 1';
-		                        List<SObject> records =  Database.query(dynamicQuery);
-		                    } catch(QueryException e) {
-		                        handler.addError(Label.InvalidFilter + handler.Class__c);
-		                    }
+                            //those four type of fields are considered valid Filter field
```
weizsforg: Some issues here:
1. The relationship field does not work here, need to consider that
2. If we use describe method here, then the object name must have namespace, but it is not true for the existing data
So, might need to reconsider the dynamic query here.

chrismcmahon:
Wei and I discussed this, and our plan for testing is to merge the PR in order to get an installable beta build, test in the real beta, and be prepared to revert the merged PR if we discover any issues. We should probably tell Chris Polcyn what we're doing first.

src/classes/THAN_Filter_TDTM.cls
weizsforg: 1. Fixed relationship validation and logic
2. Fixed the namespace issue for single field in validation 
3. used different validation on relationship and single field

src/classes/TDTM_Filter.cls
```diff
@@ -100,6 +100,28 @@ public class TDTM_Filter {
...
+
+    /*******************************************************************************************************
+    * @description apply the correct namespace of the field
+    * @param filterField the name of the filter
+    * @return String the filter field with correct namespace based on existing filter logic
+    */
+    public String handleNamespace(string filterField) {
+        String theFilter = '';
+        String hedaPrefix = String.isBlank(UTIL_Namespace.getNamespace()) ? '' : (UTIL_Namespace.getNamespace() + '__');
+        //if the filter has the heda namespace, then we need to add heda namespace again (This is for backward compatibility. Previously, we allow user put full name include the heda namespace, but the filter does not work)
```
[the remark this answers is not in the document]
weizsforg: Yes, for the existing Filter Field that has the HEDA namespace, we add the namespace again so that it doesn't work.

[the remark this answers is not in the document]
ceiroa: Great. Mind saying in the comments that adding the namespace a second time will break the filtering? I just want to be super clear so next time we look at it a few months/years from now we aren't confused. Thank you.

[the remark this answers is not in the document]
weizsforg: Sure. I can update the comment. Thanks.

ceiroa:
Hi @chrismcmahon ~ I'm going to be assigning this PR to you. I have reviewed all code changes in detailed, deployed to a clean scratch org, and run all tests there. But as you can see above, testing these fixes is an elaborate process that includes creating a separate package from a dev org with a different namespace. Please go through all the steps outlined above by @weizsforg for testing, and reach out to him if you have any questions.

Also, please write down the steps you have taken for testing and the scenarios you have covered and get approval from me and Wei **before merging**. Thank you.

weizsforg:
Hi @chrismcmahon , just as @ceiroa mentioned above, we need to test all the scenarios before merge this PR. So, that means we need to use this branch to create a Beta package and install it into a scratch org and do the test. Let me know if you need me to set up the test environment for you. Thanks

PR #514, opened 2017-11-09
Fixed TDTM_Global_API setTdtmConfig method. 1. setTdtmConfig will clear the cached Tdtm tokens 2. setTdtmConfig will save Filter Field and Filter Value

### LINKED ISSUE #467: TDTM Manager mergeToken method excludes Filter Field and Filter Value fields
Issue identified by [Brian Petersen (Queen's School of Business) ](https://powerofus.force.com/_ui/core/userprofile/UserProfilePage?u=0058000000ABNmU) 
PUB Thread: https://powerofus.force.com/_ui/core/chatter/groups/GroupProfilePage?g=0F980000000D6uj&fId=0D58000003WhXZw

Issue: When using the TDTM_Global_API.setTdtmConfig method to manage TDTM configuration the Filter Field and Filter Value fields are not set.

Steps to reproduce:
1. Run TDTM_Global_API.setTdtmConfig in the anonymous apex window with a configuration that includes a value for the Filter Field and Filter Value fields in some of the items.
2. After running the code note the values in the Filter Field and Filter Value fields were not copied over.

As stated in the PUB post by Brian, the root of the problem is most likely the TDTM_Manager.mergeToken method excluding those fields.

### LINKED ISSUE #479: TDTM_Global_API getTdtmConfig does not return customized TDTM configuration in unit tests
If a developer is running custom unit tests that interact with HEDA objects, the TDTM configuration is able to return the default trigger handler configuration without requiring any Trigger_Handler__c setup records. The configuration is built on-the-fly.

However, if a developer wants to customize the Trigger Handler setup, they can do so via TDTM_Global_API.setConfig. Within the same transaction, though, the getTdtmConfig will still return the HEDA default configuration. This can have strange results: records inserted during testSetup will  use the default HEDA TDTM configuration but the actual test method will use the custom TDTM config (since the test method starts another transaction). 

getTdtmConfig stores and returns a cached version of the configuration and there's not a simple way to clear it and pull the customized version within the same transaction.

weizsforg:
# Critical Changes

# Changes
Added a new TDTM class, THAN_ClearCache_TDTM, that clears cached TDTM tokens after Trigger Handler record insert. Enhanced the setTdtmConfig method to clear cached TDTM tokens and to save values for Filter Field and Filter Value. Verify that your Apex unit tests and trigger handlers work as expected with these changes to setTdtmConfig.

# Issues Closed
Fixes #467
Fixes #479

# New Metadata
THAN_ClearCache_TDTM
THAN_ClearCache_TEST

# Deleted Metadata

# Testing Notes
Please note that this PR includes the fix for two GitHub issues (#467 and #479).  Those two issues are related to the TDTM_Global_API that will be used in Apex unit test. Those two issues are not directly related to the Filter logic in TDTM framework. 

#467 : The setTDTMConfig method missed the Filter Field and Filter Value when save the tokens into the database.

Steps to reproduce:
1. Run TDTM_Global_API.setTdtmConfig in the anonymous apex window (or in the Apex unit test) with a configuration that includes a value for the Filter Field and Filter Value fields in some of the items.

```
List<TDTM_Global_API.TdtmToken> tokens = new List<TDTM_Global_API.TdtmToken>();
tokens.add(new TDTM_Global_API.TdtmToken('Filter_Test_TDTM', 'Contact',  'BeforeInsert', 1, true, false, 'LastName', 'Testerson', false,  ''));
TDTM_Global_API.setTdtmConfig(tokens);
```

2. Do a query about the Trigger_handler__c record **or look at the Trigger Handlers tab**. You will find that the Filter Field and Filter Value is not in the database. 
3. The root cause is that TDTM_Manager.mergeToken method excluding those fields.

#479 :  The getTDTMConfig always return the cached TDTM tokens.

Steps to reproduce:
1. Insert a Trigger_Handler__c record (for example update a field on Contact when insert Contact in Apex unit test). The insertion operation will cause system to populate and cache the TDTM Tokens with default config. 
2. Insert a Contact record **in the same transaction**. You will find that the new Trigger Handler didn’t run. This is because the system uses the cached TDTM tokens from step 1 and the cache does not include the new Trigger_Hanlder__c record. 
3. When you call ```TDTM_Global_API.getTdtmConfig``` method, it will always return the cached token from step 1 unless you manipulate the cached token directly. 

Another way to reproduce **- do all these steps in the same transaction (anonymous Apex or unit test)**:
1. Insert a Contact record or call ```TDTM_Global_API.getTDTMConfig```. The operation will cause system to populate and cache the TDTM Tokens with default config. 
2. Build a TDTM Token
3. Set the new Token
4. When you call ```TDTM_Global_API.getTDTMConfig```, you will get the cached token from step 1 instead of the new one inserted in step 2:
  
```
TDTM_Global_API.getDefaultTdtmConfig();
List<TDTM_Global_API.TdtmToken> tokens = new List<TDTM_Global_API.TdtmToken>();
tokens.add(new TDTM_Global_API.TdtmToken('Filter_Test_TDTM', 'Contact', 
'BeforeInsert', 1, true, false, 'LastName', 'Testerson', false,  ''));
TDTM_Global_API.setTdtmConfig(tokens);
List<TDTM_Global_API.TdtmToken> config = TDTM_Global_API.getTDTMConfig();
System.debug('****config: ' + JSON.serializePretty(config));
```
Basically, the TDTM_Global_API setTdtmConfig is an utility method for a developer to implement their Apex unit test. The end user will not use it. We already included our unit test to test the logic inside the function. The developer can implement additional unit test to verify the fixes in this PR if needed. 

For QA, we need to **additionally** verify four things for this PR:

1. Verify the existing regression test cases still pass
2. Verify the existing default Trigger Handlers stay the same after HEDA upgrade
3. Verify that a New TDTM Trigger Handler (THAN_ClearCache_TDTM) is added after the HEDA upgrade
4. If user created their own Trigger Handlers, verify they will stay the same after HEDA upgrade


weizsforg:
Hi @lianeyee , Please have a look about this PR when you have time. Basically, this PR is about to update a class call TDTM_Global_API. We exposed this API to the external developer to help them to set the TDTM Trigger in their Apex unit test cases.

weizsforg:
Hi @ceiroa , can you please review this PR when you have time?

lianeyee:
@weizsforg I'd like to simplify the description of changes by combining the first two sentences. Also want to clarify the the implications of this change for developers. Can you check that this revised description is accurate?

Enhanced the setTdtmConfig method to clear cached TDTM tokens and to save values for Filter Field and Filter Value. Verify that your Apex unit tests and trigger handlers work as expected with these changes to setTdtmConfig.

weizsforg:
Hi @lianeyee , Thanks. Looks Great!

weizsforg:
Hi @ceiroa , thanks for the feedback! I will let you know after I address the issues. 

Based on our discussion with @kevinswiggum last Friday, I will also add a new TDTM class (THAN_ClearCache_TDTM) to clear the cache after insert the Trigger Handler record. 

Also, please note that this PR is about the caching of the TDTM token for Apex unit test.  Your previous comment about " Also, I think this PR still needs the message about a HEDA field shadowing a custom field on filtering." will be implemented in another WI (W-019578 TDTM Filtering not working correctly when using Filter Field of Boolean type). 

Thanks

src/classes/TDTM_Filter_TEST.cls
```diff
@@ -74,7 +74,7 @@ public with sharing class TDTM_Filter_TEST {
                 token.filterValue = null;
             }
 		}
-        TDTM_Global_API.setTdtmConfig(tokens);
+        TDTM_Global_API.setTdtmConfig(tokens, 'hed');
```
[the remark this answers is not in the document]
weizsforg: Actually, the namespace 'hed' here is just a way to differentiate HEDA default configs and other configs. It is not use to differentiate the managed and unmanaged environment. Also, the handlers in TDTM_DefaultConfig class all have the Owned_by_Namespace__c = 'hed'. So, the test cases will pass in the unmanaged environment. 

In the existing test cases, the  TDTM_Global_API.setTdtmConfig(tokens) also works since there is no trigger handler in the database yet. So, TDTM_Global_API.setTdtmConfig(tokens) and TDTM_Global_API.setTdtmConfig(tokens, 'hed') will return you the same result. 

Theoretically, user should use TDTM_Global_API.setTdtmConfig(tokens, 'hed') to manipulate HEDA tokens. And use TDTM_Global_API.setTdtmConfig(tokens) to manipulate tokens without namespace. However, I can reserve the change for existing test cases since the behavior is the same.

src/classes/TDTM_Global_API.cls
```diff
@@ -90,6 +90,9 @@ global class TDTM_Global_API {
         Set<TDTM_Global_API.TdtmToken> tokenSet =  new Set<TDTM_Global_API.TdtmToken>(tokens);
         manager.updateDefaultTdtmConfig( tokenSet );
 
+        //after save the tokens into database, clear the cached tokens
+        List<TDTM_Global_API.TdtmToken> cachedTokens = TDTM_Config.getTdtmConfig();
+        cachedTokens.clear();
```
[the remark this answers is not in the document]
weizsforg: I think it is better to clear the cache after manager.updateDefaultTdtmConfig( tokenSet ) call. The reason is that it is possible that the tokens pass in as a parameter is a reference of the cached tokens. If clear the cache first, then the "tokens" itself will be clear too. 

For example: 
List<TDTM_Global_API.TdtmToken> tokens = TDTM_Global_API.getTdtmConfig();
for(TDTM_Global_API.TdtmToken token : tokens) {
    if(token.className == 'REL_Relationships_TDTM') {
        token.active = false;
    }
}
TDTM_Global_API.setTdtmConfig(tokens);

If we clear the cache before manager.updateDefaultTdtmConfig( tokenSet ), the tokens will be clear too.

src/classes/TDTM_Global_API.cls
```diff
@@ -101,6 +104,10 @@ global class TDTM_Global_API {
         TDTM_Manager manager = new TDTM_Manager(namespace);
         Set<TDTM_Global_API.TdtmToken> tokenSet =  new Set<TDTM_Global_API.TdtmToken>(tokens);
         manager.updateDefaultTdtmConfig( tokenSet );
+
+        //after save the tokens into database, clear the cached tokens
+        List<TDTM_Global_API.TdtmToken> cachedTokens = TDTM_Config.getTdtmConfig();
+        cachedTokens.clear();
```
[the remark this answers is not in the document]
weizsforg: similar to the reason above. Any suggestions?

src/classes/TDTM_Manager.cls
```diff
@@ -168,6 +168,8 @@ public class TDTM_Manager {
         h.Active__c = t.active;
         h.Load_Order__c = t.loadOrderDec;
         h.Asynchronous__c = t.async;
+		h.Filter_Field__c = t.filterField;
```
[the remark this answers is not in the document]
weizsforg: haha, I can take a look about this indentation.

src/classes/TDTM_Manager_TEST.cls
```diff
@@ -275,6 +275,185 @@ private class TDTM_Manager_TEST {
...
+				Load_Order__c = 0.2,
+				Object__c = 'Contact',
+				Trigger_Action__c = 'BeforeInsert;BeforeUpdate'
+		);
+		insert newHandler;
+
+		//Build the token for the new handler
+		TDTM_Global_API.TdtmToken newHandlerToken = new TDTM_Global_API.TdtmToken('Contact_Update', 'Contact', 'BeforeInsert;BeforeUpdate', 0.2, true, false, null, null, false, '');
+
+		//get the cached token and verify the new token is not included
```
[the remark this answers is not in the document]
weizsforg: I think I will update this test case since added a new THAN_ClearCache_TDTM to clear the cache after insert the trigger handler record.

ceiroa:
@weizsforg Also removing myself as assignee in this one until it's ready, to avoid confusion.

weizsforg:
Hi @lianeyee , I updated the "Changes" part a little bit. Please have a look about it. Basically, I added a new TDTM class THAN_ClearCache_TDTM. What it does it to clear the TDTM tokens after insert a new Trigger Handler record.

weizsforg:
Hi @ceiroa , I updated the code based on your comments. I also updated the "Test Notes" part to add the steps to reproduce the issue and added the new THAN_ClearCache_TDTM class. Please have a look when you have time.

lianeyee:
@weizsforg Looks good. One question now: how is the cache-clearing by THAN_ClearCache_TDTM related to the cache-clearing by setTdtmConfig? Since the release note mentions both, it's not clear how they are related to/different from each other.

src/classes/TDTM_Filter_TEST.cls
```diff
@@ -331,6 +331,11 @@ public with sharing class TDTM_Filter_TEST {
 	public static testmethod void lookupField() {
 		List<TDTM_Global_API.TdtmToken> tokens = setup();
         TDTM_Global_API.setTdtmConfig(tokens);
+
+        system.assertEquals(new List<TDTM_Global_API.TdtmToken>(), tokens);
+
+        //since we clear the token in setTdtmConfig now, we need to get the new tokens here
+        tokens = TDTM_Global_API.getTdtmConfig();
```
ceiroa: Would it make sense that the `setup` method in this test class calls `TDTM_Global_API.getDefaultTdtmConfig()` instead of `TDTM_Global_API.getTdtmConfig()`? 

Also, do we really need this line? The tokens are explicitly set in line 353 below. Wouldn't that be enough?

src/classes/TDTM_Global_API.cls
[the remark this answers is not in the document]
ceiroa: Got it, thanks for the clarification. 

Another thought about all this; what about if instead of clearing the cache in both `setTdtmConfig` methods we leave it to the `TDTM_Manager.updateDefaultTdtmConfig` method to replace the cache values with the new ones?

src/classes/THAN_ClearCache_TDTM.cls
```diff
@@ -0,0 +1,60 @@
...
+    * @param listOld the list of Accounts from trigger old.
+    * @param triggerAction which trigger event (AfterInsert, AfterUpdate, etc.).
+    * @param objResult the describe for Trigger Handler.
+    * @return dmlWrapper.
+    ********************************************************************************************************/
+    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist,
+            TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+
+        if(newlist != null && newlist.size() > 0) {
+            if(triggerAction == TDTM_Runnable.Action.AfterInsert || triggerAction == TDTM_Runnable.Action.AfterUpdate) {
```
ceiroa: Shouldn't it run also on afterDelete? Otherwise you'll still have in the cache TDTM records that have been deleted.

src/classes/TDTM_Manager_TEST.cls
```diff
@@ -275,6 +275,178 @@ private class TDTM_Manager_TEST {
...
+		//verify that the Filter is saved in DB
+		handler = [select Filter_Field__c, Filter_Value__c from Trigger_Handler__c where Class__c = 'Contact_Update'];
+		system.assertEquals('Gender__c', handler.Filter_Field__c);
+		system.assertEquals('Male', handler.Filter_Value__c);
+
+	}
+
+	//Test to make sure we will not overwrite Filter if it is marked as usermanaged
+	@isTest
+	public static void filterForUserManagedHandler() {
```
ceiroa: Similar to the very first comment (the one on `STG_InstallScript_TEST`). Do we really need this test? It seems like the test `updatingDefaultHandlers` should already cover this scenario.  We could enhance that test instead of creating a new one (every new test adds to the time it takes to run our builds).

src/classes/TDTM_Manager_TEST.cls
```diff
@@ -275,6 +275,178 @@ private class TDTM_Manager_TEST {
...
+
+		Trigger_Handler__c individualAccountTdtm = [select Filter_Field__c, Filter_Value__c from Trigger_Handler__c where Class__c = 'ACCT_IndividualAccounts_TDTM'];
+		system.assertEquals(null, individualAccountTdtm.Filter_Field__c);
+		system.assertEquals(null, individualAccountTdtm.Filter_Value__c);
+	}
+
+	//Test to make sure we will not modify user's handler in HEDA upgrade
+	@isTest
+	public static void userHanlderAfterHEDAUpgrade() {
+		//Get default token and save it into DB
```
ceiroa: Similar to comment above. Is this test really necessary? Can we rely on a different test to cover this scenario?

src/classes/THAN_ClearCache_TEST.cls
```diff
@@ -0,0 +1,69 @@
...
+                Asynchronous__c = false,
+                Class__c = 'Contact_Update_User',
+                Load_Order__c = 0.20,
+                Object__c = 'Contact',
+                Trigger_Action__c = 'BeforeInsert;BeforeUpdate',
+                Filter_Field__c = 'Gender__c',
+                Filter_Value__c = 'Male'
+        );
+
+        insert contactUpdate;
```
ceiroa: Could we wrap this with `startTest` and `endTest`, just to make it explicit what the action that should trigger the code is?

src/classes/THAN_ClearCache_TEST.cls
```diff
@@ -0,0 +1,69 @@
...
+
+        //After insert Trigger Handler, system should clear the cache, so getTdtmConfig should include the new TriggerHandler
+        List<TDTM_Global_API.TdtmToken> tokens = TDTM_Global_API.getTdtmConfig();
```
ceiroa: Is `TDTM_Global_API.getTdtmConfig()` going to return only the TDTM record we manually inserted, or the default config plus the one we inserted? Can we assert the size of the returned list?

weizsforg:
Thanks @carlos! I will address the issues.

src/classes/THAN_ClearCache_TEST.cls
weizsforg: Here we should only get the TDTM record that we manually created. Since we didn't insert the default config using setTdtmConfig. I will put the assertion here to verify that the size is 1

src/classes/THAN_ClearCache_TEST.cls
weizsforg: right. I need to put the startTest/stopTest here.

src/classes/THAN_ClearCache_TDTM.cls
weizsforg: True. I think we need to add afterDelete here.

src/classes/THAN_ClearCache_TEST.cls
weizsforg: Here we should only get the TDTM record that we manually created. Since we didn't insert the default config using setTdtmConfig. I will put the assertion here to verify that the size is 1

src/classes/THAN_ClearCache_TEST.cls
weizsforg: right. I need to put the startTest/stopTest here.

src/classes/THAN_ClearCache_TDTM.cls
weizsforg: True. I think we need to add afterDelete here.

src/classes/STG_InstallScript_TEST.cls
```diff
@@ -302,4 +302,30 @@ public with sharing class STG_InstallScript_TEST {
         System.assertEquals(1, errorSchedule.size());
     }
 
+    //Test to make sure the new THAN_ClearCacheTDTM is added after the upgrade
+    @isTest
+    public static void thanClearCacheTDTM() {
```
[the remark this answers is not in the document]
weizsforg: I was thinking should we add this test cases or not. it is too specific to this version. I think I can remove it.

src/classes/TDTM_DefaultConfig.cls
```diff
@@ -130,6 +130,11 @@ public without sharing class TDTM_DefaultConfig {
         handlers.add(new Trigger_Handler__c(Active__c = true, Asynchronous__c = false, 
               Class__c = 'THAN_Filter_TDTM', Load_Order__c = 1, Object__c = 'Trigger_Handler__c', 
               Owned_by_Namespace__c = 'hed' ,Trigger_Action__c = 'BeforeInsert;BeforeUpdate'));
+
+        // Clear the TDTM cache on Trigger Handler record manipulation
+        handlers.add(new Trigger_Handler__c(Active__c = true, Asynchronous__c = false,
+                Class__c = 'THAN_ClearCache_TDTM', Load_Order__c = 2, Object__c = 'Trigger_Handler__c',
+                Owned_by_Namespace__c = 'hed' ,Trigger_Action__c = 'AfterInsert;AfterUpdate'));
```
weizsforg: I think I will add 'AfterDelete' here

src/classes/TDTM_Global_API.cls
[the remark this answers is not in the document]
weizsforg: hmm, I think clear cache is kind of safer. Because in some cases, the getTdtmConfig() can still return the wrong tokens. 

For example: 
Trigger_Handler__c contactUpdate = new Trigger_Handler__c(Active__c = true, Asynchronous__c = false, Class__c = 'Contact_Update_User', Load_Order__c = 0.20, Object__c = 'Contact', Trigger_Action__c = 'BeforeInsert;BeforeUpdate',);

insert contactUpdate;

defaultTokens = TDTM_Global_API.getDefaultTdtmConfig();
TDTM_Global_API.setTdtmConfig(defaultTokens, 'hed');

Even we replace the token with new one. Since here the new one just include the default one, it will still be wrong when user call "insert contact". 

Basically, I think the idea is that whenever user try to modify the TDTM trigger setup (both insert TriggerHandler or use setTdtmConfig) we need to clear the token to insure the user always get the correct tokens.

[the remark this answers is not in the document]
ceiroa: What's the issue in the sample code you wrote? You are calling `getDefaultTdtmConfig` that will always return the defaults...

src/classes/TDTM_Filter_TEST.cls
weizsforg: I think we can use TDTM_Global_API.getDefaultTdtmConfig() in the setup, should be ok. 

We need to have "tokens = TDTM_Global_API.getTdtmConfig()" here. The reason is that line 333 called setTdtmConfig(tokens). Since we cleared the token in setTdtmConfig, then the reference will be clear too. So, the tokens will be empty after line 333.  The line 353 will not work. 

I think we might need to mention in the documentation. The easiest way to update token is manipulate the tokens directly through getTdtmConfig. If user need to use setTdtmConfig, then they can't reuse the previous token reference anymore. They need to use getTdtmConfig to get a fresh set of token if they want to modify it.

src/classes/TDTM_Manager_TEST.cls
weizsforg: I think you are correct, I can enhance updatingDefaultHandlers to include Filter stuff

src/classes/TDTM_Global_API.cls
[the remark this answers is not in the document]
weizsforg: Oh, the issue is that after setTdtmConfig(defaultTokens, 'hed'), if we replace the token with defaultTokens. And then if user say "insert Contact". In this time, since the token is cached with defaultTokens, system will ignore the new Trigger Handler that we insert before.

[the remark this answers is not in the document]
weizsforg: That is if we replace the cache values with the new ones. In this case, the "new ones" is actually just the default token.

src/classes/TDTM_Manager_TEST.cls
weizsforg: Looks like we can remove this case.

weizsforg:
Hi @lianeyee , basically there is no direct connection between the cache-clearing by THAN_ClearCache_TDTM and the cache-clearing by setTdtmConfig. Since user can modify the TDTM Triggers by directly insert/update/delete the trigger handler record or use setTdtmConfig, we just clear the cache in both scenarios.

lianeyee:
OK, thanks for the clarification, @weizsforg!

src/classes/TDTM_Filter_TEST.cls
```diff
@@ -163,7 +163,8 @@ public with sharing class TDTM_Filter_TEST {
     }
     
     public static testmethod void InvalidFilterField() {
-        List<TDTM_Global_API.TdtmToken> tokens = setup();
+        //If we use getDefaultTDTMConfig from setup, then we can't save the token with invalid field. Because even we set  THAN_Filter_TDTM = inacive and save the token (insert TriggerHandler), system will call getTdtmConfig and get it from default config. So, in here, we need to modify the cache directly.
```
ceiroa: Sorry @weizsforg, this comment is throwing me off a bit. It says "system will call getTdtmConfig and get it from default config". Shouldn't getTdtmConfig get whatever is in the database? Maybe the comment is outdated?

src/classes/TDTM_Global_API.cls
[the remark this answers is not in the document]
ceiroa: Sorry, I don't get it. If the user calls `setTdtmConfig`,  only the configuration passed as parameter should persists. The Trigger Handler record previously inserted shouldn't exist any more. Maybe talking in person will be an easier way to clarify this. I'll send you a meeting for after Thanksgiving.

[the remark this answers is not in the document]
weizsforg: Yeah, it is tricky here. Based on current logic, there are two setTdtmConfig methods. The namespace version  setTdtmConfig(tokens, 'hed') will only to modify (insert/update/delete) tigger handlers with 'hed' namespace.  The regular version setTdtmConfig(tokens) will only modify (insert/update/delete) tigger handlers with empty namespace. That means, the tokens pass in to setTdtmConfig is not a complete set of all the Trigger Handlers in the system.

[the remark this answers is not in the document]
weizsforg: One special case that user can use either version of setTdtmConfig is when the trigger handler table is empty. Because system there is only insert but not update of the existing trigger handler.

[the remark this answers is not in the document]
weizsforg: Also, another example is that. Say, user created some tigger handlers in HEDA. When we do a upgrade of HEDA, we call TDTM_Global_API.getDefaultTdtmConfig() and then TDTM_Global_API.setTdtmConfig(defaultTokens, 'hed'). Here, the default tokens is HEDA default config. If the Trigger Handler record previously inserted shouldn't exist any more, then we should delete all existing Trigger Handler that is not in HEDA default config including customer one.  So, back to my very first example. After we call TDTM_Global_API.setTdtmConfig(defaultTokens, 'hed'), the trigger handler insert previously should still in the database and as a part of TDTM triggers.

src/classes/TDTM_Filter_TEST.cls
weizsforg: Sorry about the confusion here. yes, getTdtmConfig "should" get whatever is in the database. However, if the database is empty (like in this test case), it will read the default config and return it.

I think I will modify the comments to make it more clear.

src/classes/TDTM_Global_API.cls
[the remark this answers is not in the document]
weizsforg: Thanks @ceiroa. I will move the clear parts in the manager.updateDefaultTdtmConfig function.

src/classes/TDTM_Global_API.cls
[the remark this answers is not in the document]
weizsforg: similar to another function.

src/classes/TDTM_Global_API.cls
```diff
@@ -90,6 +90,9 @@ global class TDTM_Global_API {
         Set<TDTM_Global_API.TdtmToken> tokenSet =  new Set<TDTM_Global_API.TdtmToken>(tokens);
```
weizsforg: Update the description of the the method to indicate this will only update the none namespace part.

weizsforg:
Hi @ceiroa , I updated the description of the two setTDTMConfig method. Also moved the clear cached tokens into the TDTM Manager updateDefaultTdtmConfig method. Please have a look.

chrismcmahon:
I exercised as much of the 'verify four things' list as possible. 

For a final check, I added a note to the 'HEDA Regression Checklist' to remember to check the final result of this change in the QA Sandbox, which is subject to a normal push upgrade process.

PR #519, opened 2017-12-18
fix the duplicate field issue

### LINKED ISSUE #518: TDTM duplicate fields selected error
There is a "duplicate field selected error" that has been reported by 2 community members (see Reference below) that we know of. 

Reference:
Error reported by Kalman: https://powerofus.force.com/_ui/core/chatter/groups/GroupProfilePage?g=0F980000000D6uj&fId=0D58000003dzCi3
Error reported by Richard: https://powerofus.force.com/_ui/core/userprofile/UserProfilePage?u=0058000000DCwx7&tab=sfdc.ProfilePlatformFeed&fId=0D58000003hFzwD

Jace is working to confirm if Richard's scenario is the same as Kalman's (i.e. Changed the Record Type Label of an Account & including modifications to Affiliation Mappings).

Although there is a workaround for Kalman's issue (see workaround steps below), we should do more investigations to confirm the cause. 

Workaround:
You can get around the duplicate field error by going to Setup > Installed Packages > HEDA > View Components > Affl Mappings > Manage > Edit Program Plan and remove the value for Primary Affl Field. Do your bulk upload and then put the value back in the Primary Affl Field.

### LINKED ISSUE #524: System populates the wrong Primary field when change Account record in batch mode
When we change the Account Record type from a record type didn't have a primary affiliation mapping (e.g. Admin) to a record type has a primary affiliation mapping (e.g. Academic Program), system will update the corresponding Primary Field on a Contact. It works when user try to update Account record type one by one. However, system will populate extra Primary Field on Contact if the user update multiple Account record type in batch mode. 

For example, 
1. Create two Contacts: A, B
2. Create two Accounts: Math, Football. Set both record type to Admin first
3. Create two Affiliations: A - Math, B-Football. Currently, the primary fields on A and B should be empty 
4. Update Accounts in one operation: Math - Academic Program, Football - sports organization
5. System should update Primary Academic Program of A and the Primary sports org of B. However, you will see both Primary Academic Program, Primary sports org of A becomes Math (similar thing happen to B). Which is not correct.

weizsforg:
# Critical Changes

# Changes
- Fixed an error that prevented you from mass updating the record type on Accounts. This duplicate field error occurred whenever the original record type value didn't have a primary affiliation mapping and the new record type value you were updating to did.

- Fixed an issue that incorrectly populated the primary affiliation field on Contact when mass updating the record type on Accounts. The issue occurred whenever the updates involved multiple record types and the original record type value didn't have a primary affiliation mapping and the new record type value you were updating to did.

# Issues Closed
Fixes #518 
Fixes #524 
# New Metadata

# Deleted Metadata

# Testing Notes
Steps to reproduce the issue #518  :
1. Create multiple Accounts with Administrative record type.
2. Update those Accounts record type from Administrative to other record type, such as  "Academic Program".  Please note that you need to do the update in bulk instead of one by one. Also, the record type that you change to must have a mapping in "Affl Settings". 
3. System will display the error "hed.TDTM_Account: execution of AfterUpdate
caused by: System.QueryException: duplicate field selected: Primary_Academic_Program__c" 

Verify the fix:
You should not get any error when you do the steps above. 

Steps to reproduce the issue #524 :
1. Create two Contacts: A, B
2. Create two Accounts: Math, Football. Set both record type to Admin first
3. Create two Affiliations: A - Math, B-Football. Currently, the primary fields on A and B should be empty
4. Update Accounts in one operation: Math - Academic Program, Football - sports organization
5. System should update Primary Academic Program of A and the Primary sports org of B. However, you will see both Primary Academic Program, Primary sports org of A becomes Math (similar thing happen to B). Which is not correct.

### DIFF
--- a/src/classes/AFFL_AccRecordType_TDTM.cls
+++ b/src/classes/AFFL_AccRecordType_TDTM.cls
@@ -154,7 +154,9 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
         if(accsTypeAdded.size() > 0) {
             String childrenAccsTypeAddedQuery = 'select Account__c, Contact__c';
             for(Integer i = 0; i < accsTypeAdded.values().size(); i++) {
-                childrenAccsTypeAddedQuery += ', Contact__r.' + accsTypeAdded.values()[i];
+                if(!childrenAccsTypeAddedQuery.contains(accsTypeAdded.values()[i])) {
+                    childrenAccsTypeAddedQuery += ', Contact__r.' + accsTypeAdded.values()[i];
+                }
             }
             Set<ID> accsTypeAddedIDs = accsTypeAdded.keySet();
             childrenAccsTypeAddedQuery += ' from Affiliation__c where Account__c IN :accsTypeAddedIDs';
--- a/src/classes/AFFL_AccRecordType_TDTM.cls
+++ b/src/classes/AFFL_AccRecordType_TDTM.cls
@@ -152,25 +152,15 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
         //B - Find all Contacts with an Affiliation to an Account that didn't have a record type before,
         //but that now has it. Update the matching primary affiliation field on Contact.
         if(accsTypeAdded.size() > 0) {
-            String childrenAccsTypeAddedQuery = 'select Account__c, Contact__c';
-            for(Integer i = 0; i < accsTypeAdded.values().size(); i++) {
-                if(!childrenAccsTypeAddedQuery.contains(accsTypeAdded.values()[i])) {
-                    childrenAccsTypeAddedQuery += ', Contact__r.' + accsTypeAdded.values()[i];
-                }
-            }
-            Set<ID> accsTypeAddedIDs = accsTypeAdded.keySet();
-            childrenAccsTypeAddedQuery += ' from Affiliation__c where Account__c IN :accsTypeAddedIDs';
-            List<Affiliation__c> afflsAccsTypeAdded = Database.query(childrenAccsTypeAddedQuery);
+            List<Affiliation__c> afflsAccsTypeAdded = [select Account__c, Contact__c from Affiliation__c where Account__c IN :accsTypeAdded.keySet()];
 
             //Populate the necessary primary field for each Contact
             List<Contact> contactsToUpdate = new List<Contact>();
             for(Affiliation__c affl : afflsAccsTypeAdded) {
                 Contact contact = new Contact(ID = affl.Contact__c);
-                //Iterate over the Contact primary fields
-                for(String primaryField : accsTypeAdded.values()) {
-                    contact.put(primaryField, affl.Account__c);
-                    contactsToUpdate.add(contact);
-                }
+                //We need to populate the Primary Field on Contact based on Account's Type
+                contact.put(accsTypeAdded.get(affl.Account__c), affl.Account__c);
+                contactsToUpdate.add(contact);
             }
             if(contactsToUpdate.size() > 0) {
                 dmlWrapper.objectsToUpdate.addAll((List<SObject>)contactsToUpdate);
--- a/src/classes/AFFL_AccRecordType_TEST.cls
+++ b/src/classes/AFFL_AccRecordType_TEST.cls
@@ -248,4 +248,46 @@ public with sharing class AFFL_AccRecordType_TEST {
         List<Account> hhAccounts = [Select Id from Account where RecordTypeId =: householdRecTypeID];
         system.assertEquals(5, hhAccounts.size());
     }
+
+    @isTest
+    public static void fillInPrimaryFieldForContacts() {
+        setup();
+
+        //Create Contacts
+        Contact js = new Contact(FirstName = 'John', LastName = 'Smith');
+        Contact mb = new Contact(FirstName = 'Mark', LastName = 'Bruce');
+        List<Contact> contacts = new List<Contact>{js, mb};
+        insert contacts;
+
+        //Create Accounts with Admin Record Type
+        Account company = new Account(Name = 'ABC Company', RecordTypeId = UTIL_Describe.getAdminAccRecTypeID());
+        Account hh = new Account(Name = 'Mark Household', RecordTypeId = UTIL_Describe.getAdminAccRecTypeID());
+        List<Account> accs = new List<Account>{company, hh};
+        insert accs;
+
+        //Create Affiliations
+        Affiliation__c companyAffl = new Affiliation__c(Contact__c = js.ID, Account__c = company.ID, Primary__c = true);
+        Affiliation__c hhAffl = new Affiliation__c(Contact__c = mb.ID, Account__c = hh.ID, Primary__c = true);
+        List<Affiliation__c> affls = new List<Affiliation__c>{companyAffl, hhAffl};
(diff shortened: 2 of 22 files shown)


weizsforg:
Hi @lianeyee , can you please have a look of the 'Changes' section in this PR? Basically I fixed an issue when user try to update the Account record type in bulk.

xinnanlisf:
Just a side note while reviewing the pull request, we should let the users know that this error only occurs when the new record type has primary affl mapping field. Maybe we can put this info into the changes section.

src/classes/AFFL_AccRecordType_TDTM.cls
```diff
@@ -154,7 +154,9 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
         if(accsTypeAdded.size() > 0) {
             String childrenAccsTypeAddedQuery = 'select Account__c, Contact__c';
             for(Integer i = 0; i < accsTypeAdded.values().size(); i++) {
-                childrenAccsTypeAddedQuery += ', Contact__r.' + accsTypeAdded.values()[i];
+                if(!childrenAccsTypeAddedQuery.contains(accsTypeAdded.values()[i])) {
```
xinnanlisf: This should work but I am not very comfortable with using string.contains(). Strings are likely unreliable and might cause issues.

weizsforg: hmm, any suggestion here? Any specific concern?

xinnanlisf: I don't know maybe a set of strings should work.
If you look at that method, we don't really use that map much. I am not sure if this is the right time to refactor this method but honestly I am not a big fan of it.

lianeyee:
@weizsforg I just updated the Changes text. Please check for accuracy/clarity. Thanks!

weizsforg:
@lianeyee Looks great.

src/classes/AFFL_AccRecordType_TDTM.cls
weizsforg: Agree that we need to change the logic a little bit since I think I might found a logic error in this function. But I do think we need the map in this function since we want the Account Ids and the corresponding primary field. So, we need the map to connect Account Id and its primary field.  Let's schedule sometime to discuss more. 

1. We might not need to query the Contact fields since we do not use it in the following part.
2. We might not want to loop through every primary field in the values() for each contact. We need to use the map to find the corresponding primary field for the contact.

weizsforg:
Hi @lianeyee , I updated the "Changes" part of the PR since it includes the fix of 2 issues now.

lianeyee:
Thanks, @weizsforg. I edited the second paragraph to clarify the fix (please verify) and added bullets.

weizsforg:
Thanks @lianeyee. I think I missed one condition for the second issue. 
Basically, the second issue has the same condition as the first one. But it also has an extra condition "updates involved multiple record types".

So, can we say something like this:

- Fixed an issue that incorrectly populated the primary affiliation field on Contact when mass updating the record type on Accounts. The issue occurred whenever the updates involved multiple record types and the original record type value didn't have a primary affiliation mapping and the new record type value you were updating to did.

lianeyee:
@weizsforg I just want to clarify one thing, because there are multiple record types involved. Is the condition all or any? That is, if all of the original record type values didn't have primary affiliation mappings, and all of the new record type values did, then the error occurred. What if there was a mix, i.e., some original values did have mappings?

weizsforg:
@lianeyee Sorry for the confusion. Basically, this PR fixed two issues. 

Both issue happen under the condition when user try to do mass update Account record types and the original record type value didn't have a primary affiliation mapping and the new record type value you were updating to did. 

However, the issue 2 only occurs when updates involved multiple record types. 

To answer your question, if some original values did have mappings and updates involved multiple record types. Issue 2 will not occur. 

Does that make sense?

lianeyee:
@weizsforg I see. Thanks for explaining in detail. Your proposed change looks good!

src/classes/AFFL_AccRecordType_TEST.cls
```diff
@@ -57,7 +57,7 @@ public with sharing class AFFL_AccRecordType_TEST {
     public static void changePrimaryAffiliationRecordType() {        
         setup();
 
-        Contact contact = new Contact(FirstName = 'Test', LastName = 'Testerson');
+        Contact contact = UTIL_UnitTestData_TEST.getContact();
```
xinnanlisf: Just a quick suggestion maybe, since many test methods is creating contact, would it be nice just make a static variable on top of the test class?

weizsforg: That is a good suggestion. 
I am thinking that:  Does it make sense to not using a class level static variable since the purpose of using the UTIL class is to make sure every test method has unique Account/Contact?

src/classes/AFFL_AccRecordType_TEST.cls
```diff
@@ -212,4 +212,78 @@ public with sharing class AFFL_AccRecordType_TEST {
         contact = [select Primary_Organization__c from Contact where ID=:contact.Id];
         System.assertEquals(acc.ID, contact.Primary_Organization__c);
     }
+
+    @isTest
+    public static void updateAdminAccRecordTypeBulk() {
```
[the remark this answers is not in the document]
weizsforg: Yes, it is for #518. I don't think we need to create a test method for each bug if the bugs are related issue. In this case, the issue #518 and #524 are not related issues. We put it into the same PR is because they touch the same function.

The #518 occurs when we update more than 1 account's record type to the same record type (no contact and affiliation records need to be involved). So, I use a small batch (10 records) to verify it.

Maybe I should add comments above the method?

src/classes/AFFL_AccRecordType_TEST.cls
```diff
@@ -212,4 +212,78 @@ public with sharing class AFFL_AccRecordType_TEST {
...
+        Account company = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe.getAdminAccRecTypeID())[0];
+        Account hh = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe.getAdminAccRecTypeID())[0];
+        List<Account> accs = new List<Account>{company, hh};
+        insert accs;
+
+        //Create Affiliations
+        Affiliation__c companyAffl = new Affiliation__c(Contact__c = john.ID, Account__c = company.ID, Primary__c = true);
+        Affiliation__c hhAffl = new Affiliation__c(Contact__c = mark.ID, Account__c = hh.ID, Primary__c = true);
+        List<Affiliation__c> affls = new List<Affiliation__c>{companyAffl, hhAffl};
+        TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.AFFL_MultiRecordType_TDTM_Before_Insert, false);
```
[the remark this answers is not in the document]
weizsforg: good catch.

src/classes/AFFL_AccRecordType_TEST.cls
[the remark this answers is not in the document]
xinnanlisf: Well I like the idea of more unit test methods. I am only a little concerned that test will run too long. Not a big question for this pull request but maybe we can figure out what worth creating a unit test.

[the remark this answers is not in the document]
weizsforg: That is a very good point! 
Sometimes I feel like the test case can help me to make sure this scenario works , but it does increase the total execution time.

PR #520, opened 2017-12-22
Feature/ceiroa/tdtm advancement

ceiroa:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


ceiroa:
NOT ready for review and merge yet.

This is complementary to https://github.com/SalesforceFoundation/adv/pull/2.

src/classes/TDTM_Config.cls
```diff
@@ -50,33 +50,65 @@ public class TDTM_Config {
...
         if(tdtmTokensCache != null && tdtmTokensCache.size() > 0) {
             return tdtmTokensCache;
         }
-        List<TDTM_Global_API.TdtmToken> tokens = new List<TDTM_Global_API.TdtmToken>();
+
+        Boolean useAdv = false;
+        String apiHedaClass = '';
+        //Check if Advancement is installed and enabled.
+        try {
+            List<SObject> advSettings = Database.query('select Use_Advancement__c, API_HEDA_Class__c from Advancement_Setting__mdt');
```
davidhabib: won't these fields and metadata object from the advancement package need to be namespaced?

src/classes/TDTM_Config.cls
```diff
@@ -50,33 +50,65 @@ public class TDTM_Config {
...
+            useAdv = (Boolean)advSettings[0].get('Use_Advancement__c');
+            apiHedaClass = (String)advSettings[0].get('API_HEDA_Class__c');
+        } catch(Exception e) {
```
davidhabib: rather than try / catch, why not just base it off advSettings.size() == 0?

src/classes/TDTM_Config.cls
```diff
@@ -50,33 +50,65 @@ public class TDTM_Config {
...
+                        tokens = API_HEDA.getDefaultTdtmConfigTokens();
+                    }
+                }
+            } else {
+                //Not Using Adv config.
+                tdtmConfig = TDTM_DefaultConfig.getDefaultRecords();
+            }
         }
 
+        tokens = new List<TDTM_Global_API.TdtmToken>();
```
davidhabib: won't this overwrite the tokens you loaded from API_HEDA in line 94?

src/package.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
davidhabib: i've noticed this from IntelliJ (I think), and I'm not sure we want it to go in or not.  @jlantz do you know?

src/classes/TDTM_Adapter.cls
```diff
@@ -0,0 +1,61 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2017
+* @group TDTM
+* @group-content ../../ApexDocContent/TDTM.htm
+* @description Provides an adapter for other packages to be able to inject customizations into HEDA.
+**/
+
+//@TODO: is this a good name for this class? Is the TDTM prefix correct? Later on we may want it to be an adaptor for additional functionality.
```
davidhabib: I agree that you should come up with a better name, since this won't be tied to just TDTM.  basically, you are trying to define a class that represents the contract/interaction between HEDA and Advancement.  Not sure if it would ever be used by other products as well, but maybe?  So maybe something like 'IntegrationService'?

src/classes/TDTM_Config.cls
weizsforg: True, need to fix this.

sam-knox:
@ceiroa - on the changes in this PR I wanted to make sure I understand the changes:

For the preferred email issue, is it just a match on Label or Name as well? In other words does the system look at "Work Email" or "Work_Email__c" or both?

And with Primary Contact had the previous behavior been that every time a new Contact was added to an Account, the Primary Contact would change to the newly added one?

sam-knox:
@lianeyee - Made some updates to the two Changes items in this PR.

lianeyee:
@sam-knox I'm not sure the "Fixed" format works best for these changes. They do work well for relatively simple, straightforward changes that follow the "Fixed an issue/error that prevented blah blah blah." These changes are better served by an "If" format, in my opinion. i.e.,

If you have email fields that use the same label (or name?) as a HEDA email field, we now choose the HEDA version of the field for a Contact's Preferred Email address. For example, if a user chooses Work Email as the preferred email address and your org has multiple fields on the Contact object labeled Work Email, we use the value from the HEDA Work Email field.

If a new Contact is added to an Account that already has a Primary Contact, we no longer replace the Primary Contact with the newly added Contact. The first Contact added to the Account is the Primary Contact unless you explicitly change it to another Contact.

sam-knox:
@lianeyee - Thanks Liane, much improved. Still learning the art here!

ceiroa:
Good question, @sam-knox! Even though I originally opened this PR, this fix (and most of the other code) was actually implemented by Wei. I'm still reviewing it, so don't have all the details yet. It looks like it's comparing labels. It also looks like it's not only for the Work Email field, but also for any other email field in Contact. I'll confirm with you soon.

ceiroa:
@weizsforg I have removed from the description text that now belongs in separate pull requests. May want to take another look to see if anything else needs to be changed/added. Also, the branch hit a conflict.

weizsforg:
Thanks! @ceiroa I think the conflict comes from the ACCT_IndividualAccounts_TDTM class.

I am thinking should we review/QA/merge other two PRs into master first. And then I can resolve the conflict after merge those two PRs into this branch since other two PRs also modified the ACCT_IndividualAccounts_TDTM class.

weizsforg:
@ceiroa Do you think I should pull the two fixes of HEDA out of this PR and merge it back after they get merge into master?

ceiroa:
Hey, @weizsforg ~ Could you please populate the PR notes? Specially the testing notes section. Thanks!

weizsforg:
Hi @ceiroa , Yes, I will double check and update the PR Notes of three branches. Thanks!

weizsforg:
Hi @lianeyee I just updated the notes in this PR. Can you please review it? Basically, this PR is about providing support to Adv. How detail we need to go about the implementation?

src/classes/Advancement_Adapter.cls
```diff
@@ -0,0 +1,101 @@
...
+global virtual with sharing class Advancement_Adapter {
+
+    /**
+     * @description To be called by the HEDA TDTM trigger dispatcher. It returns the HEDA TDTM tokens with the new one from Adv.
+     *
+     * @return A list of HEDA tokens.
+     */
+    global virtual List<TDTM_Global_API.TdtmToken> getDefaultTdtmConfigTokens() {
+        if(!Advancement_Info.useAdv()) {
+            throw new AdapterException('Invalid call. This function is reserved for Salesforce use.');
```
ceiroa: This string should be a label.

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -428,6 +460,10 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
                 if(defaultRecTypeID != null)
                     a.RecordTypeID = defaultRecTypeID;
 
+                if(Advancement_Info.useAdv() && Advancement_Info.getApiHEDA() != null) {
+                    Advancement_Info.getApiHEDA().configAccountForContact(a);
+                }
+
```
ceiroa: Are there unit tests for the changes made in this class in The Advancement repo? Otherwise, is there some other way to verify this code is correct and be alerted if it breaks?

src/classes/Advancement_Adapter.cls
```diff
@@ -0,0 +1,101 @@
...
+
+        return TDTM_Global_API.getDefaultTdtmConfig();
+    }
+
+    /**
+    *  @description This method is used to inject the NPSP rollup logic into HEDA Trigger Handler.
+    */
+    global virtual void rollupAccount(List<Id> listAccountId) {
+        if(!Advancement_Info.useAdv()) {
+            throw new AdapterException('Invalid call. This function is reserved for Salesforce use.');
```
ceiroa: String should also be label.

src/classes/Advancement_Adapter.cls
```diff
@@ -0,0 +1,101 @@
...
+
+        //No logic for HEDA
+    }
+
+    /**
+     * @description This method is used to inject the NPSP opportunity logic info HEDA Trigger Handler.
+     */
+    global virtual void moveOppsToContactAccount(Map<Id,Id> mapContactIdAccountIdNew, Map<Id,Id> mapContactIdAccountIdOld) {
+        if(!Advancement_Info.useAdv()) {
+            throw new AdapterException('Invalid call. This function is reserved for Salesforce use.');
```
ceiroa: Label.

src/classes/Advancement_Adapter.cls
```diff
@@ -0,0 +1,101 @@
...
+
+        //No logic for HEDA
+    }
+
+    /**
+     * @description This method is used to populate NPSP's primary contact field in HEDA Trigger Handler
+     */
+    global virtual void primaryContact(Account a, Id contactId) {
+        if(!Advancement_Info.useAdv()) {
+            throw new AdapterException('Invalid call. This function is reserved for Salesforce use.');
```
ceiroa: Label.

src/classes/Advancement_Adapter.cls
```diff
@@ -0,0 +1,101 @@
...
+
+        //No Logic for HEDA
+    }
+
+    /**
+     * @description This method is used to populate NPSP's system account field in HEDA Trigger Handler
+     */
+    global virtual void configAccountForContact(Account a) {
+        if(!Advancement_Info.useAdv()) {
+            throw new AdapterException('Invalid call. This function is reserved for Salesforce use.');
```
ceiroa: Label.

src/classes/Advancement_Adapter.cls
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        //No Logic for HEDA
+    }
+
+    /**
+     * @description This is the expection class for Advancement Adapter class
+     */
+    global class AdapterException extends Exception{}
```
ceiroa: Do we need an Advancement_Adapter_TEST class to exercise these methods that throw exceptions?

src/classes/ACCT_IndividualAccounts_TDTM.cls
weizsforg: That is a good point. I think the logic in HEDA is empty. So, if HEDA unit test should cover it. If the logic changes, HEDA unit test will break. 
But we do need to add unit test in Adv repo to see if the injected logic works. 
What do you think?

src/classes/Advancement_Adapter.cls
weizsforg: right. I will add a label for it.

src/classes/Advancement_Adapter.cls
weizsforg: Yes, I will add the unit test to verify it. Also, I think we need to skip the test if Adv is installed.

src/classes/ACCT_IndividualAccounts_TDTM.cls
ceiroa: Yes, I think we should have unit tests in the ADV repo for it. Maybe there are tests in NPSP that we can reuse for this with minor changes?

src/classes/Advancement_Adapter.cls
```diff
@@ -0,0 +1,101 @@
...
+        if(!Advancement_Info.useAdv()) {
+            throw new AdapterException('Invalid call. This function is reserved for Salesforce use.');
+        }
+
+        return TDTM_Global_API.getDefaultTdtmConfig();
+    }
+
+    /**
+    *  @description This method is used to inject the NPSP rollup logic into HEDA Trigger Handler.
+    */
```
ceiroa: Just realized some of these methods are also missing @param annotations.

weizsforg: Thanks, I will add @param for all methods.

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -243,6 +254,11 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
             updateAdminAccountsName(listContactsChangedLastName, dmlWrapper);
         }
 
+        //Inject Adv logic to handle opportunity
+        if (mapContactIdAccountIdNewMoveOpps.size() > 0 && Advancement_Info.useAdv() && Advancement_Info.getApiHEDA() != null) {
+            Advancement_Info.getApiHEDA().moveOppsToContactAccount(mapContactIdAccountIdNewMoveOpps,mapContactIdAccountIdOldMoveOpps);
```
pedrodalcol: Let's add spaces after commas for consistency.

weizsforg:
Hi @ceiroa  , this PR is ready for review. 

I added the @params in all the methods. Switched to use Label for the exception. Added the test class for Advancement_Adapter. 

Thanks

ceiroa:
Looks great, @weizsforg! Please approve (as I cannot do that, since I was the originator of the PR). Thanks!

annejackson:
@ceiroa  should we make sure all checks are successful before I continue testing this? I'm just thinking to reduce possible "duplication/rework" time. fyi @weizsforg

ceiroa:
Not sure I follow, @annejackson. All checks look green for me...

annejackson:
I see them all green for me too now @ceiroa - maybe because it had to finish merging catalan or something else?

ceiroa:
@annejackson If something else was merged into master it will also get automatically merged into any open branches, which will cause these to run too. If you see one of them in yellow it means the build is running. You did the right thing by waiting if that was the case. It usually only takes a few (10 to 30) minutes.

weizsforg:
Hi, @annejackson @ceiroa , I saw the test failed due to the build issue. So, I just rebuilt it previously.

annejackson:
Thanks @weizsforg  and @ceiroa

annejackson:
This is looking good. Merging this pull request. @ceiroa and @weizsforg

PR #526, opened 2018-01-12
Feature/xinnanli/tdtm excluded usernames

xinnanlisf:
# Critical Changes

# Changes
- Added a new field, Usernames_to_Exclude__c, to the Trigger Handler object. Use it to specify a semicolon-separated list of usernames for which a Trigger Handler should not run. Remember to add this field to Trigger Handler list views, page layouts, and user permissions.

# Issues Closed

# New Metadata
Custom Field:
Usernames_to_Exclude__c
# Deleted Metadata

# Testing Notes
Add current usernames to the Usernames_to_Exclude__c of trigger handler, for example individualAccount class.
Set default account model as admin and create a contact
Make sure admin account is not created, which means trigger handler is disabled for current user


lianeyee:
@xinnanlisf Are you ready to flesh out the release notes a bit? We should be specific about the new field name (not count on users reading the New Metadata section) and where it's been added, along with describing the problem it solves.

xinnanlisf:
@lianeyee 

Yes thank you for reviewing it. We are adding a new field, Usernames to Exclude, to trigger handler page layout. It allows users to prevent triggers from running for certain users.

lianeyee:
Yep, I actually meant that the release note was looking skimpy with just "Allow trigger handlers to filter usernames." I went ahead and wrote out more detail. Please tweak further if needed @xinnanlisf. Thanks!

weizsforg:
@xinnanlisf Great Work!

One question: Should we add Usernames_to_Exclude__c field also in "equals" function in TDTM_Global_API and "mergeToken" function in TDTM_Manager? I think every other fields on Trigger_Handler__c are in those functions.

xinnanlisf:
@weizsforg 

Great question. mergeToken() would make sense. I don't see anywhere using equals() thou.

weizsforg:
@xinnanlisf The TDTM_Manager updateDefaultTdtmConfig() function uses "equals" function.

xinnanlisf:
Hi @lianeyee,

In the release note, we would also like to remind users to configure listview/layout/permissions to use this new field. I have updated the release note. Could you take a review please? Thank you!

lianeyee:
Done. Thanks @xinnanlisf!

annejackson:
Noticed that if a user entered multiple usernames to exclude and separated them by a semicolon + a space, the excluded username function did not work. Chatted with @xinnanlisf .

annejackson:
Retested with the space fix and it all looks great. Now if a user enters multiple usernames to exclude and separates them by a semicolon + a space, the excluded username function does work.  fyi @xinnanlisf  and @sreesureka. I will wait to merge the pull request until you can join Doug and I.

PR #540, opened 2018-02-01
Allow backward compatibility for no excluded usernames

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/TDTM_Global_API.cls
+++ b/src/classes/TDTM_Global_API.cls
@@ -228,7 +228,7 @@ global class TDTM_Global_API {
         //cumulusci-managed@deprecated
         global TdtmToken(String className, String targetObject, String actions, Integer loadOrder, Boolean active, Boolean async, String filterField, String filterValue, Boolean userManaged) {
             
-            this(className, targetObject, actions, Decimal.valueOf(loadOrder), active, async, filterField, filterValue, userManaged, null, null);
+            this(className, targetObject, actions, Decimal.valueOf(loadOrder), active, async, filterField, filterValue, userManaged, null);
         }
 
         /********************************************************************************************************
--- a/src/classes/TDTM_Global_API.cls
+++ b/src/classes/TDTM_Global_API.cls
@@ -245,8 +245,8 @@ global class TDTM_Global_API {
         * @param ownedByNamespace The token's owned-by-namespace flag.
         *********************************************************************************************************/
         //cumulusci-managed@deprecated
-        global TdtmToken(String className, String targetObject, String actions, Integer loadOrder, Boolean active, Boolean async, String filterField, String filterValue, Boolean userManaged, String ownedByNamespace) {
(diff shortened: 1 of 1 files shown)


weizsforg:
@xinnanlisf For the 'old' method that you put it back, should we use the same pattern that you use for line 231. Just call the new method and set the usernamesToExclude to null.

weizsforg:
@xinnanlisf Also, do we need to put //cumulusci-managed@deprecated for the old method just like the other TdtmToken method?

xinnanlisf:
@weizsforg 

Yeah I just put it back. Although I think it is deprecated so it should not matter much.

Re: Whether we should add //cumulusci-managed@deprecated to the old constructor. I thought we want to provide backward compatibility for those users who use this api methods.

ceiroa:
@xinnanlisf I believe the @deprecated annotation would still allow existing code to continue referring to it. It would just ban any code from starting to do so. Please take a look at https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_classes_annotation_deprecated.htm to make sure I'm not misunderstanding this.

We should check, maybe with @davidjray, what the 'cumulusci-managed' comment is for.

weizsforg:
@xinnanlisf I just checked the old TDTMToken  method Signature. I think it should be 

global TdtmToken(String className, String targetObject, String actions, Decimal loadOrder, Boolean active, Boolean async, String filterField, String filterValue, Boolean userManaged, String ownedByNamespace)

instead of 

 global TdtmToken(String className, String targetObject, String actions, Integer loadOrder, Boolean active, Boolean async, String filterField, String filterValue, Boolean userManaged, String ownedByNamespace)

The type of loadOrder is wrong. Can you double check about it?

xinnanlisf:
@weizsforg 

Should be good now!

ceiroa:
@xinnanlisf Could you please include the explanation of what the 'cumulusci-managed' comment is for here? Adding the screenshot should be fine too. That way it's searchable in the Github history next time we ask ourselves why that's there. Thanks!

xinnanlisf:
The reason why we have //cumulusci-managed:
screen shot 2018-02-01 at 12 11 04 pm

weizsforg:
Hi @annejackson , this PR is a fix for the failed BETA build related to [W-021068](https://foundation--agf.na6.visual.force.com/a2x80000008j322AAA).  Can you please do another round of test to see if everything is ok?

annejackson:
Is there a way that I can see that the BETA build passed for this now @weizsforg and @xinnanlisf? 
Testing looks good on the scratch org for this pull request.

xinnanlisf:
@annejackson 

Per discussion with Chris, there is no way for us to make sure the Beta build succeed before merging.

annejackson:
Testing has passed. Merging this Pull Request.

PR #574, opened 2018-03-19
Use managed option in update_admin_profile for ci_release

cpolcyn:
Should fix the "Release Test" build failures.


cpolcyn:
Verified this fixes the ci_release flow for HEDA by creating a scratch org with release config and running the flow locally.

PR #599, opened 2018-05-04
Feature/xinnanli/gdpr

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


annejackson:
Please include testing notes @xinnanlisf and @lparrott as you review this item. For example, since a scratch org is new I won't have 90 days of data (but possibly would in a sandbox depending on what data it is collecting), so if there are developer tools and queries I can run to test this in a scratch org please let me know. Also, if there are some items I'll be able to test in a sandbox, since it's existed for more than 90 days, let me know that too. Thanks!

src/classes/ERR_DeleteOutdated.cls
```diff
@@ -0,0 +1,42 @@
...
+* @description A queueable job that detects errors which are more than 90 days old and delete them
+*/
+public virtual class ERR_DeleteOutdated implements Queueable {
+    /**
+    * @description Execute method to detect errors more than 90 days old and delete
+    */
+    public void execute(QueueableContext qc) {
+        DateTime now = DateTime.now();
+        DateTime ninetyDaysAgo = now.addDays(-90);
+        List<Error__c> errors = [SELECT Id FROM Error__c WHERE Datetime__c < :ninetyDaysAgo];
```
lparrott: Would this fail if it returned more than 50000 records? In my head we would make this process Batchable instead of Queueable, though I'm not sure about the risks. We could also just limit the query, and eventually it would catch up. @ceiroa Your thoughts?

src/classes/ERR_DeleteOutdated_TEST.cls
```diff
@@ -0,0 +1,54 @@
...
+*/
+@isTest
+public class ERR_DeleteOutdated_TEST {
+    /**
+     * @description Errors outdated should be deleted by a queueable job.
+     */
+    @isTest
+    private static void testOutdatedErrorDeleted() {
+        List<Error__c> errors = UTIL_UnitTestData_TEST.getListOfErrors(2);
+        errors[0].Datetime__c = errors[0].Datetime__c.addDays(-100);
```
lparrott: Might be good to run this on a larger data volume, especially for orgs that have been using HEDA for years.

src/classes/ERR_DeleteOutdated.cls
ceiroa: Good point, @lparrott. I've always used a Schedulable that calls a batch instead Queueable. What are the advantages of using Queueable, @xinnanlisf? How do the query limit applies in this case?

src/classes/ERR_DeleteOutdated_TEST.cls
ceiroa: Well, but for the test to be meaningful it'd have to run on a set of over 200. But that may cause the test suite to run longer. I'd like to avoid that, unless we have a good reason. (There is in fact one test in HEDA that runs on a set of over 200 records, and since the tests run in parallel it's the culprit of the tests taking about 10 mins instead of 3).

lparrott: I totally forgot about that test limitation, let's keep it small.

annejackson:
Thanks @lparrott 
Is this ready for me to test?

lparrott:
@annejackson You could certainly start getting familiar with the changes, but we do have one area of the code to discuss that might require you to re-test the scheduled deletion portion.

annejackson:
@lparrott for Step 5- how can I generate errors in the sandbox (where this PR will not be yet). The reason I want to do this is to generate errors in the sandbox pre deploy so that after the deploy I can see that they show up in the new Errors tab (and they will show up since they will be less than 90 days). If you could give steps to get that data setup in the Sandbox before the release/this PR goes there that would be great! FYI: @xinnanlisf

src/classes/ERR_DeleteOutdated.cls
xinnanlisf: Hey @lparrott @ceiroa,

Good point! I will change it to a batch class.

xinnanlisf:
@annejackson 

Regarding creating data in sandbox, you have a couple of options:
1. You can create a tab and remove it if you want.
2. You can go to dev console, open execute window and create some records using following code:
  `List<Error__c> errors = new List<Error__c>();
  for(Integer i=0; i<n; i++) {
     Error__c error = new Error();
            errors.add(error);
   }
  insert errors;
`
if you want to edit those errors, you can go to query tab in dev console and use a simple soql to query those records: select id,(any field api you want) from error__c
3. If you are familiar with any data loader, you can also create a data file(spreadsheet etc) to load data.

Let me know if this helps. We can always jump on a call if you need further help. Thanks!

xinnanlisf:
@lparrott this PR should be ready for review again :)

xinnanlisf:
@juliesnow @aligeez 

Could you review the release notes please? Thank you!

annejackson:
I do not see this @xinnanlisf  and @lparrott 
Go to Setup->Scheduled Job, make sure 'HEDA 90-day-old Error Cleaning' is scheduled.

juliesnow:
sent you mail w/ possible edits, @xinnanlisf

juliesnow:
@xinnanlisf  and @aligeez "Changes" updated based on discussions w/ you both.

annejackson:
Testing has passed for this. Also got data setup in my sandbox for testing once this is in a release/managed package. Merging pull request. Thanks @xinnanlisf

PR #616, opened 2018-07-09
Feature/xinnanli/blank addr bug

### LINKED ISSUE #588: Blank address record is created when a Contact is created w/out any Address Information
**Steps to reproduce:**

1. Go to the Contacts tab and click New
2. Enter a value in the First Name Field
3. Enter a value in the Last Name Field
4. Scroll down to the Address Information section
5. Select any value from Primary Address Type drop-down menu
6. Click Save

**Note**: We are also experiencing the same result when creating the address from the HH account.

**Outcome:**

Blank address record is created

xinnanlisf:
# Critical Changes

# Changes
We no longer create empty address records when the Contact doesn't contain Primary Address information.

# Issues Closed
Fixes #588
# New Metadata

# Deleted Metadata

# Testing Notes
1. Go to an existing contact
2. Select a primary address type without entering other Address details
3. Click save and make sure no empty address is created


annejackson:
this looks good. Merging.

PR #623, opened 2018-07-20
Setting owner field on Account. Creating test.

### LINKED ISSUE #622: When the system automatically creates an Account for a Contact it should have the same name owner as the Contact
Especially if the Contact.OwnerId is set to a value other than the current user.

This is related to this NPSP issue: https://github.com/SalesforceFoundation/Cumulus/issues/2577. Although in our case there is no importer.

ceiroa:
# Critical Changes

# Changes
- The owner ID for an automatically-created Account will now be set to match the owner ID for the Contact.

# Issues Closed
Fixes #622. 

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/UTIL_Profile.cls
+++ b/src/classes/UTIL_Profile.cls
@@ -35,7 +35,15 @@
 */
 public class UTIL_Profile {
     /** @description System Administrator Profile Name */
-    public static final String SYSTEM_ADMINISTRATOR = 'System Administrator';
+    public static final String SYSTEM_ADMINISTRATOR {
+        public get {
+            return [SELECT name FROM Profile
+                    WHERE (UserType = 'Standard'
+                            AND PermissionsCustomizeApplication = true)
+                    ORDER BY CreatedDate ASC limit 1][0].Name;
+        }
+        private set;
+    }
 
     /** @description static instance of the current class (UTIL_Profile). It is used as the instance in a Singleton context */
     private static UTIL_Profile utilProfile;
--- a/src/classes/UTIL_Profile.cls
+++ b/src/classes/UTIL_Profile.cls
@@ -1,0 +1,99 @@
+/*
+    Copyright (c) 2018, Salesforce.org
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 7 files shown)


ceiroa:
@xinnanlisf ~ The test and the helper classes were outright copied from NPSP. So in *theory* they should have been code-reviewed already.

ceiroa:
@juliesnow Mind taking a look at the sentence under "Changes" and letting me know what you think? Thanks!

src/classes/UTIL_Profile.cls
```diff
@@ -0,0 +1,99 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2018
+* @group Utilities
+* @description Utility for Profiles
+*/
+public class UTIL_Profile {
+    /** @description System Administrator Profile Name */
+    public static final String SYSTEM_ADMINISTRATOR = 'System Administrator';
```
xinnanlisf: Have a question on this if we need to translate. Would that break this utility class?

ceiroa: That's a great question! @kofijohnson, we copied this code directly from NPSP, and it believe you were the one that wrote it. Any thoughts on Xinnan's question?

annejackson:
Should we perhaps add to changes what users would see before this fix, and also what users will see after? I am not sure what we have now for the Changes section would be fully clear. Also stating- is there any action needed? If so, what? would be good here too. @ceiroa  @juliesnow  and @xinnanlisf

ceiroa:
@xinnanlisf I modified the code. Let me know what you think. Unfortunately there's no "API Name" field on Profile (see https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_profile.htm for available fields). So I used the option listed in https://developer.salesforce.com/forums/?id=906F00000008zDiIAI. I also replaced an existing reference to "System Administrator" that existed in the code.

ceiroa:
Assigning this to you for review and merge @annejackson. The only way I can think of for testing is as follows:

- Disable ACCT_IndividualAccounts_TDTM
- Create a Contact without parent Account (a private Contact)
- Set the OwnerID field to a different User
- Re-enable ACCT_IndividualAccounts_TDTM
- Make a change in the Contact that causes ACCT_IndividualAccounts_TDTM to fire
- Verify the Owner of the automatically-created Account is the same one you set for the Contact

annejackson:
Testing has passed. Changes notes look good. Note- there is no action needed from users @aligeez  if you want to include that in any release notes in Hub.

PR #626, opened 2018-08-01
Fix for parallel test error in ADV test run.

lparrott:
# Critical Changes

# Changes

- Updating test methods to fix errors when they are run in parallel.

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

- Make sure tests methods in AFFL_ContactAccChange_TEST pass when run all at once.
- Once packaged, make sure the ADV ci_feature build can be run without HEDA Test errors. If "UNABLE_TO_LOCK_ROW" shows up in errors, this still isn't resolved.

### DIFF
--- a/src/classes/AFFL_ContactAccChange_TEST.cls
+++ b/src/classes/AFFL_ContactAccChange_TEST.cls
@@ -40,8 +40,7 @@ public with sharing class AFFL_ContactAccChange_TEST {
     private static ID orgRecTypeID;
     private static ID householdRecTypeID;
 
-    @testSetup 
-    static void dataSetup() {
+    private static void dataSetup() {
 
         UTIL_CustomSettingsFacade.getSettingsForTests(new Hierarchy_Settings__c(Account_Processor__c = UTIL_Describe_API.getHhAccRecTypeID()));
 
@@ -71,7 +70,7 @@ public with sharing class AFFL_ContactAccChange_TEST {
     
     @isTest
     public static void changeContact() {
-
+        dataSetup();
         Account bizOrg1 = [SELECT Id FROM Account WHERE Name = 'Acme'];
         
         //Confirm Primary Business Organization field has been populated in contact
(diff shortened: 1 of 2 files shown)


src/classes/AFFL_ContactAccChange_TEST.cls
```diff
@@ -40,8 +40,7 @@ public with sharing class AFFL_ContactAccChange_TEST {
     private static ID orgRecTypeID;
     private static ID householdRecTypeID;
 
-    @testSetup 
-    static void dataSetup() {
+    private static void dataSetup() {
```
ceiroa: This is very interesting, @lparrott! Just curious, how were you able to figure out this is what was causing the tests to fail in parallel?

fyi, @xinnanlisf - Best practice for all unit tests. Apparently setting the custom settings in a method annotated with @testSetup causes row lock errors.

src/package.xml
```diff
@@ -113,6 +113,8 @@
         <members>UTIL_Describe_TEST</members>
         <members>UTIL_Namespace</members>
         <members>UTIL_Namespace_TEST</members>
+        <members>UTIL_Profile</members>
+        <members>UTIL_Profile_TEST</members>
```
ceiroa: Why would these two be included now?

lparrott: Since these classes were added to the repo, the deployment fails without them in the package.

src/classes/AFFL_ContactAccChange_TEST.cls
lparrott: The issue had to do with all of these methods trying to edit the same shared Contact, since tests run in parallel, they were receiving the UNABLE_TO_LOCK_ROW error.

Another way to fix this would be create a new Contact for each test, but there appears to be some setup involved, so I took the fastest approach.

This article tipped me of to the root problem: https://salesforce.stackexchange.com/questions/8058/unable-to-lock-row-error-while-running-all-apex-test-classes

src/package.xml
ceiroa: Right, that makes sense. I'm just a little surprised the PR where they got included (https://github.com/SalesforceFoundation/HEDAP/pull/623/files) passed build with the classes being there but not listed in package.xml...

src/classes/AFFL_ContactAccChange_TEST.cls
ceiroa: Cool. And we do have a class for creating uniquely-named Contacts, just fyi. 

I'm surprised this didn't fail sooner, but good to have it straighten now.

ceiroa:
@jstvz This is a good example of the need for being able to specify a product, ADV in this case, depending on the beta version of another product. This will unblock a PR in ADV (#44), but not until a prod release is cut, in the current setting. Just FYI.

ceiroa:
@lparrott I just realized why the build is taking so long. The branch name doesn't start with /feature, so the build system is not picking it up! It needs to be renamed.

PR #627, opened 2018-08-02
Fix for parallel test error in ADV test run.

lparrott:
# Critical Changes

# Changes

- Updating test methods to fix errors when they are run in parallel.

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

- Overrides #626 
- Make sure tests methods in AFFL_ContactAccChange_TEST pass when run all at once.
- Once packaged, make sure the ADV ci_feature build can be run without HEDA Test errors. If "UNABLE_TO_LOCK_ROW" shows up in errors, this still isn't resolved.

### DIFF
--- a/src/classes/AFFL_ContactAccChange_TEST.cls
+++ b/src/classes/AFFL_ContactAccChange_TEST.cls
@@ -40,8 +40,7 @@ public with sharing class AFFL_ContactAccChange_TEST {
     private static ID orgRecTypeID;
     private static ID householdRecTypeID;
 
-    @testSetup 
-    static void dataSetup() {
+    private static void dataSetup() {
 
         UTIL_CustomSettingsFacade.getSettingsForTests(new Hierarchy_Settings__c(Account_Processor__c = UTIL_Describe_API.getHhAccRecTypeID()));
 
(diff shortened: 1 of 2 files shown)


ceiroa:
@lparrott Hmm, I'd almost prefer to keep the other one and rename the branch. Otherwise we lose the history in the PR. Not sure how best to do it from the command line, but with SourceTree is really easy to rename branches:

screen shot 2018-08-02 at 1 10 19 pm

And if it can be done from SourceTree it can be done from the command line :-).

ceiroa:
This is good to go, @annejackson. There is no HEDA work item for it because we found the issue in ADV. As Luke says above, it was just an issue with the unit tests, and there shouldn't be anything to test manually.

ceiroa:
The change applies to HEDA too (though I'm surprised we didn't see it before - tests in HEDA feature org are also supposed to run in parallel, @xinnanlisf @jstvz), but it's true that it's not meaningful to users that we  can run tests in parallel.

annejackson:
OK, checking out Carlos' message above how do we think the changes section sounds? 
@juliesnow @lianeyee  and @jacebryan

lianeyee:
Made a small edit @annejackson to clarify that "they" means the tests (not the methods or the errors).

PR #632, opened 2018-08-14
Turn off custom setting for unit tests by default

xinnanlisf:
# Critical Changes

# Changes
- We've made a change so that the Affiliation_Record_Type_Enforced__c field will be set false by default. There should be no change for your existing HEDA org settings.

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
Make sure unit tests pass,
Check the Affiliation_Record_Type_Enforced__c field and that it is set to false by default.
Also make sure this ticket still works- W-023536
Notes- Turn off Affiliation_Record_Type_Enforced__c field for default settings(in both UTIL_CustomSettingsFacade.configSettings() and in Hierarchy_Settings__c the Affiliation_Record_Type_Enforced__c field is being set to 'true' by default). This is so the SAL's team unit tests can run again.

### DIFF
--- a/src/classes/AFFL_AccChange_TDTM.cls
+++ b/src/classes/AFFL_AccChange_TDTM.cls
@@ -83,7 +83,7 @@ public with sharing class AFFL_AccChange_TDTM extends TDTM_Runnable {
                    		lookupFieldNameOld = AFFL_MultiRecordType_TDTM.afflMapper.contactLabelNames.get(lookupFieldLabelOld);
                		}
 
-               		if (isAfflTypeEnforced) {
+               		if (isAfflTypeEnforced == true) {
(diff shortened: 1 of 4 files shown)


xinnanlisf:
@lparrott yeah all my unit tests are setting that value anyway so..

annejackson:
Please keep me posted when ready @xinnanlisf  and @lparrott. Also, I want to make sure I know what to test/check for the whole value being set to either "false" or "null" by default that Kevin S. brought up.

lparrott:
@xinnanlisf Changes look good, but we hit the dreaded "Apex CPU time limit exceeded" during tests.

xinnanlisf:
now worries, I just rerun it. Should be good now!

annejackson:
I tried a shot at the "Changes" section so that it made sense to users. Anything else we need to say? @lparrott  @xinnanlisf and @jacebryan

annejackson:
Merged.

PR #634, opened 2018-08-22
SOQL Optimization limiting # of records returned by nested queries to…

### LINKED ISSUE #635: Cannot Delete Triggers retrieving more records than necessary
There exist triggers on Account, Contact, Program Plan, and Plan Requirement that block deletion of records if one or more records are present on key child objects.

Currently, the child queries are not bounded.  This means the more children that are present, the more they will retrieve.  This could impact a bulk deletion operation and/or have minor performance impact.

Recommend adding limit 1 to all subqueries in these triggers (see below for list of triggers) so that it never retrieves more than one child record per subquery, as that is enough to confirm the deletion cannot be allowed to proceed.

Trigger List:
* ACCT_CannotDelete_TDTM
* CON_CannotDelete_TDTM
* PPlan_CannotDelete_TDTM
* PReq_CannotDelete_TDTM

j-brandolini:
# Critical Changes

# Changes
- To improve system performance, we've modified the CannotDelete triggers for Accounts, Contacts, Program Plans, and Program Requirements so they now retrieve only one record per related object.

# Issues Closed
Fixes #635 

# New Metadata

# Deleted Metadata

# Testing Notes
Reran all tests in unmanaged environment.  All passed.


xinnanlisf:
Hi John, thanks for your contributions! I will take a look soon. :)

ceiroa:
Great enhancement. Thank you, @j-brandolini! Just wondering, do you think it be better/worse to use ```COUNT``` instead of ```LIMIT 1``` performance-wise, or just the same?

j-brandolini:
Hey Carlos, 

Great question!  I think that, in the context of a trigger specifically (but probably in other contexts as well), limit is the safer way to go-- unless you know for sure your total records will always squeak in under governor limits.  It counts the records returned by count(), in terms of the # of records retrieved for governor limits.  

And if you compare, a use case with one main record, and 10 children on one object, and 5 on another child, that becomes 16 records retrieved versus 3 if you limit what is returned...you could also use limit and count, as another option...then it becomes a 1 or 0...  

Performance should only again matter if you start ordering the records or adding additional constraints, but IMO it would still be better than retrieving all of them.

John

ceiroa:
Thanks for the clarification, @j-brandolini!

Forgot to ask you, would you mind creating an Issue with some info and adding a "Fixes #IssueID" under the Issues Closed section of the PR? Thank you!

j-brandolini:
@ceiroa - Hi Carlos, Did you reopen this because there was a problem, or run into something from the merge?  I noticed it was closed, and then reopened and was curious. - John

ceiroa:
Hi @j-brandolini. It's all good. I created a separate PR after pulling your branch into a branch in this repo. That preserves your commits (and shows you are the contributor), but allows the build system to run it. And it the result of the build is showing here too. I was going to close this one and keep only the other one, but no need. Thanks for checking!

annejackson:
Hi @j-brandolini  - I would love to get your thoughts on how to best test this in a brand new, fairly empty scratch org? Here are my thoughts so far- use dev console to create a lots of records, and then do whatever is needed to TDTM to test that part. Looks like this one is the actual ticket to merge.

j-brandolini:
@annejackson - I replied on the other pull request (#641) as well, but wanted to reach out and see if you had what you needed, or if you were looking for more information.

annejackson:
Thanks @j-brandolini - got it! I think I am good for now. Do you mind checking out the text under the "Changes" section (I also pasted it below)? Do you think it would make sense to technical and non-technical heda users? Feel free to go ahead and revise it to what you think best and then we will check it out with the team when testing.
**Changes**
"CannotDelete Trigger Modification to add limit of 1 to nested soql, as it only needs to know if at least one record is present, rather than getting all of them. This is to keep performance, etc. linear in the event of several Program or Course Enrollments, for example.)"

j-brandolini:
@annejackson -Feel free to further modify/trim, but here are my suggested edits:
> CannotDelete Trigger Modification to limit the number of record retrieved to one per related record (added a limit of 1 to nested soql), as it only needs to know if at least one record is present, so retrieving all of them is unnecessary.  This is to keep performance, etc. linear in the event of several Program or Course Enrollments, for example.  Note:  if there is more than one related object type to check, it will look for a limit of 1 of each type of record.
>

annejackson:
Thanks @j-brandolini - I put your suggested edit up above and we will review. We will get to this item a little bit later.

ceiroa:
Hey @annejackson ~ I just remembered this is the PR that will never pass build because it's from an external repo. The internal equivalent is https://github.com/SalesforceFoundation/HEDAP/pull/634. That's the one we need to merge. I'll put you as reviewer (someone other than me needs to approve it since I opened it).

Issue #604, opened 2018-05-09
MCMP Story 8.0: Update the status of Program Plan when new Primary is selected


aligeez:
Currently the status of "Current" does not change on the Program Plan when a new Primary Program Plan is assigned for the same Account, This should update the newly elected Program Plan as "Current" and make the other ones "former".

**User Acceptance Criteria:**
1.0 - Go to an existing Account (Academic Program) and add a Program Plan as primary (if it does not already have one). Save.
1.1 - Observe that the newly elected Program Plan is set as "current" on the Account. 
2.0 - Then go back to the same Account and change the Primary Plan to a different one. 
2.1 - Observe that the status of the newly elected plan is updated to "Current".

aligeez:
@ceiroa - this already exists as WI - W-024568, apologies for creating them out of order, i still don't have Lurch capabilities. Can you please link these?

ceiroa:
**lurch: attach W-024568

j-brandolini:
I have a couple of clarifying questions on this one:

1) I assume if a new/different Program Plan associated with an Account is marked as "Current", then it would uncheck the "Is Primary: indicator on the other Program Plans also associated with that Account [Should only be more than one in merge or prior data scenario].
2) Should it leave the Status set to  "Current" on the records it is removing the "Is Primary" indicator from, or should it also set the Status to "Archived"?
3) Should there only ever be one "Current" Program Plan, and should any others should associated with the same Account always be "Archived"?  In that case, Current also means "Is Primary" right?  If the record's status  is changed to "Archived", should it also uncheck the "Is Primary" indicator?  

If the "Is Primary" indicator is essentially always dependent on the Status, recommend adding "Before" Trigger events to maintain the field values:
- New Records
-- If Status == Current, set "Is Primary" to true (checked)
-- Else set "Is Primary" to false (unchecked)
- Updated Records
-- If (New Status != Old Status) and (Status == Current), set Primary to true (checked)
-- Else If (New Status != Old Status) and (Status != Current) set Primary to false (unchecked)

BeforeInsert/BeforeUpdate trigger to maintain the Primary indicator and set it based on changes to the Status field, and it's current value.  

[...]

PR #637, opened 2018-08-23
Makes Program Plan Status Current if Is Primary is Checked

j-brandolini:
# Critical Changes

# Changes
* Moves logic to mark other plan records associated with same account non primary from Before Insert / Before Update to After Insert / After Update.  Best practice is to update same records in Before Insert/Update trigger contect, and other records in After Insert/Update context
* Adds Before Insert / Before Update logic to set status to Current for records marked as "Primary" (Is Primary checked)
* Updated Trigger Default Config to reflect existence of After Update, After Insert for PPlan_Primary_TDTM trigger
* Defined Register Trigger values for the new trigger combinations (PPlan_Primary_TTDM_AfterInsert, AfterUpdate)
* Added additional tests to cover the logic to make Status Current for new "Primary" Program Plans, as well as Primary Program Plan records newly updated to "Primary"

# Issues Closed
Addresses #604 
# New Metadata

# Deleted Metadata

# Testing Notes
* Confirmed new and existing tests all ran successfully, unit tested as well for New and Updates, as well as existing trigger functionality

### DIFF
--- a/src/classes/PPlan_Primary_TDTM.cls
+++ b/src/classes/PPlan_Primary_TDTM.cls
@@ -50,7 +50,7 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
     * @param objResult the describe for Accounts
     * @return dmlWrapper.
     ********************************************************************************************************/
-        public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist,
+    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist,
     TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
         DmlWrapper dmlWrapper = new DmlWrapper();
 
--- a/src/classes/PPlan_Primary_TDTM.cls
+++ b/src/classes/PPlan_Primary_TDTM.cls
@@ -35,6 +35,13 @@
 * @description Handles primary program plans under accounts
 */
 public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
+   
+    /*******************************************************************************************************
+    * @description The Plan status value of Current.
+    */
+    @testVisible private static final String STATUS_CURRENT_VAL = 'Current';
+	@testVisible private static final String STATUS_ARCHIVED_VAL = 'Archived';
+   
     /*******************************************************************************************************
     * @description Handles primary program plans.
     * @param listNew the list of Accounts from trigger new.
@@ -43,7 +50,7 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
     * @param objResult the describe for Accounts
     * @return dmlWrapper.
     ********************************************************************************************************/
-    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist,
+        public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist,
     TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
         DmlWrapper dmlWrapper = new DmlWrapper();
 
@@ -54,24 +61,47 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
             for (integer i=0; i<newlist.size(); i++) {
                 Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
                 if (newPPlan.Is_Primary__c == true) {
+                    // Make Plan "Current" if "Is Primary" is checked.
+					newPPlan.Status__c = STATUS_CURRENT_VAL;
+                }
+            }
+        }
+        // AFTER INSERT
+        if ( newlist != null && triggerAction == TDTM_Runnable.Action.AfterInsert && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_After_Insert)) {
+            for (integer i=0; i<newlist.size(); i++) {
+                Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
+                if (newPPlan.Is_Primary__c == true) {
                     //if multiple program plans under one single account are set as primary in same transation
                     //Use the first program plan as the primary
-                    if (!acccountIdsNeedNonPrimarycount.contains(newPPlan.Account__c)) {
+					if (!acccountIdsNeedNonPrimarycount.contains(newPPlan.Account__c)) {
                         primaryProgramIds.add(newPPlan.Id);
                     }
                     acccountIdsNeedNonPrimarycount.add(newPPlan.Account__c);
                 }
             }
         }
-        // BEFORE UPDATE
+ 		// BEFORE UPDATE
         if ( newlist != null && triggerAction == TDTM_Runnable.Action.BeforeUpdate && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_Before_Update)) {
+           for (integer i=0; i<newlist.size(); i++) {
+                Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
+                Program_Plan__c oldPPlan = (Program_Plan__c)oldlist[i];
+                if (newPPlan.Is_Primary__c == true&& newPPlan.Is_Primary__c != oldPPlan.Is_Primary__c) {
+					// Make Plan "Current" if "Is Primary" is checked.
+					newPPlan.Status__c = STATUS_CURRENT_VAL;
+                }
+            }
+
+        }
+		System.debug('Groucho');
+        // AFTER UPDATE
+        if ( newlist != null && triggerAction == TDTM_Runnable.Action.AfterUpdate && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_After_Update)) {
             for (integer i=0; i<newlist.size(); i++) {
                 Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
                 Program_Plan__c oldPPlan = (Program_Plan__c)oldlist[i];
                 if (newPPlan.Is_Primary__c == true
                     && (newPPlan.Is_Primary__c != oldPPlan.Is_Primary__c
                         || newPPlan.Account__c != oldPPlan.Account__c)) {
-                    //if multiple program plans under one single account are set as primary in same transation
+					//if multiple program plans under one single account are set as primary in same transation
                     //Use the first program plan as the primary
                     if (!acccountIdsNeedNonPrimarycount.contains(newPPlan.Account__c)) {
                         primaryProgramIds.add(newPPlan.Id);
@@ -80,13 +110,19 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
                 }
             }
         }
+        
         if (acccountIdsNeedNonPrimarycount.size()>0) {
             dmlWrapper.objectsToUpdate.addAll(makeOtherPPlanNotPrimary(acccountIdsNeedNonPrimarycount, primaryProgramIds));
         }
+		
         if (triggerAction == TDTM_Runnable.Action.BeforeUpdate) {
             TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_Before_Update, false);
         }else if (triggerAction == TDTM_Runnable.Action.BeforeInsert) {
             TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_Before_Insert, false);
+        }else if (triggerAction == TDTM_Runnable.Action.AfterUpdate) {
+            TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_After_Update, false);
+        }else if (triggerAction == TDTM_Runnable.Action.AfterInsert) {
+            TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_After_Insert, false);
         }
         return dmlWrapper;
     }
--- a/src/classes/PPlan_Primary_TDTM.cls
+++ b/src/classes/PPlan_Primary_TDTM.cls
@@ -40,7 +40,10 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
     * @description The Plan status value of Current.
     */
     @testVisible private static final String STATUS_CURRENT_VAL = 'Current';
-	@testVisible private static final String STATUS_ARCHIVED_VAL = 'Archived';
+    /*******************************************************************************************************
+    * @description The Plan status value of Archived.
+    */
+    @testVisible private static final String STATUS_ARCHIVED_VAL = 'Archived';
    
     /*******************************************************************************************************
     * @description Handles primary program plans.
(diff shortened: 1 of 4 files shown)


j-brandolini:
FWIW - I was originally going to do a different item.  I thought I had updated my branch before I started but apparently forgot to, so the branch name does not reflect the change, though I updated the pull request to.  If you want me to redo under a different branch, I can.

src/classes/PPlan_Primary_TDTM.cls
```diff
@@ -51,27 +58,50 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
         Set<Id> primaryProgramIds = new Set<Id>();
         // BEFORE INSERT
         if ( newlist != null && triggerAction == TDTM_Runnable.Action.BeforeInsert && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_Before_Insert)) {
+            for (integer i=0; i<newlist.size(); i++) {
+                Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
+                if (newPPlan.Is_Primary__c == true) {
+                    // Make Plan "Current" if "Is Primary" is checked.
+					newPPlan.Status__c = STATUS_CURRENT_VAL;
```
lifewithryan: Just another spacing thing.

src/classes/PPlan_Primary_TDTM.cls
```diff
@@ -51,27 +58,50 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
...
+        }
+        // AFTER INSERT
+        if ( newlist != null && triggerAction == TDTM_Runnable.Action.AfterInsert && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_After_Insert)) {
             for (integer i=0; i<newlist.size(); i++) {
                 Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
                 if (newPPlan.Is_Primary__c == true) {
                     //if multiple program plans under one single account are set as primary in same transation
                     //Use the first program plan as the primary
-                    if (!acccountIdsNeedNonPrimarycount.contains(newPPlan.Account__c)) {
+					if (!acccountIdsNeedNonPrimarycount.contains(newPPlan.Account__c)) {
```
lifewithryan: Looks like this line is the same except for spacing. I'm thinking the IDE is sometimes using tabs and other time using spaces. I think we try to ensure that tabs are converted to spaces via the IDE. In any case, just verify spacing.

src/classes/PPlan_Primary_TDTM.cls
```diff
@@ -51,27 +58,50 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
...
+                Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
+                Program_Plan__c oldPPlan = (Program_Plan__c)oldlist[i];
+                if (newPPlan.Is_Primary__c == true&& newPPlan.Is_Primary__c != oldPPlan.Is_Primary__c) {
+					// Make Plan "Current" if "Is Primary" is checked.
+					newPPlan.Status__c = STATUS_CURRENT_VAL;
+                }
+            }
+
+        }
+		System.debug('Groucho');
```
lifewithryan: Hah! I like 'Groucho' but sadly he must go :)

src/classes/PPlan_Primary_TDTM.cls
```diff
@@ -51,27 +58,50 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
...
         }
-        // BEFORE UPDATE
+ 		// BEFORE UPDATE
         if ( newlist != null && triggerAction == TDTM_Runnable.Action.BeforeUpdate && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_Before_Update)) {
+           for (integer i=0; i<newlist.size(); i++) {
+                Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
+                Program_Plan__c oldPPlan = (Program_Plan__c)oldlist[i];
+                if (newPPlan.Is_Primary__c == true&& newPPlan.Is_Primary__c != oldPPlan.Is_Primary__c) {
+					// Make Plan "Current" if "Is Primary" is checked.
+					newPPlan.Status__c = STATUS_CURRENT_VAL;
```
lifewithryan: verify spacing

src/classes/PPlan_Primary_TDTM.cls
```diff
@@ -51,27 +58,50 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
...
                 }
             }
         }
-        // BEFORE UPDATE
+ 		// BEFORE UPDATE
         if ( newlist != null && triggerAction == TDTM_Runnable.Action.BeforeUpdate && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_Before_Update)) {
+           for (integer i=0; i<newlist.size(); i++) {
+                Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
+                Program_Plan__c oldPPlan = (Program_Plan__c)oldlist[i];
+                if (newPPlan.Is_Primary__c == true&& newPPlan.Is_Primary__c != oldPPlan.Is_Primary__c) {
```
lifewithryan: I know this one is nit-picky but can you add a space between 'true&&' so that it is 'true &&'

src/classes/PPlan_Primary_TDTM.cls
```diff
@@ -51,27 +58,50 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
...
+        // AFTER UPDATE
+        if ( newlist != null && triggerAction == TDTM_Runnable.Action.AfterUpdate && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_After_Update)) {
             for (integer i=0; i<newlist.size(); i++) {
                 Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
                 Program_Plan__c oldPPlan = (Program_Plan__c)oldlist[i];
                 if (newPPlan.Is_Primary__c == true
                     && (newPPlan.Is_Primary__c != oldPPlan.Is_Primary__c
                         || newPPlan.Account__c != oldPPlan.Account__c)) {
-                    //if multiple program plans under one single account are set as primary in same transation
+					//if multiple program plans under one single account are set as primary in same transation
```
lifewithryan: No big deal, just spacing here

src/classes/PPlan_Primary_TDTM.cls
```diff
@@ -80,13 +110,19 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
...
+        
         if (acccountIdsNeedNonPrimarycount.size()>0) {
             dmlWrapper.objectsToUpdate.addAll(makeOtherPPlanNotPrimary(acccountIdsNeedNonPrimarycount, primaryProgramIds));
         }
+		
         if (triggerAction == TDTM_Runnable.Action.BeforeUpdate) {
             TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_Before_Update, false);
         }else if (triggerAction == TDTM_Runnable.Action.BeforeInsert) {
             TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_Before_Insert, false);
+        }else if (triggerAction == TDTM_Runnable.Action.AfterUpdate) {
```
lifewithryan: I'm not going to have you change this as it looks like that style was there before, but want to call this to Carlos (@ceiroa): someone should perhaps verify that our else blocks don't really look like this. i think its hard to read with that indentation (maybe its just GitHub).

src/classes/PPlan_Primary_TEST.cls
```diff
@@ -142,4 +142,77 @@ private class PPlan_Primary_TEST {
...
+                System.assertEquals(pPlans[0].Id, pPlan.Id);
+				System.assertEquals(PPlan_Primary_TDTM.STATUS_CURRENT_VAL, pPlan.Status__c);
+				cntPrimary++;
+            }else {
+                system.assertEquals(PPlan_Primary_TDTM.STATUS_ARCHIVED_VAL, pPlan.Status__c);
+            }
+        }
+		System.assertEquals (1, cntPrimary, 'One and only one primary program plan expected');
+	}
+
```
lifewithryan: Rather than put a comment on every instance of spacing here, I'll save time and just have you verify that we have the spacing issues looked at. I don't want to overwhelm you with comments :)

j-brandolini:
@lifewithryan - Thanks, I will address these tonight and update back when complete.

j-brandolini:
@lifewithryan - I included your feedback, and added one more comment I felt was missing for the STATUS_ARCHIVED_VAL variable.  Cheers!

src/classes/PPlan_Primary_TDTM.cls
```diff
@@ -35,6 +35,16 @@
 * @description Handles primary program plans under accounts
 */
 public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
+   
+    /*******************************************************************************************************
+    * @description The Plan status value of Current.
+    */
+    @testVisible private static final String STATUS_CURRENT_VAL = 'Current';
```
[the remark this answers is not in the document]
ceiroa: The more I think about it, the more it seems to me making it configurable is the right way to go. Then the user can select under HEDA settings which statuses to use.

src/classes/PPlan_Primary_TDTM.cls
```diff
@@ -51,6 +61,16 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
...
+            for (integer i=0; i<newlist.size(); i++) {
+                Program_Plan__c newPPlan = (Program_Plan__c)newlist[i];
+                if (newPPlan.Is_Primary__c == true) {
+                    // Make Plan "Current" if "Is Primary" is checked.
+                    newPPlan.Status__c = STATUS_CURRENT_VAL;
+                }
+            }
+        }
+        // AFTER INSERT
+        if ( newlist != null && triggerAction == TDTM_Runnable.Action.AfterInsert && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_After_Insert)) {
```
ceiroa: The logic below was happening on BeforeInsert previously. Why are we changing it to AfterInsert?

Also, if it's now done in AfterInsert, wouldn't we need to add the records to DmlWrapper?

src/classes/PPlan_Primary_TDTM.cls
```diff
@@ -65,6 +85,18 @@ public with sharing class PPlan_Primary_TDTM extends TDTM_Runnable {
...
+                Program_Plan__c oldPPlan = (Program_Plan__c)oldlist[i];
+                if (newPPlan.Is_Primary__c == true && newPPlan.Is_Primary__c != oldPPlan.Is_Primary__c) {
+                    // Make Plan "Current" if "Is Primary" is checked.
+                    newPPlan.Status__c = STATUS_CURRENT_VAL;
+                }
+            }
+
+        }
+        // AFTER UPDATE
+        if ( newlist != null && triggerAction == TDTM_Runnable.Action.AfterUpdate && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.PPlan_Primary_TDTM_After_Update)) {
```
ceiroa: Same as above. Why are we changing this logic to be performed on AfterUpdate instead BeforeUpdate, and whether we need to use DmlWrapper in this case.

ceiroa: If it's true we needed to use DmlWrapper, why weren't the tests catching this?

ceiroa:
No worries, @j-brandolini. Take your time. And let me know if the DmlWrapper observation makes sense. I haven't fully verified the behavior the way it is now, but I'm surprised we don't need it. You can take a look at https://powerofus.force.com/articles/Resource/Deploy-Custom-Apex-in-TDTM-for-NPSP and http://developer.salesforce.org/#blog/post/2016/06/10/table-driven-trigger-management.html for more info. Thank you!

j-brandolini:
Hi Carlos, 

I should have used DmlWrapper, returning the list of records to update from the run method (this was my first time working with TDTM— I will make sure I understand all it’s implementation rules going forward.  Looking closer at the TDTM code, it makes sense....that while what I did works, it is basically not doing it the way it should be.  I will refactor this, and the other items you mentioned, such as the pieces that should be configurable.  I will probably not get to this until Wednesday evening, but will update, and tag you when I do!  Thanks again for the great feedback!!

ceiroa:
Actually, after taking another look at this, I realized that DmlWrapper is in fact being properly used even after the changes. Sorry for the confusion. It seems like that code was adding records to DmlWrapper on BeforeInsert and BeforeUpdate, even though it didn't need to in that case. Since that was already happening, moving the logic to AfterInsert and AfterUpdate didn't cause any issues. We've kind of killed another bird with the same stone :-).

So only making those status configurable is necessary. Also, minor thing, would you mind adding Test.startTest and Test.stopTest to the tests?

Thank you!

aligeez:
@j-brandolini - I spoke to @ceiroa today and we agreed that we should have a Settings option for this (#637  ) as well as (#604 ), but it would need a new tab under the "HEDA Settings" tab. We will name this Tab "Programs" to keep it neutral and leave room to grow if we ever add more settings in the future to Program Enrollments and/or Program Plans. 

image

@juliesnow - we may need help with the text-on-page that appears with these 2 new settings options under HEDA settings. Something along the lines of: 

**parent tab**: "Programs"

**Sub-heade**r: "Program Plan Settings"

**First setting** : "Select this to option to make the Program Plan status 'current' if 'Primary' is checked:"
**Second setting** :  _@ceiroa - what did we say the second option should say? Both #604 and #637 look similar to me._

ceiroa:
Hi again @j-brandolini ~ Sorry for the back and forth on this, but it looks like the string ```STATUS_ARCHIVED_VAL``` is not being used, other than in the tests. Shouldn't this be the status that all Program Plans other than the Current/Primary get?

Also, just to be explicit, I imagine this would work a bit similar as primary Affiliations work so that if a Current/Primary Program Plan exist and a new one is made Current/Primary, the previous one is changed to Archived/Non-Primary. (Please confirm, @aligeez).

ceiroa:
@aligeez ~ This "Archived" status is the second option. Basically, the status to give to those Program Plans that aren't primary.

j-brandolini:
Hi @ceiroa , 

I added/defined the STATUS_ARCHIVED_VAL field in the Trigger, because it seemed like a better place for it (with the other status [STATUS_CURRENT_VAL])-- I know it was not used in the class, but my thinking was that it probably would be eventually, and it would be more consistent to have them together.  Additionally, I thought we would eventually modify/add code to archive, as well.  That logic was not yet defined, so I did not implement further.  It was one of the items I indicated would not be addressed on the issue itself.  

For the future: What process should I follow to more explicitly indicate when a pull request may not cover all items in an issue?  Archiving was open/waiting for further clarification/requirements, but would be added on later if confirmed as an automatic feature.

Thanks!

John

ceiroa:
Oh sorry, I missed that. The issue is that once a PR gets merged into master it goes into the next release. So the functionality needs to be complete enough that it can stand by itself once it's in the product. I think we need all this functionality to go in at the same time, even if that means getting additional requirements from @aligeez. Hope that makes sense...

j-brandolini:
@ceiroa, totally get it.  At the time, I made @aligeez aware of the question via the issue comments, but felt that since implementing program archiving was not part of the issue requirements/success criteria, and since it could be done later when the requirements were determined as a separate issue/enhancement that is was ok to proceed.  I indicated this in the issue comments to Ali, but probably should have brought it forward to the pull request as well.  Or followed up for more feedback.  Or both.

aligeez:
@ceiroa & @j-brandolini  - Apologies on the delay, still catching up post-Dreamforce. Is there anything outstanding with this ticket that is waiting on me at this point? Are we all set?

j-brandolini:
Hi @aligeez, 

Other than confirming whether other program archival behavior is required, I do not believe so.

John

jbrandolinisf:
Closing this given the decisions made regarding the work item.

PR #641, opened 2018-09-07
Feature/j brandolini/update delete triggers

### LINKED ISSUE #635: Cannot Delete Triggers retrieving more records than necessary

ceiroa:
# Critical Changes

# Changes
- To improve system performance, we've modified the CannotDelete triggers for Accounts, Contacts, Program Plans, and Program Requirements so they now retrieve only one record per related object.

# Issues Closed
Fixes #635 

# New Metadata

# Deleted Metadata

# Testing Notes
Reran all tests in unmanaged environment.  All passed.


annejackson:
Hi @j-brandolini - I would love to get your thoughts on how to best test this in a brand new, fairly empty scratch org? Here are my thoughts so far- use dev console to create a lots of records, and then do whatever is needed to TDTM to test that part.

j-brandolini:
Hi Anne,

That makes sense.  For each of the triggers, generically, I would test the
following:

1) Deletion with no relevant related records present.  Result:  Allow
deletion
2) Deletion with 1 relevant related record present.  Result: Block deletion
3) Deletion with many related records present.  Result: Block deletion

I am sure your test cases already call it out, but some of these block
across two or three related objects.  You would probably want to check that
deletion is blocked for all of them.  The performance for 2 and 3 above
should be essentially the same, coming back quickly with whatever our
"Nope, can't do it because..." message is.  As always, scenarios like 1 or
more records present in any of the relevant child objects...so if their are
three blocking related objects, we should check all three individually and
collectively (even though I know they will work)!!!

Let me know if you need more info, or if you need the which objects for
which tests...like I said, certain you already have it well in hand, but I
like to ask anyways...

Hope this helps!!!

John

j-brandolini:
Hi @annejackson - You could also export from an existing org, create the parent record, update the id in the export file for the new parent, and do a dataloader import, following similar test scenarios as outlined above.

annejackson:
Testing complete. Merging PR.

PR #666, opened 2018-12-03
Feature/rheadley/fix unit tests for shield

### LINKED ISSUE #650: Shield: Unit Test SOQL queries in HEDA are limiting the ability for certain fields to be encrypted
There are a number of HEDA unit tests with SOQL queries that break (hinder) platform encryption. As a general rule, if a field is to be encrypted using Shield, the field cannot be used as a filter in a WHERE clause nor can it be used in an ORDER BY clause. Since these queries are in TEST classes, it should be fairly easy to modify the queries to eliminate use of those fields.

The affected fields and test classes are:

_Standard fields_
- *Account.Name*
    -     ADDR_HouseholdAcc_TEST, line 2181
    -     ADDR_OtherAcc_TEST, line 107, 111, 154, 159
    -     AFFL_ContactAccChange_TEST, line 71, 107, 125, 126, 150, 151, 176
    -     AFFL_MultiRecordType_TEST, line 284
- *Contact.Name*
    -     ADDR_Contact_TEST, line 980, 984, 1026, 1031
    -     AFFL_ContactAccChange_TEST, line 74, 75, 106, 124, 148, 175
    -     CON_Preferred_TEST, line 57, 99
- *Contact.Email*
    -     CON_Email_TEST, line 105

_HEDA Fields_

- *Address.MailingStreet* is referenced  in SOQL queries in HEDA unit test code
   -     ADDR_HouseholdAcc_TEST
- *Course.Description* is referenced  in SOQL queries in HEDA unit test code
   -     COUR_DescriptionCopy_TEST
   -     STG_Courses_TEST
- *Course_Offering.SectionID* is in SOQL queries in HEDA unit test code
   -     TERM_CourseOff_TEST, line 94

lifewithryan:
# Critical Changes
There are no critical changes to this codebase

# Changes

- Unit-Test pre-work for eventual compatibility with Shield encryption.

# Issues Closed
Fixes #650 

# New Metadata
No new metadata

# Deleted Metadata
No deleted Metadata

# Testing Notes
Checkout and ensure Unit tests still pass

### DIFF
--- a/src/classes/ADDR_HouseholdAcc_TEST.cls
+++ b/src/classes/ADDR_HouseholdAcc_TEST.cls
@@ -1851,9 +1851,16 @@ public with sharing class ADDR_HouseholdAcc_TEST {
         Test.stopTest();
         
         // verify the Address objects split street.
-        List<Address__c> listAddr = [select Id, MailingStreet__c, MailingStreet2__c, Formula_MailingStreetAddress__c from Address__c 
-        where MailingStreet__c = 'new street'];
+        List<Address__c> addresses = [SELECT Id, MailingStreet__c, MailingStreet2__c, Formula_MailingStreetAddress__c from Address__c];
+        
+        List<Address__c> listAddr = new List<Address__c>();
+        for(Address__c addy : addresses) {
+            if(addy.MailingStreet__c == 'new street') {
+                listAddr.add(addy);
+            }
+        }
         System.assertEquals(2, listAddr.size());
+
         for (Address__c addr : listAddr) {
             System.assertEquals('new street', addr.MailingStreet__c);
             System.assertEquals('second line', addr.MailingStreet2__c);
@@ -2178,7 +2185,9 @@ public with sharing class ADDR_HouseholdAcc_TEST {
         Test.stopTest();
 
         // verify results
-        Account acc = [select Id, Name, BillingStreet from Account where Name like '%Household%'];
+        String hhAccRecTypeID = UTIL_Describe_API.getHhAccRecTypeID();
+        //Account acc = [select Id, Name, BillingStreet from Account where Name like '%Household%'];
+        Account acc = [select Id, Name, BillingStreet from Account where RecordTypeId = :hhAccRecTypeID];
         UTIL_Debug.debug('****Parent account: ' + JSON.serializePretty(acc));
         System.assertNotEquals(null, acc);
         List<Address__c> listAddr = [select Id, Default_Address__c, MailingStreet__c from Address__c];
--- a/src/classes/CON_Preferred_TEST.cls
+++ b/src/classes/CON_Preferred_TEST.cls
@@ -54,12 +54,19 @@ public with sharing class CON_Preferred_TEST {
 		insert contacts;
 		Test.stopTest();
 		
-		contacts = [select Email, Phone, UniversityEmail__c, WorkEmail__c, AlternateEmail__c, HomePhone, WorkPhone__c, OtherPhone, MobilePhone 
-					from Contact where ID in :contacts order by LastName];
-		contact1 = contacts[0];
-		contact2 = contacts[1];
-		contact3 = contacts[2];
-		contact4 = contacts[3];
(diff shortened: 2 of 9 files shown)


src/classes/ADDR_HouseholdAcc_TEST.cls
```diff
@@ -1851,9 +1851,16 @@ public with sharing class ADDR_HouseholdAcc_TEST {
         Test.stopTest();
         
         // verify the Address objects split street.
-        List<Address__c> listAddr = [select Id, MailingStreet__c, MailingStreet2__c, Formula_MailingStreetAddress__c from Address__c 
-        where MailingStreet__c = 'new street'];
+        List<Address__c> addresses = [SELECT Id, MailingStreet__c, MailingStreet2__c, Formula_MailingStreetAddress__c from Address__c];
```
xinnanlisf: Curious about this. Why do we need to create a new list of address? Can we loop through the addresses and update the mailing street field?

src/classes/CON_Preferred_TEST.cls
```diff
@@ -54,12 +54,19 @@ public with sharing class CON_Preferred_TEST {
 		insert contacts;
 		Test.stopTest();
 		
-		contacts = [select Email, Phone, UniversityEmail__c, WorkEmail__c, AlternateEmail__c, HomePhone, WorkPhone__c, OtherPhone, MobilePhone 
-					from Contact where ID in :contacts order by LastName];
-		contact1 = contacts[0];
-		contact2 = contacts[1];
-		contact3 = contacts[2];
-		contact4 = contacts[3];
+		contacts = [SELECT Email, LastName, Phone, UniversityEmail__c, WorkEmail__c, AlternateEmail__c, HomePhone, WorkPhone__c, OtherPhone, MobilePhone
```
xinnanlisf: This looks a bit hardcoded but I guess it was like this before. We do have a utility class to create test contacts. Not a big deal thou. Just fyi :)

src/classes/ADDR_HouseholdAcc_TEST.cls
xinnanlisf: Curious about this. Why can't we loop through addresses and update the mailing street field?

lifewithryan: Originally there was a filter on this query: "MailingStreet__c = 'new street'" however we have to remove that or Shield encryption will fail. Therefore, not knowing exactly what could be coming back from the select, I wanted to ensure that we were creating a list that exactly matches what the original query was intending which was to only contain addresses whose Mailing Street was 'new street' I am basically 'guaranteeing' that the data I'm working with is exactly the data I need.
We could probably loop through and update but I felt this more exactly duplicated the intent of using the filtered query.

lifewithryan: We could do an update here I suppose, but its more DML. This way I could just query all addresses and simply filter out the ones we care about, in this case 'new street'

I think that's what you're asking...if I'm wrong lemme know.

src/classes/CON_Preferred_TEST.cls
lifewithryan: Yes its VERY hard-coded but to me that's kind of the point. In a unit test, we need to know exactly the data we are woking with. The original issue here was that we were ordering by last name and the last name's were named such that doing an order by gave us a guarantee that element 0 was the first ordered name, etc. However, we can't use the lastname field in an order by if Shield is enabled. This means we couldn't necessarily say that the 0 element was the record we wanted. 
By creating a map by name we can then retrieve the exact records we want to compare. It feels very hard-coded but in a unit test, I would argue that it should be very brittle. Overall, I feel we should never rely on the order in which a query result is handed back to us for testing data. This was a way to ensure we knew exactly what record we are working with, without having to rely on result being returned in a specific order.

aligeez:
@xinnanlisf  & @lifewithryan  - Let me know if you want me to take a first stab at writing a 1-2 sentence blurb about this for the Release Notes.

PR #671, opened 2018-12-14
Feature/xinnanli/acc naming

xinnanlisf:
# Critical Changes

# Changes
Moving naming functions to a new class: Utility_ACCT_Naming.cls
# Issues Closed

# New Metadata
New Class: Utility_ACCT_Naming.cls
# Deleted Metadata

# Testing Notes
Regression tests for account naming

### DIFF
--- a/src/classes/UTIL_ACCT_Naming.cls
+++ b/src/classes/UTIL_ACCT_Naming.cls
@@ -1,0 +1,283 @@
+/*
+    Copyright (c) 2018, Salesforce.org
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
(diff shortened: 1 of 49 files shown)


lifewithryan:
@dashattuck I've already done a brief QA and all appears well.  Assigning this to you in case you want to take it through regression testing for account naming, otherwise, I can merge, please advise.

src/classes/UTIL_ACCT_Naming.cls
```diff
@@ -0,0 +1,285 @@
+/*
```
robertbwatson: @lifewithryan @xinnanlisf When I initially pushed the latest from master into my new scratch org, VS Code (or CCI? not sure which) automatically added this class to the package.xml file. 

Is there a reason it wasn't added to the package.xml, or was it just missed?

I can add it in my branch that I'm working in now, but I wasn't sure how we go about these minor changes here at SFDO.

PR #679, opened 2019-01-04
Add GEM QA org

jstvz:
Kevin's GEM config org was not included in the last production push for as yet unknown reason. This PR includes that org's ID in the HEDA QA push list.
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


ceiroa:
@jstvz I don't think the build will ever run because the branch doesn't start with feature/, will it? In that case we won't be able to merge.

PR #688, opened 2019-01-16
Acc naming enhancement

### LINKED ISSUE #621: Erroneous Account renaming in ACCT_IndividualAccounts_TDTM
We received a question in the HUB where a user was seeing non-Household accounts getting renamed with Household naming logic whenever a new Contact was added to that account. In this case, they had an Account with *no recordtype* that existing before HEDA was installed. With HEDA installed, when they modified or added a Contact to that Account, the household name was applied. There are a number of vulnerabilities in the ACCT_IndividualAccounts_TDTM that can lead to this behavior.

**Contact On Delete Events:**
If an existing contact linked to an existing account is Deleted, line 115 of the trigger handler is checking if the account should be renamed. That if statement is ONLY checking if there is an account Id. It is _not_ checking if the Account is a household account. So the Account would be renamed for ALL (including blank) recordtypes. 

**AccountId is Cleared on an existing Contact:**
If the Account field on an existing contact is CLEARED (not modified to another account, just removed entirely), line 174 is checking if the originally linked Account should be renamed...but it is not checking if that Account is a Household account. So the Account would be renamed for ALL (including blank) recordtypes. 

**The Household record type (HH_Account) is deleted from the org (or never installed)**
It's possible that an org decides it doesn't need the Household record type and chooses to delete it. Or they may install HEDA directly from the package url (shown for each release) rather than use the installer. Many of the checks in the trigger handler are calling Util_Describe.getHhAccRecTypeID. That method looks at the available record types available on Account and returns the one where the developer name is HH_Account. If it does not find a RecordType with that name, the method returns null. So the comparison "if (c.Account.RecordTypeId == UTIL_Describe.getHhAccRecTypeID()) " is actually matching if the Account's RecordType is null. So Accounts with blank RecortTypeIds are treated as Households...and so are renamed when a linked Contact is edited or new Contact added. This is the exact scenario encountered by our Hub question. Ideally an additional check would see if the Account's RecordTypeId isn't blank (on lines 141 and 204).

**Potential for Hierarchy_Settings.Automatic_Household_Naming__c to be null**
A checkbox field can have 3 values: true, false and null. When a new field is added to to an existing custom settings object (or any object), that field is going to be Null for any existing records (not False). So the comparison "UTIL_CustomSettingsFacade.getSettings(). Automatic_Household_Naming__c != false" would succeed even if the actual value of the field is null instead of true). A better comparison would distinctly check for "== true" instead of != false. I'm not sure if this is actually happening in an org but noticed it as an additional vulnerability. The workaround in this case would just be to edit HEDA settings, uncheck "Automatic Household Naming" and click Save (this makes sure the value stored is actually False, not Null)

xinnanlisf:
# Critical Changes

# Changes
- We are adding enhancements for account naming
  - Avoid null value comparison for custom setting: Automatic_Household_Naming__c
  - Check if record type id is null
  - When deleting a contact, automatic naming only applies on household account when it is enabled
# Issues Closed
Fixes #621
# New Metadata

# Deleted Metadata

# Testing Notes
Issue 1:
1. Go to HEDA setting page and set "Administrative" as default account model
2. Create a test contact: test contact 1
3. Create another test contact under the parent account of test contact 1 
4. Delete test contact 1 and make sure the name of parent account is not changed

Issue 2: 
1. Go to HEDA setting page and set "Administrative" as default account model
2. Create a test contact: test contact 1
3. Under record page of test contact 1, clear up account field
4. Make sure the parent account of test contact 1 is not renamed

Issue 3:
1. Create a different record type under account: test record type
2. Go to HEDA setting page and set "Household" as "Default account model"
3. Set "test record type" as the "Household Account Record Type"
4. Make sure Household account naming is still working

Issue 4: 
1. Go to Setup -> Custom Settings -> Manage Hierarchy settings
2. Clear up Automatic_Household_Naming__c
3. Household account should be deactivated right now but make sure no null pointer exception is thrown


src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -120,12 +120,25 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
...
+            List<Account> relatedAccounts = new List<Account>();
+            for (Contact c : oldContacts) {
                 listContactAccountDelete.add(c);
+                //Collect related account ids
+                relatedAccountIds.add(c.AccountId);
+            }
+
+            //Re-query to get correct Account values
+            relatedAccounts = [SELECT Id,
+                                    RecordType.Id
```
lifewithryan: I think we can do SELEC Id, RecordTypeId FROM Account... here rather than RecordType.Id -- I believe the standard objects get a "RecordTypeId" field once we introduce recordtype on the object. Its one less table to traverse on the backend, probably not a big deal though. It would then also change "account.RecordType.Id to account.RecordTypeId in the logic. Just a thought.

That said perhaps you're aware of issues that I am unaware of. Both ways work so if you want to leave it, that is fine. It looks like line 165 uses that field even though it's coming from the contact "contact.Account.RecordTypeId"

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -149,8 +162,9 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
                         if(c.Account.Primary_Contact__c == null) {
                             mapAccountIdContactId.put(c.AccountId, c.Id);
                         }
-                        if (c.Account.RecordTypeId == userDefinedHHRecordTypeId
-                                && UTIL_CustomSettingsFacade.getSettings().Automatic_Household_Naming__c != false) {
+                        if (c.Account.RecordTypeId != null
```
lifewithryan: I'd be curious to see if we do this check anywhere else besides just these two places and wonder if it's worth breaking that check out into a method. Its probably fine as is but if we keep adding to this class which is already pretty big, we may want to consider it in the future. (referring to this if check on lines 165-167)

src/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -213,8 +227,9 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
                                 || c.FirstName != oldContacts[i].FirstName
                                 || c.Exclude_from_Household_Name__c != oldContacts[i].Exclude_from_Household_Name__c)
                             && c.AccountId != null
+                            && c.Account.RecordTypeId != null
                             && c.Account.RecordTypeId == userDefinedHHRecordTypeId
-                            && UTIL_CustomSettingsFacade.getSettings().Automatic_Household_Naming__c != false) {
+                            && UTIL_CustomSettingsFacade.getSettings().Automatic_Household_Naming__c == true) {
```
lifewithryan: Yes I like this change :)

src/classes/ACCT_IndividualAccounts_TDTM.cls
xinnanlisf: @lifewithryan those two ways should return the same value but I like the idea of using RecordTypeId!

src/classes/ACCT_IndividualAccounts_TDTM.cls
xinnanlisf: That's a great idea! I will create boolean methods for both household and admin.

xinnanlisf:
@ryanguest should be ready now!

lifewithryan:
@dashattuck we should probably run this through the account naming regressions at least.

xinnanlisf:
@dashattuck @lifewithryan I have updated the testing scenarios here. Hopefully it helps. Let me know if any questions.

dashattuck:
@xinnanlisf - I have to assume, until I hear back from you, whether 'Set "test record type" as the "Household Account Record Type"' means that we need to go into HEDA Settings, and choose 'test record type' as the default Household Account .

dashattuck:
Verified on both namespaced and not namespaced. 
All issues #1 through #4 passed.

PR #694, opened 2019-01-19
Duplicate Department Affiliations are no longer inserted

### LINKED ISSUE #651: Faculty contact getting duplicate Department affiliation records when inserting multiple Course Connections
This Power of Us Hub post started an investigation of Course Offerings (with Primary Faculty populated) creating multiple affiliation records linking a faculty contact to the same department.
https://powerofus.force.com/_ui/core/chatter/groups/GroupProfilePage?g=0F980000000D6uj&fId=0D58000004Xrgok

If you insert a Course Offering with primary faculty (or a Faculty course connection), HEDA will check their existing affiliation records to see if that contact is already linked to the course's department. If they are not, HEDA will create the affiliation. That works fine for single-record transactions. 

However, if you insert multiple course offerings that relate that faculty member to the department (e.g., two course offerings in the same department) within the same transaction (data loader, integration, etc), HEDA is adding the new department affiliation to the "list to be inserted" without checking if there's already another affiliation linking the same affiliation to the same contact in that list. That causes the affiliation to get created multiple times.

The affected code is found in CCON_Faculty_TDTM when affiliationsToInsert list gets populated (before insert section).

robertbwatson:
# Critical Changes

# Changes
- If within a single transaction (via Data Loader, for example) you insert multiple Course Connection or Course Offering records that relate to the same faculty Contact and department, duplicate Affiliation records are no longer created.

# Issues Closed
Fixes #651 

# New Metadata

# Deleted Metadata

# Testing Notes

1. Ensure that Course Connections is enabled in HEDA Settings.
2. Create two Course Connections at the same time (through data loader, etc) for the same Faculty & Department - this should not create duplicate affiliations.
3. Create or update two Course Offerings at the same time (through data loader, etc) for the same Faculty & Department - this should not create duplicate affiliations.
4. Disable Course Connections and repeat step 3.


src/classes/CCON_Faculty_TDTM.cls
```diff
@@ -53,9 +53,8 @@ public with sharing class CCON_Faculty_TDTM extends TDTM_Runnable {
         Id facultyRecType = (Id)UTIL_Describe.getFacultyConnectionRecType();
 
         List<Course_Offering__c> courseOfferingsToUpdate = new List<Course_Offering__c>();
-        List<Course_Enrollment__c> courseConnsBeforeInsert = new List<Course_Enrollment__c>();
```
robertbwatson: FYI, I removed this because it wasn't used anywhere.

src/classes/CCON_Faculty_TEST.cls
```diff
@@ -439,8 +483,8 @@ public with sharing class CCON_Faculty_TEST {
     }
 
     /*********************************************************************************************************
-    * @description Verifies that when a Course Connection is removed the related Affiliation is also removed.
```
robertbwatson: This was just a bad copy/pasta so I fixed it.

robertbwatson:
@lifewithryan I added you as a requested reviewer on this. Should I also request a review from Liane to verify the release notes?

src/classes/CCON_Faculty_TDTM.cls
lifewithryan: i love deleting code :)

src/classes/CCON_Faculty_TDTM.cls
```diff
@@ -96,29 +95,43 @@ public with sharing class CCON_Faculty_TDTM extends TDTM_Runnable {
...
+                        // Skip over the Course Connection if it is not for a Faculty.
+                        if(enroll.get('RecordTypeId') != facultyRecType) {
+                            continue;
+                        }
+
+                        enroll.Account__c = offerings.get(enroll.Course_Offering__c).Course__r.Account__c;
+
+                        if (contactIdAccountIdAfflList.containsKey(enroll.Contact__c) && (
+                                contactIdAccountIdAfflList.get(enroll.Contact__c).containsKey(enroll.Account__c))) {
+                            enroll.Affiliation__c = contactIdAccountIdAfflList.get(enroll.Contact__c).get(enroll.Account__c)[0].Id;
```
lifewithryan: I'm normally not a huge fan of chaining so deep. It's rarely ever an issue but easy to lost track of (in my mind) when scanning through code if one isn't super familiar with it.  I feel like we could assign: contactIdAccountIdAfflList.get(enroll.Contact__c) to a variable here to reduce that affect, especially if we are going to use it more than once. Its not a huge deal however.

src/classes/CCON_Faculty_TDTM.cls
```diff
@@ -96,29 +95,43 @@ public with sharing class CCON_Faculty_TDTM extends TDTM_Runnable {
...
+                        } else {
+                            connsNeedingAffls.add(enroll);
+                            
+                            // Is the Affiliation already being inserted? If so, do not insert a duplicate.
+                            String contactAccountKey = enroll.Contact__c + '' + enroll.Account__c;
+                            if (affiliationsToInsertByContactAccountString.containsKey(contactAccountKey)) {
```
lifewithryan: a big part of me wishes this didn't work so damn well because I hate reading it, but I'll be damned if its not much more efficient :). Actually we should consider doing this more, its something I rarely think to do so kudos to you.

src/classes/CCON_Faculty_TEST.cls
```diff
@@ -372,7 +372,7 @@ public with sharing class CCON_Faculty_TEST {
 
             insert conn;
 
-            // Check that an Affilition was created.
+            // Check that an Affiliation was created.
```
lifewithryan: thanks for the cleanup here (and above)

src/classes/CCON_Faculty_TEST.cls
lifewithryan: thanks

src/classes/CCON_Faculty_TDTM.cls
robertbwatson: Great suggestion - I'll commit up a quick change momentarily.

src/classes/CCON_Faculty_TDTM.cls
robertbwatson: Yeah I was taught to exit early when possible (or continue in this case)... it usually results in less nested if statements.

````
if (true condition) {
    // logic here, with its own potential nesting
}

// vs

if (false condition) {
    exit
}
// logic here instead
````

**Comment by lifewithryan** [comment]
> @lifewithryan I added you as a requested reviewer on this. Should I also request a review from Liane to verify the release notes?

Let me verify with Liane how she wants to be notified. We haven't had dedicated docs since I started on HEDA so don't actually know how to answer that :)

**Comment by lifewithryan** [comment]
> > @lifewithryan I added you as a requested reviewer on this. Should I also request a review from Liane to verify the release notes?
> 
> Let me verify with Liane how she wants to be notified. We haven't had dedicated docs since I started on HEDA so don't actually know how to answer that :)

@robertbwatson this was Liane's response: When the dev opens the PR for code review, it's ideal to add the Doc person as a reviewer at that time. That lets Doc look at any UI text changes and also review/edit the release note.

**Comment by robertbwatson** [comment]
> > > @lifewithryan I added you as a requested reviewer on this. Should I also request a review from Liane to verify the release notes?
> > 
> > 
> > Let me verify with Liane how she wants to be notified. We haven't had dedicated docs since I started on HEDA so don't actually know how to answer that :)
> 
> @robertbwatson this was Liane's response: When the dev opens the PR for code review, it's ideal to add the Doc person as a reviewer at that time. That lets Doc look at any UI text changes and also review/edit the release note.

@lifewithryan Thanks, I'll add her now. I made the code review update you suggested, btw :)

File: src/classes/CCON_Faculty_TDTM.cls (hunk 4 above)
Comment (lifewithryan) [reply src/classes/CCON_Faculty_TDTM.cls]: Yea, I'm always thrown by the continue/break stuff. For me perhaps in reminds of a GOTO statement and feels like I'm forced to jump about in the code. That said, I should probably get used to it because I feel its efficient. I just hate it :)

**Comment by lifewithryan** [comment]
as an FYI/Reminder to myself -- this is back to dev in progress. The approval by Liane is confusing me and making me thing this is ready, but @robertbwatson is still working in this issue.

File: src/classes/CCON_Faculty_TDTM.cls [hunk 6]
```diff
@@ -96,29 +95,45 @@ public with sharing class CCON_Faculty_TDTM extends TDTM_Runnable {
...
+                                continue;
                             }
+
+                            Affiliation__c affl = new Affiliation__c(
```
robertbwatson: Note that here we don't set the Start & End Date of the Affiliation, but we do in the COFF_Affiliation_TDTM class when Course Connections is disabled. I noted in the WI that we should consider making this consistent (but out of scope of this bug fix)

src/classes/COFF_Affiliation_TEST.cls
```diff
@@ -121,43 +121,86 @@ public with sharing class COFF_Affiliation_TEST {
         System.assertEquals(1, affls.size());
     }
 
+    /*********************************************************************************************************
```
robertbwatson: I'm sorry the diff here is so bad. I believe I added one unit test, and decided to just update two other unit tests to test with more than one record being inserted/updated at one time. I originally wanted to create new unit tests for the bulk insert/update scenarios, but there are already 24 unit tests in this class...

robertbwatson:
> as an FYI/Reminder to myself -- this is back to dev in progress. The approval by Liane is confusing me and making me thing this is ready, but @robertbwatson is still working in this issue.

@lifewithryan This is now ready for your review again.

src/classes/COFF_Affiliation_TEST.cls
lifewithryan: Yea, I'm with you on wanting to do that, but I do know that historically they have a concern with numerous individual tests...it bugs me, but I get it :/

src/classes/COFF_Affiliation_TDTM.cls
```diff
@@ -268,7 +276,8 @@ public class COFF_Affiliation_TDTM extends TDTM_Runnable {
     * @description Sets each Affiliation pointing to the new Faculty member that is teaching the Course Offering.
     ********************************************************************************************************/
     private List<SObject> getAfflsToUpdate(List<ID> outdatedFacultyIDs, List<ID> updatedFacultyIDs, Map<Id, Course_Offering__c> mapNewFacultyIdToNewCourseOffering) {
-        List<Affiliation__c> afflsWithContact = [select Contact__c from Affiliation__c where Contact__r.ID in :outdatedFacultyIDs and Role__c = 'Faculty'];
+        List<Affiliation__c> afflsWithContact = [SELECT Contact__c FROM Affiliation__c WHERE Contact__c IN :outdatedFacultyIDs AND Role__c = 'Faculty'];
```
lifewithryan: It looks like you're simply fixing what was already there so this is not you, BUT -- this query will break if a customer decides to encrypt the Role__c field I believe. Since encrypted fields can't be used in SOQL filters and ORDER BY statements. What really sucks about that is, it would force us to query on all of those whose Contact Id is in that list, and then loop over to filter out non Faculty values.

It's the same in the unit tests below. We cannot have queries there with the strings if those fields get encrypted. It's an extra step for us, but we just announced a release or two ago that we were now Shield compliant.

However, since it looks like you were simply fixing something we had there already (3 years ago it appears) and our scans didn't seem to find it when we were encrypting fields, I think we can leave it for now. I might just test encrypting that field when I pull this down to see if it freaks out. (Maybe its find if its a picklist value?)

PR #700, opened 2019-01-29
CON_Preferred_TDTM fix for batch sizes over 200 records

### LINKED ISSUE #639: Recursion Flag stopping CON_Preferred_TDTM from working for batch sizes over 200
CON_Preferred_TDTM is one of the classes that was updated to address some of the static flag issues we saw with large batch sizes. However, the current code is still falling victim to the DML behavior where larger sets of records inserted/updated via DML are getting broken into 200 record chunks behind the scenes without resetting static variables. So CON_Preferred_TDTM will run for the first 200 records and then thinks it shouldn't run for the next set of 200.

Since this class runs before insert/update, the recursion check probably isn't necessary since it's just setting field values (email and phone) on the contact and not inserting/updating any other related records.

The current unit tests in TDTM_DMLgt200_TEST do not test the preferred email/phone functionality (i.e., do not assert the email and phone fields are getting set correctly on all records). I verified the bug with my own unit test:

`
@isTest
    static void insertUpdateContact() {
        List<Contact> testContacts = new List<Contact>();
        
        for (Integer i = 0; i < 300; i++) {
            Contact con = new Contact(
                FirstName='Test',
                LastName='Contact_forTests ' + i,
                hed__WorkEmail__c = 'junk@test.net',
                hed__Preferred_Email__c = 'Work',
                hed__WorkPhone__c = '206-777-8888',
                hed__PreferredPhone__c = 'Work');
            testContacts.add(con);
        }
        insert testContacts;
        
        List<Contact> insertedContacts = [Select FirstName, LastName, hed__WorkPhone__c, Phone, hed__WorkEmail__c, Email
                                          from Contact];
        
        for (Contact con : insertedContacts) {
            System.assertEquals(con.hed__WorkEmail__c, con.Email);
        }
        
    }`

Currently the workaround is to set batch size to 200 (not idea for large Bulk API updates) or manually set the Email and Phone values in the Bulk API import.

robertbwatson:
# Critical Changes

# Changes
- When inserting or updating more than 200 Contact records at a time, _all_ records now have their standard Email and Phone fields populated based on their Preferred Email and Preferred Phone values. Previously, only the first 200 records were populated.

# Issues Closed
- Fixes #639

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/CON_Email.cls
+++ b/src/classes/CON_Email.cls
@@ -192,15 +192,17 @@ public class CON_Email {
                 }
             }
 
-            for(String fieldName : contactEmailFieldMap.keySet() ) {
+            for(String fieldName : contactEmailFieldMap.keySet()) {
+                String fieldLabel = contactEmailFieldMap.get(fieldName).getLabel();
+                
                 //If the field is not a standard email field and it is not duplicate with HEDA email, then add it to the list
-                if ( fieldName != emailLabel && !(hedaEmailField.get(contactEmailFieldMap.get(fieldName).getLabel()) != null &&
-                                fieldName != hedaEmailField.get(contactEmailFieldMap.get(fieldName).getLabel()))) {
+                if (fieldName != emailLabel && 
+                        !(hedaEmailField.get(fieldLabel) != null && fieldName != hedaEmailField.get(fieldLabel))) {
                     String emailField = (String)contact.get(fieldName);
 
-                    allFields.add( new CON_EmailField( emailField, contactEmailFieldMap.get(fieldName).getLabel(), fieldName ) );
+                    allFields.add(new CON_EmailField(emailField, fieldLabel, fieldName));
                     if(String.isNotBlank(emailField)) {
-                        valuedFields.add( new CON_EmailField( emailField, contactEmailFieldMap.get(fieldName).getLabel(), fieldName ) );
+                        valuedFields.add(new CON_EmailField(emailField, fieldLabel, fieldName));
                     }
                 }
             }
--- a/src/classes/CON_Email.cls
+++ b/src/classes/CON_Email.cls
@@ -162,6 +162,8 @@ public class CON_Email {
         }
     }
 
+    private static Map<String, Schema.DescribeFieldResult> contactEmailFieldMap;
+
     /*******************************************************************************************************
     * @description An inner wrapper class to contain a list of email fields and perform functions on them
     * @param Contact the contact to build list from.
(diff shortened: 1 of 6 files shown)


src/classes/TDTM_DMLgt200_TEST.cls
```diff
@@ -69,39 +69,32 @@ private class TDTM_DMLgt200_TEST {
...
 
-        //one contact should have been created
+        //many contacts should have been created
         system.assertEquals(inputSize, insertedContacts.size());
 
         //relationship should be bi-directional
         for (Contact con : insertedContacts) {
-	        system.assertEquals(con.id, con.Account.Primary_Contact__c);   
+	        system.assertEquals(con.id, con.Account.Primary_Contact__c);
+            system.assertEquals(con.WorkEmail__c, con.Email);
```
robertbwatson: These two lines are the two new pieces of code that really test that the bug is fixed for > 200 records. When inserting > 200 records, these assertions were false after record 200.

robertbwatson:
Not quite ready for review - there's a few unit test failures related to the CPU timeout limit so I want to try to do some benchmarking to see if maybe the CON_Preferred_TDTM class & methods it calls needs to be updated to run more efficiently in bulk.

xinnanlisf:
@robertbwatson sometimes it happens..it should pass if you rerun the build.

src/classes/CON_Email.cls
```diff
@@ -176,26 +178,29 @@ public class CON_Email {
 
             this.allFields = new List<CON_EmailField>();
             this.valuedFields = new List<CON_EmailField>();
-            Map<String, Schema.DescribeFieldResult> fieldMap = UTIL_Describe.getFieldsOfType('Contact', 'EMAIL');
```
robertbwatson: This seems like a safe enough change but let me know if its preferred to open a new PR/WI for this. When testing the bulk preferred change, there were a few intermittent unit test failures and I noticed this was inefficient.

Before the cache fix:
```
Class.hed.CON_Preferred_TDTM.run: line 56, column 1: public static void processPreferredEmail(Contact, List<Contact>): executed 1500 times in 6557 ms
Class.hed.CON_Email.processPreferredEmail: line 55, column 1: public hed.CON_Email.CON_EmailFieldList(Contact): executed 1500 times in 5714 ms
```

After the cache fix:
```
Class.hed.CON_Preferred_TDTM.run: line 56, column 1: public static void processPreferredEmail(Contact, List<Contact>): executed 1500 times in 1854 ms
Class.hed.CON_Email.processPreferredEmail: line 55, column 1: public hed.CON_Email.CON_EmailFieldList(Contact): executed 1500 times in 1139 ms
```

src/classes/CON_Email.cls
```diff
@@ -176,26 +178,29 @@ public class CON_Email {
...
-            this.emailLabel = fieldMap.get('Email').getLabel();
+
+            if (contactEmailFieldMap == null) {
+                contactEmailFieldMap = UTIL_Describe.getFieldsOfType('Contact', 'EMAIL');
```
lifewithryan: I think we are in the clear here but this may be the kind of thing that clues us into notifying QA to test in both namespaced and non-namespaced environments. Since this is a standard object and field, we're good, just want to point it out, more as a reference/reminder to myself.

src/classes/CON_Email.cls
```diff
@@ -176,26 +178,29 @@ public class CON_Email {
...
+                    hedaEmailField.put(contactEmailFieldMap.get(fieldName).getLabel(), fieldName);
                 }
             }
 
-            for(String fieldName : fieldMap.keySet() ) {
+            for(String fieldName : contactEmailFieldMap.keySet() ) {
                 //If the field is not a standard email field and it is not duplicate with HEDA email, then add it to the list
-                if ( fieldName != emailLabel && !(hedaEmailField.get(fieldMap.get(fieldName).getLabel()) != null &&
-                                fieldName != hedaEmailField.get(fieldMap.get(fieldName).getLabel()))) {
+                if ( fieldName != emailLabel && !(hedaEmailField.get(contactEmailFieldMap.get(fieldName).getLabel()) != null &&
```
lifewithryan: since these checks are longer, would you mind putting one condition per line:
`if(check 1 &&
     check2 && 
     check3) `
or something similar. Also there are a ton of parenthesis here maybe breaking out some of these checks into variables? I'm less concerned about this than I am the breaking up of the conditions though.

src/classes/CON_Email.cls
```diff
@@ -176,26 +178,29 @@ public class CON_Email {
...
+                                fieldName != hedaEmailField.get(contactEmailFieldMap.get(fieldName).getLabel()))) {
                     String emailField = (String)contact.get(fieldName);
 
-                    allFields.add( new CON_EmailField( emailField, fieldMap.get(fieldName).getLabel(), fieldName ) );
+                    allFields.add( new CON_EmailField( emailField, contactEmailFieldMap.get(fieldName).getLabel(), fieldName ) );
```
lifewithryan: if you do what I mentioned above by putting things like contactEmailFieldMap.get(fieldName).getLabel() into a variable then you can re-use it throughout this method. I feel it makes it more readable. if you're uber concerned about performance in so doing however, let me know and we can chat.

src/classes/CON_Email.cls
lifewithryan: I see my carriage returns didn't take above...those checks should have been like:

check1 &&
check2 &&
check3

src/classes/CON_Email.cls
ceiroa: I think it makes sense to have it as part of the same PR.

src/classes/CON_Email.cls
robertbwatson: Definitely not concerned about performance - I think storing it in a variable is more efficient.

src/classes/CON_Email.cls
robertbwatson: @lifewithryan Thanks - I've made the requested changes.

robertbwatson:
@lianeyee Can you do a review of the release notes here sometime next week? Thanks :)

PR #701, opened 2019-01-29
Avoid creating duplicate custom settings

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
1. Go to HEDA settings page
2. Go to Setup -> Custom settings -> Manage "Hierarchy Settings"
3. Make sure no "User User" custom setting is created

### DIFF
--- a/src/classes/STG_InstallScript.cls
+++ b/src/classes/STG_InstallScript.cls
@@ -65,6 +65,8 @@ global without sharing class STG_InstallScript implements InstallHandler {
             Version previousVersion = context.previousVersion();
 
             disablePreferredEmailEnforcement(previousVersion);
+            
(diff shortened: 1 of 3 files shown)


src/classes/STG_InstallScript.cls
```diff
@@ -65,6 +65,8 @@ global without sharing class STG_InstallScript implements InstallHandler {
             Version previousVersion = context.previousVersion();
 
             disablePreferredEmailEnforcement(previousVersion);
+            
+            populdateAdminRecordTypeSetting(previousVersion);
```
lifewithryan: super simple but "populate" has been mispelled here. We could move forward but its a 5 second change so I figure we might as well fix it.

PR #705, opened 2019-02-04
Label rename of Advancement Namespace to GEM Namespace

jbrandolinisf:
# Critical Changes

# Changes
Renamed Advancement Namespace label of field Hierarchy_Settings__c.Advancement_Settings__c to GEM Namespace

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

Confirm label is updated in Hierarchy Settings

Relates to Work Item: [W-030968](https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000007fR4AQAU/view)

### DIFF
--- a/src/objects/Hierarchy_Settings__c.object
+++ b/src/objects/Hierarchy_Settings__c.object
@@ -75,7 +75,7 @@
         <description>The namespace of the Gift Processing package.</description>
         <externalId>false</externalId>
         <inlineHelpText>The namespace of the Gift Processing package.</inlineHelpText>
-        <label>Advancement Namespace</label>
+        <label>GEM Namespace</label>
         <length>10</length>
         <required>false</required>
         <trackTrending>false</trackTrending>
--- a/src/objects/Hierarchy_Settings__c.object
+++ b/src/objects/Hierarchy_Settings__c.object
@@ -72,9 +72,9 @@
     <fields>
         <fullName>Advancement_Namespace__c</fullName>
         <defaultValue>&quot;adv1&quot;</defaultValue>
-        <description>The namespace of the Gift Processing package.</description>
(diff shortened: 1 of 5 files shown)


src/objects/Hierarchy_Settings__c.object
```diff
@@ -75,7 +75,7 @@
         <description>The namespace of the Gift Processing package.</description>
```
lparrott: Minor change, but should we also update the description to say "... Gift Entry Management package"? 

I do actually notice some references to "Gift Processing" in the Docs, so maybe this isn't a big deal (actually, I KNOW it isn't a big deal...)

jbrandolinisf: Do we want to go with Gift Entry Management package, or Gift Entry Management application?  I don't mind changing it, either way.

lparrott: It's a mouthful, but I'd vote for "Gift Entry Management Managed Package", or "GEM Managed Package" because the value of this setting has to match the Namespace of the GEM package that is installed in the org.

annejackson:
@lianeyee and @juliesnow  I think you are going to check out release notes...please make sure this changes section is okay.

lianeyee:
@annejackson Since we don't have any customers using the package yet, I don't think we need a release note. Does that sound reasonable?

annejackson:
I think that works! Just wanted to make sure it wouldn't show up in HEDA release notes since it's in the Heda repo. So you are just thinking move the changes section notes up above critical changes @lianeyee?

lianeyee:
@annejackson I just deleted the item under Changes. That'll keep it out of the release notes. Thanks for checking!

PR #712, opened 2019-02-12
Feature/rheadley/prenaffl label updates

lifewithryan:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/objects/Hierarchy_Settings__c.object
+++ b/src/objects/Hierarchy_Settings__c.object
@@ -95,29 +95,29 @@
     <fields>
         <fullName>Affl_ProgEnroll_Copy_End_Date__c</fullName>
         <defaultValue>false</defaultValue>
-        <description>Determines if the Affiliation automatically created by creating a Program Enrollment will inherit the Program Enrollment&apos;s Start Date.</description>
+        <description>Indicates whether to set the End Date of the Affiliation created with a Program Enrollment. If enabled, set the End Date to the Program Enrollment&apos;s End Date. If disabled, don't specify the Affiliation&apos;s End Date.</description>
         <externalId>false</externalId>
-        <inlineHelpText>Determines if the Affiliation automatically created by creating a Program Enrollment will inherit the Program Enrollment&apos;s Start Date.</inlineHelpText>
-        <label>Affl End Date on Prog Enroll Creation</label>
+        <inlineHelpText>Indicates whether to set the End Date of the Affiliation created with a Program Enrollment. If enabled, set the End Date to the Program Enrollment&apos;s End Date. If disabled, don't specify the Affiliation&apos;s End Date.</inlineHelpText>
+        <label>Affil End Date Copy from Prog Enroll</label>
         <trackTrending>false</trackTrending>
         <type>Checkbox</type>
     </fields>
     <fields>
         <fullName>Affl_ProgEnroll_Copy_Start_Date__c</fullName>
         <defaultValue>false</defaultValue>
-        <description>Determines if the Affiliation automatically created by creating a Program Enrollment will inherit the Program Enrollment&apos;s Start Date.</description>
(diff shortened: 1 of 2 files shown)


robertbwatson:
I added @lianeyee to this so that she can review the release notes.

src/objects/Hierarchy_Settings__c.object
```diff
@@ -95,29 +95,29 @@
     <fields>
         <fullName>Affl_ProgEnroll_Copy_End_Date__c</fullName>
         <defaultValue>false</defaultValue>
-        <description>Determines if the Affiliation automatically created by creating a Program Enrollment will inherit the Program Enrollment&apos;s Start Date.</description>
+        <description>Indicates whether to set the End Date of the Affiliation created with a Program Enrollment. If enabled, set the End Date to the Program Enrollment&apos;s End Date. If disabled, don't specify the Affiliation&apos;s End Date.</description>
```
robertbwatson: @lifewithryan I don't know how much it matters, but some of the apostrophes here in this metadata are represented as "`&apos;`" instead of '

lifewithryan: I went with what I was seeing in the line before replacement. It looked like they were being HTML encoded to ensure consistency (that was my guess) so I just went with it. After spinning them up on an org, I think we are good. HOWEVER -- a very good point here is if we start automatically translating things, I wonder if it will have an effect...

lianeyee:
@robertbwatson This PR actually doesn't need a release note. It's cleanup work for a feature that will be announced as entirely new in 1.70, so we'll have just a single large release note. In an hour, I'll do the review of the labels in the PR.

src/objects/Hierarchy_Settings__c.object
```diff
@@ -95,29 +95,29 @@
...
-        <label>Affl Start Date on Prog Enroll Creation</label>
+        <inlineHelpText>Indicates whether to set the Start Date of the Affiliation created with a Program Enrollment. If enabled, set the Start Date to the Program Enrollment&apos;s Start Date. If disabled, don't specify the Affiliation&apos;s Start Date.</inlineHelpText>
+        <label>Affil Start Date Copy from Prog Enroll</label>
         <trackTrending>false</trackTrending>
         <type>Checkbox</type>
     </fields>
     <fields>
         <fullName>Affl_ProgEnroll_Del_Status__c</fullName>
-        <description>Status to change an Affilation to when the related Program Enrollment is deleted, if the option selected is to change its status instead of deleting it.</description>
+        <description>If the related setting, Affil Delete With Prog Enroll, is disabled, this setting specifies the value to set for the Affiliation's Status, when its related Program Enrollment is deleted.</description>
```
lianeyee: We changed the label of the related setting and forgot (as in, I forgot) to update it here in this description: "If the related setting, Affil Delete When Delete Prog Enroll,"

src/objects/Hierarchy_Settings__c.object
```diff
@@ -95,29 +95,29 @@
...
         <externalId>false</externalId>
-        <inlineHelpText>Status to change an Affilation to when the related Program Enrollment is deleted, if the option selected is to change its status instead of deleting it.</inlineHelpText>
-        <label>Affl status on Prog Enrollment deletion</label>
+        <inlineHelpText>If the related setting, Affil Delete With Prog Enroll, is disabled, this setting specifies the value to set for the Affiliation's Status, when its related Program Enrollment is deleted.</inlineHelpText>
```
lianeyee: Same change needed here: ""If the related setting, Affil Delete When Delete Prog Enroll,"

robertbwatson:
> @robertbwatson This PR actually doesn't need a release note. It's cleanup work for a feature that will be announced as entirely new in 1.70, so we'll have just a single large release note. In an hour, I'll do the review of the labels in the PR.

Sounds good! Should the Changes bullet above be deleted then?

lifewithryan:
@lianeyee I've made the changes to the label and inline help text (lines 117 and 119) per your request.
cc: @robertbwatson

dgamboni:
I also took a gander and it looks good. Waiting for tests to pass.

PR #714, opened 2019-02-15
Do not update the Affiliation status to Former when there other Program Enrollment records attached

robertbwatson:
# Critical Changes

# Changes
* When deleting a Program Enrollment, if its related Affiliation's Status is updated (instead of the Affiliation being deleted), the Status is now updated only if the Affiliation isn't related to any other existing Program Enrollment records. In other words, if "Delete Related Affiliation When Deleting Program Enrollment" is disabled in HEDA Settings | Affiliations | Settings, the value in "Status Specified for Affiliation Not Deleted" does not apply unless the related Affiliation has no other Program Enrollments related to it.

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
1. Ensure that the "Delete Related Affiliation When Deleting Program Enrollment" setting is false.
2. Create a Program Enrollment for a Contact.
3. Go to the Program Enrollment Affiliation record that was automatically created.
4. On the Program Enrollment related list, create another Program Enrollment (into the same Program) for the Affiliation.
5. Delete the Program Enrollment record. The Program Enrollment record will be deleted, but the Affiliation record (& the other Program Enrollment record) will still exist. The Affiliation's Status will NOT be updated to the "Status of Affiliation Related to Deleted Program Enrollment" setting.
6. Delete the remaining Program Enrollment record. The Affiliation's Status will be updated to the "Status of Affiliation Related to Deleted Program Enrollment" setting.

### DIFF
--- a/src/classes/PREN_Affiliation_TEST.cls
+++ b/src/classes/PREN_Affiliation_TEST.cls
@@ -61,12 +61,12 @@ public with sharing class PREN_Affiliation_TEST {
         Test.stopTest();
         
         //An Affiliation should have been automatically created fron the Program Enrollment
-        List<Affiliation__c> affls = [SELECT Contact__c, Account__c FROM Affiliation__c WHERE Account__r.ID = :acc.ID];
+        List<Affiliation__c> affls = [SELECT Contact__c, Account__c FROM Affiliation__c WHERE Account__c = :acc.ID];
         System.assertEquals(1, affls.size());
         
         //The Program Enrollment record should be related to the Affiliation just created
-        enrollment = [SELECT Affiliation__r.ID FROM Program_Enrollment__c WHERE ID = :enrollment.ID];
-        System.assertEquals(affls[0].ID, enrollment.Affiliation__r.ID);
+        enrollment = [SELECT Affiliation__c FROM Program_Enrollment__c WHERE ID = :enrollment.ID];
+        System.assertEquals(affls[0].ID, enrollment.Affiliation__c);
     }
     
     /*********************************************************************************************************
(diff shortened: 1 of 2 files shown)


src/classes/PREN_Affiliation_TEST.cls
```diff
@@ -61,12 +61,12 @@ public with sharing class PREN_Affiliation_TEST {
         Test.stopTest();
         
         //An Affiliation should have been automatically created fron the Program Enrollment
-        List<Affiliation__c> affls = [SELECT Contact__c, Account__c FROM Affiliation__c WHERE Account__r.ID = :acc.ID];
+        List<Affiliation__c> affls = [SELECT Contact__c, Account__c FROM Affiliation__c WHERE Account__c = :acc.ID];
```
robertbwatson: Even though __r.ID is technically correct, for some reason it just really bothers me so I changed it 😬

lifewithryan: I'm fine with that and prefer it.

ceiroa:
No issue listed under "Issues Closed"?

robertbwatson:
> No issue listed under "Issues Closed"?

@ceiroa, We discussed in the work item that @aligeez was going to create a new GH issue after confirming with Jace. I don't see it yet so I've pinged him to follow up.

robertbwatson:
> I added some whitespace around some if's and for's -- just because I feel it's easier to read and calls attention to the decision tree, etc.
> 
> Other than that, it works as expected if I follow the testing criteria. I did notice an issue if I add a PE from the Contact's related list versus the Affiliations, but I believe that is a separate issue/work item.

Thanks - I usually prefer a bit of whitespace too :)

aligeez:
> > No issue listed under "Issues Closed"?
> 
> @ceiroa, We discussed in the work item that @aligeez was going to create a new GH issue after confirming with Jace. I don't see it yet so I've pinged him to follow up.

https://github.com/SalesforceFoundation/HEDAP/issues/730

Issue #720, opened 2019-02-26
Facility Naming: Include a "Display Name" formula field on the Facility that includes the Account name or Parent Facility name


robertbwatson:
**Note: FOR INTERNAL USE ONLY. For support questions, please use the HEDA Chatter group in [The Power of Us Hub](https://powerofus.force.com).
For feature requests, please use the Ideas tab in [The Power of Us Hub](https://powerofus.force.com).**

* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
After #716 is delivered, a user must manually name a Facility record. There could be many "Room 201" records.

* **What is the expected behavior?**
By adding a new "Display Name" unmanaged formula field on to the Facility object, the Facility record can have a name displayed that incorporates the Account Name or Parent Facility Name. (For example, "Thomas Edison Building - Room 201".)

The user will be able to update this formula field if they want to choose a different format to use.

* **What is the motivation / use case for changing the behavior?**
Having 10 "Room 201" records with no friendly name will make it difficult for a user to know which record to choose.

* **Acceptance Criteria**

  - **Given**: I create a new Facility record with the name "Room 201" and associate it to the "Thomas Edison Building" Parent Facility
  - **When**: I save the record
  -  **Then**: The "Display Name" field on the record will display "Thomas Edison Building - Room 201"

* **Acceptance Criteria**

  - **Given**: I create a new Facility record with the name "Room 201" and associate it to the "College of Business" Account
  - **When**: I save the record
  -  **Then**: The "Display Name" field on the record will display "College of Business - Room 201"

* **Acceptance Criteria**

  - **Given**: I create a new Facility record with the name "Room 201" and associate it to the "Thomas Edison Building" Parent Facility and "College of Business" Account
  - **When**: I save the record
  -  **Then**: The "Display Name" field on the record will display "Thomas Edison Building - Room 201" (the Parent Facility is used when both Account and Parent Facility are specified)

robertbwatson:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008Ll4hQAC" target="_blank">W-031480</a>

mrbelvedere:
Included in beta release 1.73 (Beta 6)

mrbelvedere:
Included in production release 1.73

PR #725, opened 2019-03-01
Creation of new Facility object & fields to support Facilities in HEDA

### LINKED ISSUE #716: Creation of new Facility object & fields to support Facilities in HEDA

Feature

The current functionality does not exist in HEDA.

This issue will result in new objects & fields and not any new logic/functionality. The user will interact with this object the same way they interact with other objects in Salesforce.

A user can navigate to a Course Offering record and see where the Couse Offering is being held. A user can manage various facilities in the educational institution.

Short of providing full-on Facility management, this allows an educational organization to house information with respect to where classes are being held and other school property assets. This could be a classroom location, a campus location, location of a lecture hall, location of a laboratory, or simply locations for campus buildings that don't necessarily pertain to student activity, e.g. Maintenance/IT Facilities. 

1.) A Facility will look up to Account, e.g. "Department of Computer Science" account could have 3 "Locations" that are Building A, Building B, Building C, and Building A has child "Locations" that are Lecture Hall 121, Lecture Hall 131" - ex: Building A in this story would be something like "The Walter Smith Sciences Building" 

2.) Course Offerings would look up to Facility. For example, Course Offering "Biology 101 - Spring 2019 - Taught by Professor Smith" would be offered in the location of Lecture Hall 121 > Building A

robertbwatson:
# Critical Changes

# Changes
* We've created a new Facility object and relevant fields that you can use to capture the location of a class or other school property assets. A Facility can be a classroom, a lecture hall, a laboratory, or the location of a campus building that doesn't directly host student activity (for example, Maintenance/IT Facilities).
* We've also added a Facility lookup field on the Course Offering object so you can associate a single location to multiple course offerings (for example, a single lecture hall can host various courses). If you want to track locations this way, add the Facility field to the Course Offering page layout.

# Issues Closed
Closes #716 

# New Metadata
* New Custom Field on the Course Offering: `Facility__c`
* New Custom Object: `Facility__c`
  * `Facility__c.Account__c`
  * `Facility__c.Capacity__c`
  * `Facility__c.Description__c`
  * `Facility__c.Parent_Facility__c`
  * `Facility__c.Type__c`
* New Tab: `Facility__c`

# Deleted Metadata

# Testing Notes


src/objects/Facility__c.object
```diff
@@ -0,0 +1,175 @@
...
+        <fields>Capacity__c</fields>
+        <fields>Type__c</fields>
+        <label>HEDA Facility Compact Layout</label>
+    </compactLayouts>
+    <deploymentStatus>Deployed</deploymentStatus>
+    <description>Physical location of a building or room that allows an educational institution to house information with respect to where classes are being held and other school property assets.</description>
+    <enableActivities>false</enableActivities>
+    <enableBulkApi>true</enableBulkApi>
+    <enableFeeds>false</enableFeeds>
+    <enableHistory>false</enableHistory>
```
robertbwatson: Only Affiliations & Relationships have History Tracking enabled by default - I'm not really sure why (if there is a reason)

src/objects/Facility__c.object
```diff
@@ -0,0 +1,175 @@
...
+    <enableReports>true</enableReports>
+    <enableSearch>true</enableSearch>
+    <enableSharing>true</enableSharing>
+    <enableStreamingApi>true</enableStreamingApi>
+    <fields>
+        <fullName>Account__c</fullName>
+        <deleteConstraint>SetNull</deleteConstraint>
+        <description>The organization that owns/controls this facility.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The organization that owns/controls this facility.</inlineHelpText>
```
robertbwatson: Most objects don't have help text, but I think we need to have it. I'm not sure why - it's either tech debt or it's assumed that most fields are self-explanatory?

src/objects/Facility__c.object
lifewithryan: we should talk about this because I normally enable all of that. Maybe I shouldn't be or maybe we all should be? :)

robertbwatson: Yeah, I'm not sure - I can do a quick check this week/next week to see if these are things that can be enabled by an admin. If not, I'm not sure why they aren't enabled on our objects.

src/layouts/Course_Offering__c-HEDA Course Offering Layout.layout
```diff
@@ -22,6 +22,10 @@
                 <behavior>Edit</behavior>
                 <field>Capacity__c</field>
             </layoutItems>
+            <layoutItems>
```
robertbwatson: I'm assuming this location is okay but I need to confirm the PMs next week.

src/layouts/Facility__c-HEDA Facility Layout.layout
```diff
@@ -0,0 +1,96 @@
...
+    <excludeButtons>Submit</excludeButtons>
+    <layoutSections>
+        <customLabel>false</customLabel>
+        <detailHeading>false</detailHeading>
+        <editHeading>true</editHeading>
+        <label>Information</label>
+        <layoutColumns>
+            <layoutItems>
+                <behavior>Required</behavior>
+                <field>Name</field>
```
robertbwatson: I haven't yet reviewed this layout with PMs, so will need to do so next week.

robertbwatson:
@juliesnow I added you as a reviewer on this so that you can read over the release notes and make any changes before we merge.

src/layouts/Facility__c-HEDA Facility Layout.layout
robertbwatson: Reviewed

src/objects/Facility__c.object
robertbwatson: Confirmed that these are things that can be enabled by an admin who's installed the managed package.

src/objects/Facility__c.object
```diff
@@ -0,0 +1,189 @@
...
+        <type>Picklist</type>
+    </fields>
+    <label>Facility</label>
+    <listViews>
+        <fullName>All</fullName>
+        <filterScope>Everything</filterScope>
+        <label>All</label>
+    </listViews>
+    <nameField>
+        <label>Facility Name</label>
```
robertbwatson: @lifewithryan I don't quite remember why we landed on doing it this way, but I'm wondering why I didn't explore the option of just having this be an autonumber. There could be a separate "Facility Name" field that would be something like "Room 201". The admin could decide to create a formula field that displays a more friendly name based on the Account or Parent Facility. This could negate our need for custom facility naming logic. Perhaps search would be affected by this? (Perhaps it's too late to go down this path...)

Account naming makes sense because it can't be achieved through a formula since its many contacts to an account. But this is a one-to-one relationship between a Facility and either a Parent Facility or an Account...

src/objects/Facility__c.object
lifewithryan: likely tech debt

src/objects/Facility__c.object
lifewithryan: This is a very good point. So are you suggesting we have something like FAC-1021 and then we perhaps "ProTip" their friendly name? They could then add it to the fields that show in the search view?

I think I like this idea, but before we jump on it, we should ask @aligeez and @JasonFitzy. We can always add code later.

lifewithryan: I mentioned this in our HEDA chat so we can have a bigger discussion. :)

aligeez:
> My only concern here is if we are comfortable with the list of facilities we've added or is it too much? Otherwise it all looks good.

@lifewithryan - If you want to revisit this list and super simplify it the way we did with Attribute types I am okay with that. We can always move the fuller list of "most likely want to include these" in unmanaged metadata (per Blanksi's suggestion) or in a how-to/Pro-Tips area.

robertbwatson:
@juliesnow Similar to my comment in the other PR, we might want to consider adding in some text about page layout updates.

For this PR, I would suggest:
* Customers should add the "Facility" field to their Course Offering page layout if they want to track the location of a course offering.

cc @lifewithryan

aligeez:
> @juliesnow Similar to my comment in the other PR, we might want to consider adding in some text about page layout updates.
> 
> For this PR, I would suggest:
> 
> * Customers should add the "Facility" field to their Course Offering page layout if they want to track the location of a course offering.
> 
> cc @lifewithryan

@juliesnow @lifewithryan  -- Just need to make sure all the release notes at this juncture of our development to inform the customers it's just schema (not logic/automation). Is that best handled on a per-bullet release or some sort of parent bullet point that says _**"everything delivered below is schema only, start brainstorming on how your institution could use this object and report back either directly with our product managers or on the power of us hub.  We'll be working later in the year on the logic and automation around these objects with two tenets in mind; 1.) Don't solution; keep HEDA a flexible and open architecture, and 2.) Community Driven development!"**_

robertbwatson:
@lifewithryan Minor change since you last approved in order to get this passing on API version 45.0:  https://github.com/SalesforceFoundation/HEDAP/pull/725/commits/b13eccfddba40b9727deb83ce930041497b55464

There have also been some other minor commits to update the descriptions & help text after more review from the docs team.

src/objects/Facility__c.object
robertbwatson: Update here... I left this as-is as we discussed earlier today. We're going to add in an unmanaged formula field in #720

PR #726, opened 2019-03-01
Creation of new Objects (Language & Contact Language) & fields to support Languages in HEDA

### LINKED ISSUE #715: Creation of new Objects (Language & Contact Language) & fields to support Languages in HEDA

Feature

The current functionality does not exist in HEDA.

This issue will result in new objects & fields and not any new logic/functionality. The user will interact with these objects the same way they interact with other objects in Salesforce.

A user can navigate to a Contact record to view or set their preferred language. There may be other languages that the Contact can speak that will also be tied to that Contact, as well as their proficiency in those languages.

These new objects & fields will allow an organization to track what languages are spoken by students/staff/parents, their fluency with each, and where they are spoken. 

The language record will track the preferred (primary) language of a Contact in HEDA. The ability for an educational institution to know that Student John's primary language is English, but the best way to communicate with his parents' is Spanish. The ability to know the fluency of an instructor's languages in a global institution being taught in multiple countries or countries with more than one language; e.g. Belgium with Flemmish and French.

robertbwatson:
# Critical Changes

# Changes
* We've added two new objects to track the languages that are spoken by students, staff, and parents, as well as their fluency in each. First, create a Language record at the org level for each language recognized at your school, then create Contact Language records for the primary language of a Contact and the fluency of other languages for that Contact.
Because a Contact can be any type of individual, you have a lot of flexibility in the type of data you track. For example, you might use a Contact Language record to track the primary language for a student as English, but then identify German as an intermediate level language because he's studying it. Or, you might track the fluency of an instructor's languages at a global institution where courses are taught in multiple countries.
To enable tracking the fluency of certain languages for a Contact, add the Contact Languages related list to the Contact page layout. To display the primary language for a Contact, add the Primary Language field to the Contact page layout. This field will be automatically populated in a future release, so we recommend you set it as read-only.

# Issues Closed
Closes #715 

# New Metadata
* New Custom Field on the Contact: `Primary_Language__c`
* New Custom Object: `Contact_Language__c`
  * `Contact_Language__c.Contact__c`
  * `Contact_Language__c.Fluency__c`
  * `Contact_Language__c.Language__c`
  * `Contact_Language__c.Primary_Language__c`
* New Custom Object: `Language__c`
* New Tab: `Contact_Language__c`
* New Tab: `Language__c`

# Deleted Metadata

# Testing Notes


src/objects/Contact_Language__c.object
```diff
@@ -0,0 +1,115 @@
...
+        <externalId>false</externalId>
+        <inlineHelpText>The language this contact is associated with.</inlineHelpText>
+        <label>Language</label>
+        <referenceTo>Language__c</referenceTo>
+        <relationshipLabel>Contact Languages</relationshipLabel>
+        <relationshipName>Contact_Languages</relationshipName>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Lookup</type>
+    </fields>
```
lifewithryan: @robertbwatson -- are we not also providing a picklist for fluency here? I review the document and it looks like we decided that we may have some EdFi specific values (for certification) but I still think we want the field on the object in HEDA with perhaps "Conversational|Can Read" or something similar.

robertbwatson: @lifewithryan Sorry, I created these pull requests as "draft" PRs and should have been more clear that I did so - I fully intend to make more modifications here and it is not ready for review. (Just announced recently - https://github.blog/2019-02-14-introducing-draft-pull-requests/)

lifewithryan: AHHHHH :) thanks

robertbwatson:
@juliesnow  I added you as a reviewer on this so that you can read over the release notes and make any changes before we merge.

dev_config/src/admin_config/profiles/Admin.profile
```diff
@@ -62,6 +62,10 @@
         <tab>%%%NAMESPACE%%%Affiliation__c</tab>
         <visibility>DefaultOn</visibility>
     </tabVisibilities>
+    <tabVisibilities>
```
robertbwatson: @lifewithryan @xinnanlisf You both might already be aware of this, but this is what I was talking about the other day during standup where I couldn't figure out why I had no access to the new tabs when I spun up the scratch org.

The trick is to edit this file and add the tab visibilities for the Admin profile. This gets deployed to the scratch org in the deploy_dev_config task as part of the dev_org flow.

src/layouts/Contact-HEDA Contact Layout.layout
```diff
@@ -86,6 +86,10 @@
                 <behavior>Edit</behavior>
                 <field>Financial_Aid_Applicant__c</field>
             </layoutItems>
+            <layoutItems>
```
robertbwatson: I guessed where this field should be. I should review this with the PMs in a meeting next week.

src/objects/Contact.object
```diff
@@ -3671,6 +3671,20 @@
         <trackFeedHistory>false</trackFeedHistory>
         <type>Lookup</type>
     </fields>
+    <fields>
+        <fullName>Primary_Language__c</fullName>
+        <deleteConstraint>SetNull</deleteConstraint>
+        <description>The language the contact prefers to use.</description>
```
robertbwatson: Per Julie, this should be "Contact" (capitalized) - will fix this afternoon

src/layouts/Contact-HEDA Contact Layout.layout
```diff
@@ -322,6 +322,13 @@
         <fields>Grade__c</fields>
         <relatedList>Course_Enrollment__c.Contact__c</relatedList>
     </relatedLists>
+    <relatedLists>
```
lifewithryan: I don't think we want to add the related list by default but rather call out in the docs, "if you plan to leverage the new language tracking functionality, you will want to add the Contact Language related list to the HEDA Contact Page Layout"

lifewithryan: @juliesnow ^^please see above comment.

src/objectTranslations/Contact-ca.objectTranslation
```diff
@@ -4444,6 +4444,12 @@
         <name>Primary_Household__c</name>
         <relationshipLabel>Contactes (llar primària)</relationshipLabel>
     </fields>
+    <fields>
+        <help><!-- The language the contact prefers to use. --></help>
```
lifewithryan: I thought I saw somewhere in reading this that the "contact" in the help should be "Contact" with a capital "C" -- maybe I misinterpreted but wanted to call out.

src/objectTranslations/Contact-es.objectTranslation
```diff
@@ -3522,6 +3522,12 @@
         <name>Primary_Household__c</name>
         <relationshipLabel>Contactos (domicilio principal)</relationshipLabel>
     </fields>
+    <fields>
+        <help><!-- The language the contact prefers to use. --></help>
```
lifewithryan: Contact vs contact -- assume this comment for any instance like that in the help lines. Unless @juliesnow says differently, will defer to docs :)

src/objects/Contact.object
```diff
@@ -3671,6 +3671,20 @@
...
+        <externalId>false</externalId>
+        <inlineHelpText>The language the contact prefers to use.</inlineHelpText>
+        <label>Primary Language</label>
+        <referenceTo>Contact_Language__c</referenceTo>
```
lifewithryan: @robertbwatson -- just a quick question here -- should this be a reference to the "Contact Language" or the actual Language base object? I ask because the reference is circular, Contact_Language looks up to Contact which has a lookup to Contact Language.  if we populate the Language__c in the lookup instead can potentially go to a Language object and view its related list and see all the contacts who have that language set as primary. I think this is what I had in mind originally. Does that break things? I feel it doesn't but feel free to yell at me :)

src/layouts/Contact-HEDA Contact Layout.layout
robertbwatson: @lifewithryan I just reviewed this with @aligeez  and @JasonFitzy yesterday so I'm tagging them here - should the Contact Language related list be added to the HEDA Contact page layout by default, or should it be a documentation item for them to add it if they want it?

src/objectTranslations/Contact-ca.objectTranslation
robertbwatson: Yep, I added a comment for myself further down in this diff to change it. Will do today.

src/objects/Contact.object
robertbwatson: @lifewithryan That shouldn't break things and probably makes more sense. This will get around the issue of potentially selecting a Contact Language record that isn't even related to the contact.

I would argue then that this should still not be on the page layout by default, and if it's added, we should recommend that it is read-only (so that the custom code sets it when creating the Contact Langauge record)

The spec has it as Contact Language so I'll be sure to update that, and we'll need to update the ERD.

robertbwatson: "view its related list and see all the contacts who have that language set as primary." -- Since we have the Contact Language related list on the Language page layout already, which shows the contact, fluency, and if it's the primary language, I think we're okay here and probably don't need to add the additional related list on the Language object. But the lookup is still good to have.

robertbwatson:
@lifewithryan I made the requested changes here, except for removing the related list (waiting to hear back from PMs on this)

src/layouts/Contact-HEDA Contact Layout.layout
lifewithryan: It sounds like we should add them by default. That is the precedence that has been set. Page Layouts apparently don't get push upgraded so the additions will only apply to new installations. Existing customers will have to add them on their own as well as setup FLS for new fields on existing objects, etc. Things that we need to call out in Docs. cc: @juliesnow

lifewithryan: Ignore my first comment up there about not including it...I was w-ww-wro--- I was wwwroo....I was wrong...

src/objects/Contact.object
lifewithryan: agreed, at this point, having the related list would just be a "convenience" sort of thing. Largely unusable in sizeable orgs...I think I'm just used to working with smaller footprints :)

lifewithryan: With respect to having the field be Read Only. I think that is okay but, maybe we should consider that if someone sets that field, we would in turn create (if it didn't exist) a Contact Language Record for that language on the contact and set the Primary checkbox to true.

robertbwatson: I think that could maybe be a future work item/improvement. I'm not sure if we want to handle that now. Would we assume that the Fluency would be "Fluent"? (Let the user decide with a new custom setting?) etc

It might be worthwhile to have now, in which case I could ask if it makes sense to do in W-031856 or a separate WI.

lifewithryan: If it's my primary language it better be fluent, but that's a good point...

robertbwatson:
> Updated Changes text as discussed offline.

@juliesnow  Thanks! As @lifewithryan & others have discussed in Quip, should we consider adding something about page layout changes to the release notes?

In this PR, I might recommend something like:
* Existing customers may want to consider adding the "Contact Languages" related list to their Contact page layouts if they want to use this functionality.
* Additionally, all current and new customers may want to consider adding the "Primary Language" field to their Contact page layouts as read-only.

Thoughts?

aligeez:
> > Updated Changes text as discussed offline.
> 
> @juliesnow Thanks! As @lifewithryan & others have discussed in Quip, should we consider adding something about page layout changes to the release notes?
> 
> In this PR, I might recommend something like:
> 
> * Existing customers may want to consider adding the "Contact Languages" related list to their Contact page layouts if they want to use this functionality.
> * Additionally, all current and new customers may want to consider adding the "Primary Language" field to their Contact page layouts as read-only.
> 
> Thoughts?

@juliesnow - I like including both bullet points in the release notes, but the term "functionality" in the first bullet might make some customer think there is a bit of automation or logic being delivered with this object. @lifewithryan - correct me if I'm wrong, but this is just the object/fields/scheme we're delivering, correct? If that is accurate we may want to state that somehow in the Rel Notes bullet.

robertbwatson:
@lifewithryan Minor change since you last approved in order to get this passing on API version 45.0: https://github.com/SalesforceFoundation/HEDAP/pull/726/commits/399489c889267048079c96201bd020c658e8d4f8

PR #733, opened 2019-03-06
Update HEDA API Version

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


cumulusci.properties
```diff
@@ -1,6 +1,6 @@
 cumulusci.package.name=HEDA
```
jstvz: Actually, can we delete `cumulusci.properties`?

xinnanlisf: Well is that a question or a request ;?

lifewithryan:
@jstvz we need a review from Release Engineering but don't merge upon approval because we want QA to test first (that may be obvious).

src/applications/HEDA.app
```diff
@@ -2,13 +2,16 @@
 <CustomApplication xmlns="http://soap.sforce.com/2006/04/metadata">
     <defaultLandingTab>standard-home</defaultLandingTab>
     <description>Higher Education Data Architecture</description>
+    <formFactors>Large</formFactors>
```
xinnanlisf: @lifewithryan I am hesitating on these lines too. Not sure if we want to include them here..

src/objects/Hierarchy_Settings__c.object
```diff
@@ -95,28 +95,28 @@
     <fields>
         <fullName>Affl_ProgEnroll_Copy_End_Date__c</fullName>
         <defaultValue>false</defaultValue>
-        <description>Indicates whether to set the End Date of the Affiliation created with a Program Enrollment. If enabled, set the End Date to the Program Enrollment&apos;s End Date. If disabled, don&apos;t specify the Affiliation&apos;s End Date.</description>
+        <description>Indicates whether to set the End Date of the Affiliation created with a Program Enrollment. If enabled, set the End Date to the Program Enrollment&apos;s End Date. If disabled, don't specify the Affiliation&apos;s End Date.</description>
```
lifewithryan: @xinnanlisf -- I remember these coming across in a different PR and we had replaced the individual apostrophe marks with the encoded &apos; value instead.  I see that here we appear to be putting them back to the individual apostrophe mark. Is there a reason for this? if you look at these lines, it appears that for we aren't being consistent. We have Enrollment&apos;s for "Enrollment's" and we have "don't" later on in the text without the &apos; encoded version

xinnanlisf: Hi @lifewithryan, yeah I saw that this morning too and I believe the changes have been reverted too ;)

xinnanlisf:
Hi @juliesnow, could you take a review on the release note please? Thank you!

PR #754, opened 2019-03-20
Add "Display Name" formula field to the Facility object

### LINKED ISSUE #720: Facility Naming: Include a "Display Name" formula field on the Facility that includes the Account name or Parent Facility name

robertbwatson:
# Critical Changes

# Changes
* We've added a formula field (Display Name) to the Facility object. The Display Name includes the name of the Parent Facility or Account if a Parent Facility or Account is associated with the Facility, for example, "Thomas Edison Building - Room 201" instead of "Room 201". If the Facility has both an Account and a Parent Facility, the name of the Parent Facility is used. This formula field can be updated with different criteria if a better naming format fits your needs.
  * Note: This field is not added to the Facility layout by default. We recommend that you add it to page layout, as well as to the various Search Layouts in `Setup -> Object Manager -> Facility -> Search Layouts`.

# Issues Closed
Closes #720 

# New Metadata
* New custom field on the Facility: `Display_Name__c`

# Deleted Metadata

# Testing Notes


robertbwatson:
Leaving this as a draft PR for now. The Facility PR #725 needs to be merged in first.

unpackaged/post/facility_display_name/objects/___NAMESPACE___Facility__c.object
```diff
@@ -0,0 +1,22 @@
...
+        <description>The full name of the Facility. For example, &quot;Thomas Edison Building - Room 201.&quot;</description>
+        <externalId>false</externalId>
+        <formula>IF(
+    NOT(ISBLANK(Parent_Facility__c)), Parent_Facility__r.Name + &apos; - &apos; + Name,
+    IF(
+        NOT(ISBLANK(Account__c)), Account__r.Name + &apos; - &apos; + Name, 
+        Name
+    )
+)</formula>
+        <formulaTreatBlanksAs>BlankAsZero</formulaTreatBlanksAs>
```
lifewithryan: @robertbwatson -- I noticed this is blanks as zero, in my experience the platform does the right thing when working with text and this set as such, but just thought I'd ask if we want to consider "blanks as null." I believe we are fine either way.

robertbwatson: @lifewithryan Oh, weird. I'm going to keep this as-is, but good catch. I see the "Relationship Explanation" field on the Relationship object is the same. There's not actually a way to change this value in the UI, so this is what I got back from Salesforce when doing a retrieve.

PR #756, opened 2019-03-20
Do not allow deletion of Language records if there are Contact Languages records for that Language.

### LINKED ISSUE #748: Language: Do not allow deletion of a Language record if there are Contact Language records associated to it
Feature

A Language record can be deleted. If there are Contact Language records associated to it, the Language lookup field is simply cleared.

A friendly error should be shown telling the user that the Language record cannot be deleted.
An admin cannot create a Contact Language record that does not lookup to a Language. In other words, "Language" is a required field on "Contact Language" 

Deleting a Language record when there are Contact Language records still associated to it will lead to data loss in the customer org.

* **Acceptance Criteria**

  - **Given**: There's a Language record with one or multiple Contact Language records associated to it
  - **When**: I attempt to delete that Language record
  - **Then**: A friendly error telling me that the Language record cannot be deleted should appear.

robertbwatson:
# Critical Changes

# Changes
* You can no longer delete Language records if a Contact Language record is associated with that Language. You'll receive an error if you attempt to do so.

# Issues Closed
Closes #748 

# New Metadata

# Deleted Metadata

# Testing Notes


robertbwatson:
@lifewithryan This is a really quick and easy PR. My only concern is whether or not we are actually allowed to do this after the field has already been released - I don't know if there's an easy way to know this until we try to upload a new beta...

lifewithryan:
> @lifewithryan This is a really quick and easy PR. My only concern is whether or not we are actually allowed to do this after the field has already been released - I don't know if there's an easy way to know this until we try to upload a new beta...

that is a very good question. Let's see if we can find out from RelEng if we can actually do that.

Issue #746, opened 2019-03-14
Language: Contact "Primary Language" Updates Contact Language Record


robertbwatson:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
With the release of #741, the Contact's "Primary Language" field is automatically updated when a Contact Language record is marked as the Primary Language.

However, a user can decide to update this lookup on the Contact at any time. Nothing currently happens when the Primary Language lookup field on the Contact is updated.

* **What is the expected behavior?**
If a user updates the Primary Language lookup on the Contact, the system should search for an existing Contact Language record for that Language and set it as the Primary Language. If a record doesn't already exist, a record should be created.

* **What is the motivation / use case for changing the behavior?**
We have automation for setting this lookup field when the user creates or modifies a Contact Language record, but not the other way around.

"As an Admin, when I set the Primary Language field on a contact, it should create a Contact Language record for the contact and mark it as primary if it doesn't already exist."

* **Acceptance Criteria**

  - **Given**: I have an "English" Language record and am updating a Contact record
  - **When**: I set the Contact's Primary Language lookup to "English"
  - **Then**: The system will search for an existing Contact Language record for the given Contact/Language. Since one does not exist, it will create a new Contact Language record with Fluency set to the default value and Primary Language checked.

* **Acceptance Criteria**

  - **Given**: I have an "English" Language record and am updating a Contact record. The Contact already has an "English" Contact Language record, but it is not set as the contact's Primary Language.
  - **When**: I set the Contact's Primary Language lookup to "English"
  - **Then**: The system will search for an existing Contact Language record for the given Contact/Language. It finds the existing record and checks the Primary Language checkbox.

* **Acceptance Criteria**

  - **Given**: I have an "English" Language record and am updating a Contact record. The Contact's Primary Language is currently set to "French".
  - **When**: I set the Contact's Primary Language lookup to "English"
  - **Then**: In addition to searching for an existing "English" Contact Language record and setting the Primary Language checkbox to true, the system also updates the "French" Contact Language record to no longer have the Primary Language checkbox checked.

aligeez:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000008LljBQAS" target="_blank">W-031985</a>

mrbelvedere:
Included in beta release 1.75 (Beta 2)

mrbelvedere:
[...]

PR #761, opened 2019-03-25
Contact "Primary Language" Automatic Population

### LINKED ISSUE #741: Language: Contact "Primary Language" Automatic Population
Feature

When a new Contact Language record is marked as the Primary Language for that Contact, the Primary Language lookup field on the Contact does NOT get populated.

When a new Contact Language record is marked as the Primary Language for the Contact, the Primary Language lookup field on the Contact should be automatically populated with the Language record.

By automatically setting the lookup field, the user can easily see the primary language for that Contact without having to go to the related list.

* **Acceptance Criteria**

  - **Given**: I set the Primary Language checkbox on the Contact Language record
  - **When**: I save the Contact Language record
  - **Then**: The Primary Language lookup on the related Contact should be set to the Language record

* **Acceptance Criteria**

  - **Given**: I unset the Primary Language checkbox on the Contact Language record
  - **When**: I save the Contact Language record
  - **Then**: The Primary Language lookup on the related Contact should be cleared

* **Acceptance Criteria**

  - **Given**: I set the Primary Language checkbox on the Contact Language record and there's an existing Contact Language record already set as the Primary Language
  - **When**: I save the Contact Language record
  - **Then**: The other Contact Language record should be unset as the Primary Language, and the Primary Language lookup on the related Contact should be set to the new Language record

* **Acceptance Criteria**

  - **Given**: There's an existing Contact Language record that is set as the Primary Language
  - **When**: I delete the Contact Language record
  - **Then**: The Primary Language lookup on the related Contact should be cleared

robertbwatson:
# Critical Changes

# Changes
* Now, when you select the "Primary Language" checkbox on the Contact Language object, the associated Contact's "Primary Language" field is updated to match the Language record. Likewise, if you deselect the "Primary Language" checkbox, the associated Contact's "Primary Language" field is cleared.
Only one Contact Language can be defined as the Primary Language for a Contact. If you attempt to select a second Contact Language as the Primary Language, your second selection replaces the first.

# Issues Closed
Closes #741 

# New Metadata
* `CLAN_PrimaryLanguage_TDTM.cls`
* `CLAN_PrimaryLanguage_TEST.cls`
* `TDTM_ContactLanguage.trigger`

# Deleted Metadata

# Testing Notes


src/classes/CCON_ConnectionBackfill_TEST.cls
```diff
@@ -30,7 +30,7 @@
 /**
 * @author Salesforce.org
 * @date 2016
-* @group CourseEnrollments
+* @group Course Enrollments
```
robertbwatson: I updated several files when reviewing our ApexDocs and noticing some inconsistencies. If we don't want to include these changes in this PR, I can just revert the one commit.

robertbwatson:
@xinnanlisf, I assigned this to Ryan but feel free to have a look through as well, if you have time!

src/classes/CLAN_PrimaryLanguage_TDTM.cls
```diff
@@ -0,0 +1,240 @@
...
+
+        // Query for any existing Contact Language records for the Contacts with Primary Language set to true, but not
+        // in the given set of primary Contact Languages.
+        List<Contact_Language__c> oldPrimaryContactLanguages = 
+            [SELECT Id, Primary_Language__c FROM Contact_Language__c 
+                WHERE Id NOT IN :primaryContactLanguages AND Contact__c IN :contacts AND Primary_Language__c = true];
+        
+        // Update the Contact Language records to no longer be marked as the Primary Language.
+        for (Contact_Language__c oldPrimaryContactLanguage : oldPrimaryContactLanguages) {
+            oldPrimaryContactLanguage.Primary_Language__c = false;
```
robertbwatson: I don't really know if there is a use case where a customer may want to be able to have more than one primary language defined -  I wouldn't think so. But if they do, I'd expect they can just turn off this trigger.

src/classes/CCON_ConnectionBackfill_TEST.cls
lifewithryan: No this is fine.

src/classes/CLAN_PrimaryLanguage_TDTM.cls
lifewithryan: Correct. I suppose someone always could, but I would go so far as to say it's an edge case and we're not going to support it outside of letting them turn off the trigger.

src/classes/CLAN_PrimaryLanguage_TEST.cls
```diff
@@ -0,0 +1,268 @@
...
+        contact = UTIL_UnitTestData_TEST.getContact();
+        insert contact;
+    }
+    
+    /*******************************************************************************************************
+    * @description Inserts a new Language record for testing. 
+    * @param name The name of the Language record.
+    * @return The inserted Language record.
+    ********************************************************************************************************/
+    private static Language__c insertLanguage(String name) {
```
lifewithryan: @robertbwatson should we consider adding this to the UTIL_UnitTestData_Test.cls and potentially the Unit Test API classes? Any of the data we use to setup our tests, I'm thinking should probably go in that class. Whether or not we expose a global method int he API class is debatable, but thinking we probably want to leverage UTIL_UnitTestDAta_Test.cls

src/classes/CLAN_PrimaryLanguage_TEST.cls
```diff
@@ -0,0 +1,268 @@
...
+
+        assertPrimaryLanguageOnContact(language.Id);
+    }
+
+    /*******************************************************************************************************
+    * @description Tests that when multiple Contact Languages are updated with Primary Language checked, the 
+    * Primary Language is set on the associated Contact of the most recent Primary Language and the other
+    * Contact Language record is unset as the Primary Language.
+    ********************************************************************************************************/
+    @isTest
```
lifewithryan: it might just be because I am tired but I'm having a hard time following this test. Maybe its because the logic is broken out and I have to go back and ensure I am looking at the method you're calling. I'm questioning here though if updating the primary languages at the same time in a list like that guarantees the same order of operations? Can we safely assume that firstLanguage is the one that will get marked as not primary and instead the secondLanguage will be primary? I'm not sure I am making sense.
Generally in unit tests, and maybe we should chat about this -- I actually prefer brittle, explicit code as opposed to re-usable code. (you may feel differently, that's fine). Is this testing creating both at the same time? Or is it testing that one is overwriting the other? (Sorry if I am not making sense...)

src/classes/TDTM_DefaultConfig.cls
```diff
@@ -49,137 +49,142 @@ public without sharing class TDTM_DefaultConfig {
         // Account Management on Contact
         handlers.add(new Trigger_Handler__c(Active__c = true, Asynchronous__c = false, 
               Class__c = 'ACCT_IndividualAccounts_TDTM', Load_Order__c = 1.00, Object__c = 'Contact', 
-              Owned_by_Namespace__c = 'hed' ,Trigger_Action__c = 'AfterInsert;AfterUpdate;AfterDelete'));
+              Owned_by_Namespace__c = 'hed', Trigger_Action__c = 'AfterInsert;AfterUpdate;AfterDelete'));
```
lifewithryan: THANK YOU :)

src/classes/CLAN_PrimaryLanguage_TEST.cls
robertbwatson: I'm torn here. I definitely considered it, and I definitely think it should be done if we end up having more than one unit test class that creates Language records (which will/should happen when I complete the #746 issue).

src/classes/CLAN_PrimaryLanguage_TEST.cls
robertbwatson: No, your concern is valid and this is the test that I like the least, so thanks for calling it out. The test is updating both at the same time. I don't think we can assume that firstLanguage is the one that will get marked as not primary. Perhaps a better test is to assert that only one of the two records remains Primary after the unit tests runs.

src/classes/CLAN_PrimaryLanguage_TEST.cls
robertbwatson: Done!

src/classes/CLAN_PrimaryLanguage_TEST.cls
robertbwatson: Done!

robertbwatson:
@lifewithryan Ready for your re-review. See commit https://github.com/SalesforceFoundation/HEDAP/pull/761/commits/2bf748769160ae3cfe3278bcdf8b5f187e589157

PR #762, opened 2019-03-26
Create Object/Fields for Term Grade

### LINKED ISSUE #738: New HEDA Object: Term Grades
**New Object in HEDA Framework**: The purpose of this new object is to allow institutions to record grades for students in a given course offering during a particular term and across terms. It is not intended to be used for grading an assignment or quiz, but recording a final grade, or a progress report grade, or recording of a grade for an institutionally defined grading period.

In our effort to support P-20 lifelong learning/education commitment.

* **Acceptance Criteria**

- As an admin I can log into my HEDA org and see the new object "Term Grades"
- As an admin I can see the necessary fields included in  "Term Grades" (see quip doc for specifics)

xinnanlisf:
# Critical Changes

# Changes
## New Term Grade Object
* We've created a new Term Grade object along with relevant fields that you can use to track grades for students in a given Course Offering during a particular Term. At this time, the Term Grade object is not intended for use as a general gradebook for assignments, quizzes, and the like. Instead, use it to track key grades for a given Term like mid-term exams or final grades. You can track these grades using numerical, percentage, or letter values.
* The default Term Grade Page Layout comes with all the fields listed for you to customize (except for Contact, which will be coming in a future release.)  Go to Setup | Object Manager | Term Grades | Page Layouts to select and arrange the fields in an order most relevant to your use cases. We also recommend requiring important fields (for example, Term, Term Grade, and/or Course Connection).
* To best use this new object, add the Term Grades related list to the HEDA Course Connection Page Layout.

# Issues Closed
Fixes #738 

# New Metadata
Term_Grade__c-HEDA Term Grade Layout.layout
Term_Grade__c.object
Term_Grade__c-ca.objectTranslation
Term_Grade__c-en_US.objectTranslation
Term_Grade__c-es.objectTranslation
Term_Grade__c-fr.objectTranslation

# Deleted Metadata

# Testing Notes


robertbwatson:
@xinnanlisf What issue # does this fix?

I also think we need release notes review from Julie - I've added her as a reviewer.

We might also want to think about being more consistent about what goes under "New Metadata" - for example, in #726 I included the fields as well. (Not as big of a deal I think, since we have the Data Dictionary.)

src/layouts/Term_Grade__c-HEDA Term Grade Layout.layout
```diff
@@ -0,0 +1,110 @@
...
+    <layoutSections>
+        <customLabel>false</customLabel>
+        <detailHeading>false</detailHeading>
+        <editHeading>true</editHeading>
+        <layoutColumns/>
+        <layoutColumns/>
+        <layoutColumns/>
+        <style>CustomLinks</style>
+    </layoutSections>
+    <relatedLists>
```
robertbwatson: I personally don't think I was including the tasks related list by default, but I don't know if it matters much. (Would be nice to know which layouts we do include these and which ones we don't - how inconsistent are we?)

src/objects/Term_Grade__c.object
```diff
@@ -0,0 +1,233 @@
...
+        <fullName>Course_Connection__c</fullName>
+        <deleteConstraint>SetNull</deleteConstraint>
+        <description>The class in which the student is received this grade.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The class in which the student is received this grade.</inlineHelpText>
+        <label>Course Connection</label>
+        <referenceTo>Course_Enrollment__c</referenceTo>
+        <relationshipLabel>Term Grades</relationshipLabel>
+        <relationshipName>Term_Grades</relationshipName>
+        <required>false</required>
```
robertbwatson: Just checking - this isn't required, is that intentional? This is currently the only way that a student is tied to a grade, unless we want to make this flexible such that a customer can add a Contact lookup or something if that's desired instead.

src/objects/Term_Grade__c.object
```diff
@@ -0,0 +1,233 @@
...
+        <description>The type of term or grading period for which this grade is being recorded.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The type of term or grading period for which this grade is being recorded.</inlineHelpText>
+        <label>Term Grade Type</label>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Picklist</type>
+        <valueSet>
+            <restricted>true</restricted>
+            <valueSetDefinition>
```
robertbwatson: Ali's comment in the specs say these should be alphabetized.

src/objects/Term_Grade__c.object
```diff
@@ -0,0 +1,233 @@
...
+    </fields>
+    <fields>
+        <fullName>Percent_Grade__c</fullName>
+        <description>The grade the student earned for this course represented as a percentage. For example, 87.5%, which is equivalent to a B letter grade in most cases.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The grade the student earned for this course represented as a percentage. For example, 87.5%, which is equivalent to a B letter grade in most cases.</inlineHelpText>
+        <label>Percent Grade</label>
+        <precision>18</precision>
+        <required>false</required>
+        <scale>0</scale>
```
robertbwatson: Is this correct? I may need to deploy to my org to see. This reads to me that only full percentages are allowed.

src/package.xml
```diff
@@ -754,6 +769,7 @@
         <members>Program_Enrollment__c-HEDA Program Enrollment Layout</members>
         <members>Program_Plan__c-HEDA Program Plan Layout</members>
         <members>Relationship__c-HEDA Relationship Layout</members>
+        <members>Term_Grade__c-Term Grade Layout</members>
```
robertbwatson: It's actually the HEDA Term Grade Layout (so I don't think this layout would deploy successfully to a new org)

src/objects/Term_Grade__c.object
```diff
@@ -0,0 +1,233 @@
...
+        <description>The class in which the student is received this grade.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The class in which the student is received this grade.</inlineHelpText>
+        <label>Course Connection</label>
+        <referenceTo>Course_Enrollment__c</referenceTo>
+        <relationshipLabel>Term Grades</relationshipLabel>
+        <relationshipName>Term_Grades</relationshipName>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Lookup</type>
```
robertbwatson: Should we add a new related list to the Course Connection layout? Or have that be a "pro tip" / part of the K12 kit? (I also asked this in the spec doc, I think)

aligeez: If we add it, can we remove it later? if the answer is no - let's not go that far yet. We can always add it in a subsequent release. My vote is for "Pro Tip" for now. @robertbwatson

src/objects/Term_Grade__c.object
aligeez: def. not correct. Should be digits to the right of the decimal. I believe we settled on 3 for % and 5 for numerical grade.

src/objects/Term_Grade__c.object
aligeez: Would need this explained a bit more to weigh in. What's going on here exactly?

src/package.xml
robertbwatson: It did actually deploy successfully - surprising, but it should still be updated to "HEDA" here.

robertbwatson:
> Reviewed and will pull down to my own scratch org momentarily to do a quick test.
> 
> One final change: The admin.profile dev_config changes are missing for the new tab, which means that the new object won't be visible by default in all new dev/QA orgs. I think it should be added.

Interesting to note that the tab defaulted to "Default On" for my admin profile even without this change. I'm not sure why though, as for past objects I've had them not appear until I did this...

src/objects/Term_Grade__c.object
```diff
@@ -0,0 +1,233 @@
...
+    </fields>
+    <label>Term Grade</label>
+    <listViews>
+        <fullName>All</fullName>
+        <filterScope>Everything</filterScope>
+        <label>All</label>
+    </listViews>
+    <nameField>
+        <label>Term Grade Name</label>
```
robertbwatson: Per the specs, I expected this to be an autonumber and not have to specify a name.

src/objects/Term_Grade__c.object
robertbwatson: Course Connection isn't set as a required field, so someone could create a Term Grade record for nobody, essentially.

We can probably discuss this tomorrow when discussing Attendance, as I think similar questions also apply there.

src/objects/Term_Grade__c.object
```diff
@@ -0,0 +1,233 @@
...
+        <label>Letter Grade</label>
+        <length>255</length>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Text</type>
+        <unique>false</unique>
+    </fields>
+    <fields>
+        <fullName>Numerical_Grade__c</fullName>
+        <description>The numerical grade the student earned for this course.</description>
```
robertbwatson: As a general note, it's not clear to me, even with the help text, what the difference between this field and Percent Grade is.

src/objects/Term_Grade__c.object
aligeez: Gotcha. I bet I can conjure up a few use-cases where a  grade needs to be affiliated with a student (contact) without necessarily requiring a Course... like maybe there was a study abroad program and you get one final grade for it, and it's a program (Account), so we should tread lightly here. Agreed we can discuss Friday.

src/objects/Term_Grade__c.object
aligeez: I would defer to julie for help text, but in general, Numerical grade is usually out of 100 points. Very common for individual graded items, rubrics (K-12), etc.  e.g. 87.5/100 on Final Exam.  Sometimes it can even be out of 20, for example, Student Sophie got 18.5/20 on her final quiz. Percent grade is a % and can be calculated off Numerical sometimes.  We just have to cover both use-cases as different parts of the world, schools, verticals use diff. methods.

xinnanlisf:
> @xinnanlisf What issue # does this fix?
> 
> I also think we need release notes review from Julie - I've added her as a reviewer.
> 
> We might also want to think about being more consistent about what goes under "New Metadata" - for example, in #726 I included the fields as well. (Not as big of a deal I think, since we have the Data Dictionary.)

@robertbwatson I believe we use that section so that it is easier for release engineer to cut release. Once the object is added to the package, fields should be included automatically. Correct me if I am wrong for sure @jstvz

dgamboni:
> @xinnanlisf What issue # does this fix?
 (I added the issue #....)

src/objects/Term_Grade__c.object
dgamboni: @aligeez Sounds like you'd need a text field for your example of 18.5/20, or would they just be recording "18.5"?

src/layouts/Term_Grade__c-HEDA Term Grade Layout.layout
xinnanlisf: I think these are generated when activity enabled.

xinnanlisf:
> > @xinnanlisf What issue # does this fix?
> > (I added the issue #....)

@dgamboni thanks for adding the issue number. We usually say: Fixes #[Github Issue Number] ;)

src/objects/Term_Grade__c.object
robertbwatson: I talked to @lifewithryan about this and he suggested that maybe it is made required on the page layout.

src/objects/Term_Grade__c.object
robertbwatson: Looks like Julie updated this in the spec.

src/objects/Term_Grade__c.object
robertbwatson: Confirmed that I was not able to save with decimal points after the %

src/objects/Term_Grade__c.object
robertbwatson: We can remove it, but it wouldn't be removed for the new customers that got the update. We should be consistent here though - we've added them for other objects, I think. Thoughts @lifewithryan?

jstvz:
Basically, all the release notes generator does is move text from an issue heading to the corresponding one in the release notes.

But it would be nice to parse metadata as it is added, but that's a non-trivial addition that isn't planned at this time.

xinnanlisf:
@robertbwatson this PR is ready for review again!

src/objects/Term_Grade__c.object
robertbwatson: @xinnanlisf I'm not sure if you saw my last comment here but we should discuss with @aligeez & @lifewithryan if this should be required on the page layout.

src/objects/Term_Grade__c.object
robertbwatson: @xinnanlisf This field needs to be adjusted to match the spec definition.

src/objects/Term_Grade__c.object
robertbwatson: @lifewithryan  @xinnanlisf Did we decide today we're adding the Term Grades to the Course Connection by default (for new installs), or are we keeping it off?

robertbwatson: I am okay leaving it off for now since we can always add it in for new installs later, fyi

src/objects/Term_Grade__c.object
robertbwatson: I think since we decided today that we are going to add the Contact here in a new WI, that it's okay if this is not required.

PR #765, opened 2019-04-02
Missing class attribute in CMP_Acc_Naming_Dropdown

### LINKED ISSUE #757: Access Check Failed error on HEDA Settings. Missing class attribute in CMP_Acc_Naming_Dropdown
Bug

If Lighting debug mode is enabled for a user, that user will see the following error at the bottom of the HEDA Settings page

> This page has an error. You might just need to refresh it. Access Check Failed! AttributeSet.get(): attribute 'class' of component 'markup://hed:CMP_Acc_Naming_Dropdown {1150:0}' is not visible to 'markup://aura:if {1152:0}'. Failing descriptor: {aura:if}

The cause is in the CMP_Acc_Naming_Dropdown component. "v.class" is referenced several times in the markup but there is no aura:attribute named "class" defined. The error is getting caught and reported in debug mode, but otherwise you wouldn't see it.

robertbwatson:
# Critical Changes

# Changes
* An error is no longer displayed when viewing the HEDA Settings with Lightning Debug Mode enabled.

# Issues Closed
Fixes #757 

# New Metadata

# Deleted Metadata

# Testing Notes
* If Lightning Debug Mode is enabled for a user, that user would see an error at the bottom of the HEDA Settings page. The error was also displayed in a red toast message if the user saved the settings.
  * When the user navigates to the HEDA Settings, an error should no longer display. Settings should save successfully.

### DIFF
--- a/src/aura/STG_CMP_System/STG_CMP_System.cmp
+++ b/src/aura/STG_CMP_System/STG_CMP_System.cmp
@@ -200,7 +200,9 @@
         </div>
         
         <div class="slds-col slds-size--1-of-2" >
-            <c:CMP_Acc_Naming_Dropdown isView="{!v.isView}"
+            <c:CMP_Acc_Naming_Dropdown 
+                class="admin-account-naming"
+                isView="{!v.isView}"
                 setting="{!v.hierarchySettings.Admin_Account_Naming_Format__c}"
                 nameFormat="{!v.adminNameFormat}"
                 targetSetting="Admin_Account_Naming_Format__c"
@@ -218,7 +220,9 @@
         </div>
         
         <div class="slds-col slds-size--1-of-2">
-            <c:CMP_Acc_Naming_Dropdown isView="{!v.isView}"
+            <c:CMP_Acc_Naming_Dropdown 
+                class="household-account-naming"
+                isView="{!v.isView}"
                 setting="{!v.hierarchySettings.Household_Account_Naming_Format__c}"
                 nameFormat="{!v.hhNameFormat}"
                 targetSetting="Household_Account_Naming_Format__c"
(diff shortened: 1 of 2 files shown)


robertbwatson:
@lifewithryan I'm going to assign this review to @xinnanlisf, as it looks like he did a lot of work on this component in the past.

src/aura/STG_CMP_System/STG_CMP_System.cmp
```diff
@@ -200,7 +200,9 @@
         </div>
         
         <div class="slds-col slds-size--1-of-2" >
-            <c:CMP_Acc_Naming_Dropdown isView="{!v.isView}"
+            <c:CMP_Acc_Naming_Dropdown 
+                class="admin-account-naming"
```
robertbwatson: This is similar to how we include the other CMP_* lightning components in the HEDA Settings.

Without defining this, then there is no need to even have the class attribute in the CMP_Acc_Naming_Dropdown component. I just picked a random name here - it doesn't look like this is used in any of our robot tests yet as the robot tests passed successfully.

lifewithryan: @robertbwatson I think this works. I'm trying to figure out though where the "requirement" for having to have the attribute "class" specified. I see many of our components have it, but don't know where/why it's being forced/accessed.

lifewithryan: @robertbwatson Never mind...I'm blind and seeing it now. It's because every one of those elements in that component are referencing v.class, but we didn't have an attribute there. 

I'm good with this.

robertbwatson: @lifewithryan Yeah it's not "required" but without a value specified, the class ends up being something like "-output-text" instead of "admin-account-naming-output-text"

PR #767, opened 2019-04-09
Contact "Primary Language" Updates Contact Language Record

### LINKED ISSUE #746: Language: Contact "Primary Language" Updates Contact Language Record

robertbwatson:
# Critical Changes

# Changes
## Setting a Primary Contact Language
* When you change the language specified on the "Primary Language" field on a Contact, we also set the Contact Language record that corresponds to the newly selected language to be Primary. If a Contact Language record doesn't exist for that language, one is created and is set to Primary. If a different Contact Language record was previously identified as Primary, we clear the "Primary Language" checkbox on that record.
* When a new Contact Language record is created in this manner, define its "Fluency" value by updating "Default Primary Language Fluency" on the "Accounts and Contacts" tab in the HEDA Settings.
* These updates expand upon the Primary Contact Language functionality we delivered in our last release, 1.74.

# Issues Closed
Closes #746 

# New Metadata
* Class: `CON_PrimaryLanguage_TDTM.cls`
* Class: `CON_PrimaryLanguage_TEST.cls`
* Custom Label: `stgDefaultContactLanguageFluency`
* Custom Label: `stgFluent`
* Custom Label: `stgHelpDefaultContactLanguageFluency`
* New `Hierarchy_Settings__c` field: `Default_Contact_Language_Fluency__c`

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/STG_InstallScript.cls
+++ b/src/classes/STG_InstallScript.cls
@@ -275,7 +275,7 @@ global without sharing class STG_InstallScript implements InstallHandler {
     * @description Populate Default Contact Language Fluency custom setting when upgrading to 1.75
     * @return void
     */
-    private static void setDefaultContactLanguageFluency(Version previousVersion) {
+    global static void setDefaultContactLanguageFluency(Version previousVersion) {
         if(previousVersion != null && previousVersion.compareTo(new Version(1,75)) < 0) {
             Hierarchy_Settings__c orgSettings = UTIL_CustomSettingsFacade.getSettings();
             if(orgSettings.Default_Contact_Language_Fluency__c == null || orgSettings.Default_Contact_Language_Fluency__c == '') {
--- a/src/classes/STG_InstallScript.cls
+++ b/src/classes/STG_InstallScript.cls
@@ -68,6 +68,8 @@ global without sharing class STG_InstallScript implements InstallHandler {
 
             populateAdminRecordTypeSetting(previousVersion);
 
+            setDefaultContactLanguageFluency(previousVersion);
+
             if(previousVersion != null && previousVersion.compareTo(new Version(1,38)) < 0) {
                 // We need to handle a special case to populate the Owned_by_Namespace__c field for HEDA Orgs before 1.38.
                 populateNamespaceHandlers();
@@ -264,6 +266,20 @@ global without sharing class STG_InstallScript implements InstallHandler {
             Hierarchy_Settings__c orgSettings = UTIL_CustomSettingsFacade.getSettings();
             if(orgSettings.Administrative_Account_Record_Type__c == null || orgSettings.Administrative_Account_Record_Type__c == '') {
                 orgSettings.Administrative_Account_Record_Type__c = UTIL_Describe.getAdminAccRecTypeID();
+            }
+            upsert orgSettings;
+        }
+    }
+    
+    /*******************************************************************************************************
+    * @description Populate Default Contact Language Fluency custom setting when upgrading to 1.74
+    * @return void
+    */
+    private static void setDefaultContactLanguageFluency(Version previousVersion) {
+        if(previousVersion != null && previousVersion.compareTo(new Version(1,74)) < 0) {
+            Hierarchy_Settings__c orgSettings = UTIL_CustomSettingsFacade.getSettings();
+            if(orgSettings.Default_Contact_Language_Fluency__c == null || orgSettings.Default_Contact_Language_Fluency__c == '') {
+                orgSettings.Default_Contact_Language_Fluency__c = 'Fluent';
             }
             upsert orgSettings;
         }
--- a/src/classes/STG_InstallScript.cls
+++ b/src/classes/STG_InstallScript.cls
@@ -272,14 +272,14 @@ global without sharing class STG_InstallScript implements InstallHandler {
     }
     
     /*******************************************************************************************************
-    * @description Populate Default Contact Language Fluency custom setting when upgrading to 1.74
+    * @description Populate Default Contact Language Fluency custom setting when upgrading to 1.75
     * @return void
     */
     private static void setDefaultContactLanguageFluency(Version previousVersion) {
-        if(previousVersion != null && previousVersion.compareTo(new Version(1,74)) < 0) {
+        if(previousVersion != null && previousVersion.compareTo(new Version(1,75)) < 0) {
             Hierarchy_Settings__c orgSettings = UTIL_CustomSettingsFacade.getSettings();
             if(orgSettings.Default_Contact_Language_Fluency__c == null || orgSettings.Default_Contact_Language_Fluency__c == '') {
-                orgSettings.Default_Contact_Language_Fluency__c = 'Fluent';
+                orgSettings.Default_Contact_Language_Fluency__c = Label.stgFluent;
             }
             upsert orgSettings;
         }
--- a/src/classes/UTIL_CustomSettingsFacade.cls
+++ b/src/classes/UTIL_CustomSettingsFacade.cls
@@ -113,7 +113,7 @@ public without sharing class UTIL_CustomSettingsFacade {
         hs.Affl_ProgEnroll_Status_Map__c = 'Current';
         hs.Affiliation_Record_Type_Enforced__c = false;
         hs.Administrative_Account_Record_Type__c = UTIL_Describe.getAdminAccRecTypeID();
-
+        hs.Default_Contact_Language_Fluency__c = 'Fluent';
     }
 
     /*******************************************************************************************************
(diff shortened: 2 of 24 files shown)


src/classes/UTIL_CustomSettingsFacade.cls
```diff
@@ -113,7 +113,7 @@ public without sharing class UTIL_CustomSettingsFacade {
         hs.Affl_ProgEnroll_Status_Map__c = 'Current';
         hs.Affiliation_Record_Type_Enforced__c = false;
         hs.Administrative_Account_Record_Type__c = UTIL_Describe.getAdminAccRecTypeID();
-
+        hs.Default_Contact_Language_Fluency__c = 'Fluent';
```
robertbwatson: The only thing I don't like about this is that for new customers/new installs, this will default to Fluent. For current customers, it's a blank value and they must go to the HEDA Settings UI and select a value.

Alternatively, I could set the default value of the custom setting field to "Fluent" right in the object metadata, but it doesn't seem like we've done that in the past so I wasn't sure on the best approach here...

src/labels/CustomLabels.labels
```diff
@@ -871,6 +871,14 @@
         <shortDescription>Setting Course Description Copy Email Subject</shortDescription>
         <value>Course Description Copy Job Completed with a status of:</value>
     </labels>
+    <labels>
```
robertbwatson: These still need to be reviewed by Julie but I will try to run it by her in the next day or two.

src/classes/UTIL_CustomSettingsFacade.cls
lifewithryan: @robertbwatson We did something similar with Admin Account naming. I think our install script handled it, but upon install, if we found that the Admin RecordType wasn't specified, we set it to 'Administrative' then they could refer to it like they are doing on line 115 there.

Could we do something like this: https://github.com/SalesforceFoundation/HEDAP/blob/master/src/classes/STG_InstallScript.cls (line 262). This would set it for existing (because it would be empty since this is new). Just a thought.

lifewithryan: Actually, this is probably fine. I see that we are doing it for other options like 'Current' etc. That said, on install we could still probably add something like this to the Install Script so that it is consistent.

robertbwatson: @lifewithryan  Fantastic suggestion! I forgot about the install script, honestly. I'll look into this.

robertbwatson: @lifewithryan  Done - see STG_InstallScript & STG_InstallScript_TEST below. I'll update the WI to note a few extra tests we should consider with this approach.

src/classes/STG_InstallScript.cls
```diff
@@ -268,4 +270,18 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+    
+    /*******************************************************************************************************
+    * @description Populate Default Contact Language Fluency custom setting when upgrading to 1.74
+    * @return void
+    */
+    private static void setDefaultContactLanguageFluency(Version previousVersion) {
+        if(previousVersion != null && previousVersion.compareTo(new Version(1,74)) < 0) {
+            Hierarchy_Settings__c orgSettings = UTIL_CustomSettingsFacade.getSettings();
+            if(orgSettings.Default_Contact_Language_Fluency__c == null || orgSettings.Default_Contact_Language_Fluency__c == '') {
+                orgSettings.Default_Contact_Language_Fluency__c = 'Fluent';
```
robertbwatson: @lifewithryan I'm wondering if I should be concerned about translations here? Right now, we don't have the picklist translated. But when it's translated, then it would likely need to be a different default value here (& in the UTIL_CustomSettingsFacade class)

Maybe I should just choose the first value in the picklist as the default setting?

src/classes/STG_InstallScript.cls
```diff
@@ -268,4 +270,18 @@ global without sharing class STG_InstallScript implements InstallHandler {
             upsert orgSettings;
         }
     }
+    
+    /*******************************************************************************************************
+    * @description Populate Default Contact Language Fluency custom setting when upgrading to 1.74
+    * @return void
+    */
+    private static void setDefaultContactLanguageFluency(Version previousVersion) {
+        if(previousVersion != null && previousVersion.compareTo(new Version(1,74)) < 0) {
```
robertbwatson: This will definitely need to change if this PR doesn't make it in 1.74!!!

src/classes/STG_InstallScript.cls
```diff
@@ -268,4 +270,18 @@ global without sharing class STG_InstallScript implements InstallHandler {
             upsert orgSettings;
         }
     }
+    
+    /*******************************************************************************************************
+    * @description Populate Default Contact Language Fluency custom setting when upgrading to 1.74
+    * @return void
+    */
+    private static void setDefaultContactLanguageFluency(Version previousVersion) {
```
robertbwatson: The other methods above that perform certain logic on install/upgrade are global but I have no idea why. It doesn't really make sense to me.

src/classes/STG_InstallScript.cls
lifewithryan: @Robert Can we create a custom label for the word "Fluent" and have you grab that? That way it would be translated (I think). AS long as the translations line up we'd be good.

src/classes/STG_InstallScript.cls
lifewithryan: I'm not seeing how we are gonna make it into 1.74 at this point. Time is running out.

robertbwatson: Yep, I'll change it.

src/classes/STG_InstallScript.cls
robertbwatson: @lifewithryan I don't know why I didn't think of that, but yes, a label would be great.

src/labels/CustomLabels.labels
robertbwatson: Currently discussing with @juliesnow in Quip and hope to resolve by EOD or Monday

src/classes/STG_InstallScript.cls
robertbwatson: Done

src/classes/STG_InstallScript.cls
robertbwatson: Done

src/classes/STG_InstallScript.cls
lifewithryan: @robertbwatson -- I'll see if I can find out why. It might have something to do with being able to have the cci installer stuff call into the class and run specific pieces of setup.

robertbwatson:
@juliesnow Since this didn't get out in 1.74, should we update the release notes to mention something like, "This expands on functionality delivered in 1.74."?

robertbwatson:
> @robertbwatson -- for consistency, and I believe so we can reference the install script methods via anonymous apex (and probably via the cci install process) we should change your new install script method to global. I just tried running it manually via the debug console and as I suspected it wasn't visible so it threw an error.
> Everything else seems to be working just fine though. nice work.

@lifewithryan Done! See https://github.com/SalesforceFoundation/HEDAP/pull/767/commits/eb41ea8d3275cc7ffa3b9b94c921101eb3870191

PR #769, opened 2019-04-12
Fix the metadeploy install plan

davisagli:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -304,3 +304,5 @@ plans:
                         name: Course Connection Record Types for HEDA
                     facility_display_name:
                         name: Facility Display Name Formula Field
+                options:
+                    unmanaged: false


cumulusci.yml
```diff
@@ -304,3 +304,5 @@ plans:
                         name: Course Connection Record Types for HEDA
                     facility_display_name:
                         name: Facility Display Name Formula Field
+                options:
+                    unmanaged: false
```
robertbwatson: @davisagli Just curious, what does this option do?

I took a quick look at the new metadeploy installer and realized that it has Facility Display Name Formula Field as a required option. If we wanted to change it, is it easy to make it an optional selection? (not sure if we do -- but wanted to ask...)

davisagli: There are parts of the post-install metadata that refer to objects in the HEDA package, and therefore need to have a namespace prefix when HEDA is installed as a managed package but no namespace prefix when it's unmanaged. This makes sure that CumulusCI will inject the prefix where necessary.

It's now possible to make an installation step optional; that's a possibility we didn't have before. We would probably want to work with the docs team to make sure we're providing a clear description so the user knows why they would want to choose it or not.

davisagli:
@jstvz I updated the most recent installer manually, so no need to republish.

cumulusci.yml
robertbwatson: Thanks for the explanation!

PR #771, opened 2019-04-23
Add a prerelease org config

davisagli:
This makes it possible to get a scratch org on the prerelease version of Salesforce by using `--org prerelease`.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/orgs/prerelease.json
+++ b/orgs/prerelease.json
@@ -1,0 +1,12 @@
+{
+  "orgName": "HEDA - Prerelease Dev Workspace",
+  "edition": "Developer",
+  "instance": "CS46",
+  "orgPreferences" : {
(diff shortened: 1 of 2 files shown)


orgs/prerelease.json
```diff
@@ -0,0 +1,12 @@
+{
+  "orgName": "HEDA - Prerelease Dev Workspace",
```
robertbwatson: @davisagli Not really a big deal, but in #768 I renamed the org configs to "EDA" instead of "HEDA" - can this just be "EDA" too?

davisagli: @robertbwatson good catch, I made that update.

PR #779, opened 2019-04-29
Adding reentrancy checks to the AFFL_AccRecordType_TDTM class.

### LINKED ISSUE #532: Add reentrancy flag to any AFFL_AccRecordType_TDTM class
AFFL_AccRecordType_TDTM currently does not implement the TDTM_ProcessControl.getRecursionFlag design for preventing reentrancy into the class. The code watches for Record Type changes on an Account and then updates any related Affiliations and Contacts (related by one of the Primary_x fields). Adding reentrancy flags to this class can avoid repeat actions on the same objects during the same transactions.

Part of the reentrancy design is to perform the DML within the class rather than letting DMLWrapper handle it...this could could affect DML limits...worth investigating and assessing impact.

robertbwatson:
# Critical Changes

# Changes
* We've updated the AFFL_AccRecordType_TDTM class to prevent reentrancy issues.

# Issues Closed
* Fixes #532 

# New Metadata

# Deleted Metadata

# Testing Notes
1. Create an account with an Educational Institution record type.
2. Create a new contact. Specify the account you just created as the Primary Educational Institution (you may need to drag this field onto the contact page layout).
3. Verify that an affiliation is automatically creating linking the contact to the account.
4. Change the account record type to Business Organization.
5. Verify that the Primary Educational Institution on the contact is now blank. The Primary Business Organization is set to the account instead.


src/classes/AFFL_AccRecordType_TDTM.cls
```diff
@@ -48,6 +51,12 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
         TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
 
     	 DmlWrapper dmlWrapper = new DmlWrapper();
+
+         if (TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.AFFL_AccRecordType_TDTM)) {
+            reentrancyPrevented = true;
```
robertbwatson: I don't necessarily like doing this, but it's the only way I know of to easily test that the reentrancy check works as expected in a unit test.

src/classes/AFFL_AccRecordType_TDTM.cls
```diff
@@ -62,36 +71,47 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
...
+                    } else if (keyAfflFieldNewType != null) {
+                        accsTypeAdded.put(acc.Id, keyAfflFieldNewType);
+                    }
     	 		}
 		    }
+
 		    updatePrimaryAfflsOnChange(accsTypeChanged, dmlWrapper);
             updatePrimaryAfflsOnAdd(accsTypeAdded, dmlWrapper);
+
+            TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.AFFL_AccRecordType_TDTM, true);
```
lifewithryan: This "feels" like a significant change (I know you warned me about it). We would want to get this tested, and any products built on top of EDA, asap. Would you mind writing up something that discusses this implementation so that we can give other's a heads up?

src/classes/AFFL_AccRecordType_TDTM.cls
```diff
@@ -62,36 +71,47 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
...
     	 		if(triggerAction == TDTM_Runnable.Action.AfterUpdate) {
 	    	 		Account acc = (Account)so;
 			        Account oldAcc = (Account)oldList[i];
 
-			        if(acc.RecordTypeId != oldAcc.RecordTypeId) {
-			        	AFFL_MultiRecordTypeMapper afflMapper = new AFFL_MultiRecordTypeMapper();
+                    // If the record type isn't changing, move on
+                    if(acc.RecordTypeId == oldAcc.RecordTypeId) {
+                        continue;
+                    }
```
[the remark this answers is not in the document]
robertbwatson: Honestly I did it here because of our Apex PMD rules - the alert came up re: too many nested if statements. By short circuiting early, you can keep code cleaner because, in many cases, it results in less nesting :)

I'll think about the blog post idea...

src/classes/AFFL_AccRecordType_TDTM.cls
robertbwatson: I'll have to think about this more. I did write up a Quip doc for bulk data loader testing. I think the biggest change is that instead of grouping the affiliation record updates at the end with the other potential record updates, we're doing it here.  The likelihood that account record types are being changed in bulk at the same time as other updates (like address updates) is probably pretty low.

I expect that if anyone is using the bulk api to perform up to 10,000 account record type updates today, they are probably already running into potential CPU limits and DML limits. But maybe we need to call out something around this in the release notes?

robertbwatson: Interesting - we call out some best practices in this article with regards to TDTM and bulk data processing / API users already: https://powerofus.force.com/s/article/HEDA-Disable-Trigger-Handlers

src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -35,7 +35,36 @@
...
+
+        /*******************************************************************************************************
+        * @description Does nothing unless the static List of sObjectsToUpdate is not null.
+        * @param listNew the list of SObjects from trigger new.
+        * @param listOld the list of SObjects from trigger old.
+        * @param triggerAction which trigger event (BeforeInsert, AfterInsert, etc.).
+        * @param objResult the describe for the SObject.
+        * @return dmlWrapper
+        ********************************************************************************************************/
+        public override TDTM_Runnable.DmlWrapper run(List<SObject> newList, List<SObject> oldList, TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
```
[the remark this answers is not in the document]
robertbwatson: @lifewithryan This is all actually the same line, but perhaps it should be two separate lines for readability?

robertbwatson:
> Code looks good. I made a comment or two, but the big take away is with respect to use processing the DML inside the trigger. Feels like it could be an impactful changes, maybe more so to dependent products than to EDA itself. I'd like to pass this concept around a bit to see if anyone on the other teams that depend on us and TDTM think about it.

@lifewithryan  Who would you suggest we reach out to?

src/classes/UTIL_UnitTestData_TEST.cls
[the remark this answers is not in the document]
lifewithryan: @robertbwatson  Oh I see...yea, if we could maybe line up the arguments underneath a bit more to make it more obvious they go together.

src/classes/AFFL_AccRecordType_TDTM.cls
lifewithryan: @robertbwatson that's a good find. I say we go ahead as planned. I tend to agree that flipping the account RTs in bulk along with other triggers is probably an edge case. I think we will be okay. I know you're working on other items like this at the moment as well, do we plan on doing the same sort of thing there?

src/classes/AFFL_AccRecordType_TDTM.cls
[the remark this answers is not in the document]
lifewithryan: I just think it's something not everyone does. There's a stigma around things like continue and break but I think you're using them correctly here. Exactly what continue is for. :)

src/classes/AFFL_AccRecordType_TDTM.cls
robertbwatson: @lifewithryan Yeah, I plan on writing up some sort of document for bulk data loader testing for all of the other WIs as well.

src/classes/UTIL_UnitTestData_TEST.cls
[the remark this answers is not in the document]
robertbwatson: Done!

src/classes/AFFL_AccRecordType_TDTM.cls
dgamboni: @lifewithryan I have been using the Data Import Wizard rather than the DataLoader tool for testing. Pls let me know if that's OK.

dgamboni: I'm getting this error with 2000 records : 

> CANNOT_INSERT_UPDATE_ACTIVATE_ENTITY:hed.TDTM_Account: System.LimitException: Apex CPU time limit exceeded:--

Do I need to disable a particular trigger handler, as referred to in the doc linked above?

lifewithryan: @dgamboni I'd be curious to see if you get the error with Dataloader as well. Import Wizard is a bit different IIRC.

robertbwatson: Not surprising that a CPU timeout is being thrown when working with a large number of records, but I do want to investigate further to make sure that we get about the same results when working with the master branch.

This could spawn a few entirely different work items that tries to improve the efficiency of our code - the fact that only X number of accounts can be inserted/updated at one time without timing out is a bit of a concern to me. Many organizations probably have multiple other packages that also perform logic on account insert/updates so the failure rate would be even higher in a customer org. We want our code to be as fast as possible, so it might be worth spending some time to determine where we might be able to make our code better.

PR #780, opened 2019-04-30
Rename package to EDA

jstvz:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -1,13 +1,13 @@
 project:
-    name: HEDAP
+    name: EDA
     package:
-        name: HEDA
+        name: EDA
         namespace: hed
         api_version: 45.0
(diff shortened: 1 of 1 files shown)


cumulusci.yml
```diff
@@ -1,13 +1,13 @@
 project:
-    name: HEDAP
+    name: EDA
```
davisagli: I think this will break access to existing scratch orgs in the sfdx keychain.

jstvz: I'll notify the team. We can work around this via `force:alias:set`, no?

davisagli: Ah, probably yeah.

PR #687, opened 2019-01-14
Feature/jstvz/reparentable course offerings

jstvz:
# Critical Changes

# Changes
- We've enabled reparentable Course Offerings. This will allow updating the Course field of the Course Offering with a new record. 
# Issues Closed
- Fixes #644
# New Metadata

# Deleted Metadata

# Testing Notes
1. Create Course *x*
2. Create Course Offering *y*
3. Create new Course *z*
4. Reparent Course Offering *x* with Course *z*

PR #786, opened 2019-05-03
Making Account on Term & Term on Course Offering reparentable

### LINKED ISSUE #721: Enhancement: Allow reparenting of other objects/records to parent objects/records
We recently released the ability to reparent a course offering to a course in [HEDA Release 1.70](https://github.com/SalesforceFoundation/HEDAP/releases/tag/rel/1.70). This started as an idea here: https://powerofus.force.com/hub-ideas/1423 and was delivered via work item #644. 

Subsequently, HEDA users would like this reparenting concept to be applicable to all other objects/records where it makes sense, as explained here: https://powerofus.force.com/0D51E00004kTUmt

Other reparenting on other objects is possible - it's really only an issue when the link is a Master-Detail relationship, which is the case with Course Offering to Term and Course, for example

Only allows reparenting an orphaned Course Offering record with a new or different Course that was not it's original parent. 

The ability to reparent any child term with a parent term (or any object) where it makes sense. 
* We will allow you to reparent the Term on a Course Offering.
* We will allow you to reparent the Account on a Term.

An upcoming HEDA release will apparently allow reparenting of CourseOfferings to new Course objects.  That's great, but in fact when that happens the Term may also need to be changed.  That is, the Term lookup may need to allow reparenting as well.

* **Acceptance Criteria**
  * On an existing Course Offering, you can now change the Term to another record.
  * On an existing Term, you can now change the Account to another record.

robertbwatson:
# Critical Changes

# Changes
## General Updates
* We've enabled a reparentable field on Terms and additional fields on Course Offerings. Now, you can change the records specified for the Term field on Course Offering and for the Account field on Term.

* If you leave the date fields blank when you create a new Course Offering, EDA will pull the Start Date and End Date from the Term. However, if you change the Term for an existing Course Offering, the Course Offering's Start and End Dates won't be updated this way.

# Issues Closed
Closes #721 

# New Metadata

# Deleted Metadata

# Testing Notes
1. Create a new Term for an existing Account.
2. Change the Account on the Term to another record.
3. Create a new Course Offering for an existing Course and Term.
4. Change the Term on the Course Offering to another record.


robertbwatson:
This is similar to the PR #687

robertbwatson:
@juliesnow  Just wanted to ping you quick on this since you were out last week - I've added you as a reviewer here for the release notes.

PR #798, opened 2019-05-17
Replacing Readme image

ceiroa:
Replacing image at the top of the Readme file. 

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


README.md
```diff
@@ -1,4 +1,4 @@
-![Salesforce Education Data Architecture](https://cloud.githubusercontent.com/assets/1894238/13441500/13d366ec-dfbd-11e5-9df6-d9dee50ce87d.jpg "Salesforce Education Data Architecture")
+![Salesforce Education Data Architecture](https://github.com/SalesforceFoundation/EDA/blob/feature/eda-image/EDA%20GitHub.png "Salesforce Education Data Architecture")
```
lifewithryan: @ceiroa do we know why the two domains in the URL are different? Does it matter?

ceiroa: I think the old URL was pointing to an image hosted in Bromer's user files. To avoid having it point to a specific user's file I have uploaded the image directly to the repo and pointed to that.

PR #810, opened 2019-05-28
Prevent nested plan requirement having lookup to program plan

### LINKED ISSUE #812: MCMP Prevent Nested Plan Reqs from linking to Parent Plan
Update 5/29/2019:
MCMP wants to ensure the Plan Requirements that descend from a given Program Plan can't be associated with a different Program Plan from its parent/grandparent(s). 

Analysis

RE: MCMP ticket -- off the top of my head:

1. Create Program Plan (via the UI)
2. Create Program Plan Requirement (via the UI)
3. Select Program Plan for the requirement (via the UI) *(Optional)*
4. CODE: Since this is the first plan requirement, its Plan_Requirement__c field should be blank, so exit, there's nothing for us to enforce.
5. Now create a CHILD requirement from the first plan requirement. (via the UI)
6.  This is a CHILD of the first Plan Requirement (its Plan_Requirement__c field is not blank)
7. 5/21 Update: CODE: Since this is a CHILD of the first Plan Requirement, Program Plan must not be populated if the new setting is enabled.

Outdated:

Description: 
Regardless of where a "Plan Requirement" falls (either parent requirement or a child requirement in a thread of requirements), at the end of the day should only roll up to ONE "Program Plan". This creates less confusion in the long-run and makes the use-case of the deep clone more valid. This is cleaner to handle architecturally and makes the Deep Clone feature more useful/efficient. 

This work in this ticket should make sure an Admin cannot affiliate the same Plan Requirement record (ex: Biology 101a) to both the "Bachelors in Biology" program plan and the "Biomedical Engineering" program plan. Each program plan would have it's own. If they attempt to affiliate the same plan requirement with 2 different program plans, it would be nice if they received an error message along the lines of "You cannot affiliate the same plan requirement with 2 different Program Plans, would you like to clone this requirement?" (this latter part is a nice-to-have, not a must-have).

Acceptance Criteria:
1.) As an Admin you attempt to take an existing Plan Requirement and affiliate it with 2 (or more) different Program Plans. Should not be allowed, receive clean error message.
2.) As an Admin you attempt to create a new Plan Requirement and affiliate it with 2 (or more) Program Plans. Should not be allowed, receive clean error message.

xinnanlisf:
# Critical Changes

# Changes
- We've included a custom setting that you can use to ensure that a child Plan Requirement can't look up to a Program Plan. This prevents users from incorrectly including Plan Requirements in a Program Plan that is not the Program Plan used by the parent requirement.
For example, if a Program Plan called B.S. in Biology 2018/2019 has a parent Plan Requirement of Organic Chemistry with child Plan Requirements for Organic Chemistry I and Organic Chemistry II, we wouldn't want either of those child requirements looking up to a Program Plan for B.S. in Forensic Pathology.
This new setting is turned on by default for new customers and is disabled for existing customers. You can access the new setting from under Validate Program Plan for Nested Plan Requirements in EDA Settings | Program Plans | Settings.
# Issues Closed
Fixes #812
# New Metadata
PREQ_PreventPPlanParent_TDTM.cls
PREQ_PreventPPlanParent_TEST
Hierarchy_Settings__c.Validate_Program_Plan_for_Nested_PR__c
STG_CMP_ProgramPlans
# Deleted Metadata

# Testing Notes
Turn on custom setting (Validate Program Plan for Nested Plan Requirements) under EDA setting page -> Program Plan
1. Create a program plan: Test Program Plan 1
2. Create a plan requirement: Test Plan Requirement 1 and populate the program plan field with Test Program Plan 1
3. Go to the related list of Test Plan Requirement 1 and create a child plan requirement: Test Child Plan Requirement 1. Try to fill in the program plan field of the child plan requirement and click save
4. Make sure the record is not saved and the error message is shown
5. Find an existing plan requirement and try to fill in the program plan field. The result should be the same.


xinnanlisf:
@lifewithryan this PR is ready for review now!

src/aura/STG_CMP_ProgramPlans/STG_CMP_ProgramPlans.cmp
```diff
@@ -0,0 +1,33 @@
...
+            <ui:outputText value="{!$Label.c.stgNestedPlanRequirementPP}" />
+        </div>
+        <div class="slds-col slds-size--1-of-2">
+            <div class="slds-form-element">
+                <div class="slds-form-element__control">
+                    <label class="slds-checkbox">
+                        <aura:if isTrue="{!v.isView}">
+                            <ui:outputCheckbox value="{!v.hierarchySettings.Validate_Program_Plan_for_Nested_PR__c}" class="unique-program-plan" />
+                            <aura:set attribute="else">
+                                <ui:inputCheckbox aura:id="storeErrors" value="{!v.hierarchySettings.Validate_Program_Plan_for_Nested_PR__c}"
```
lifewithryan: @xinnanlisf just a nit-picky thing here. Why is the aura:id and class "store-errors" here?

src/objects/Hierarchy_Settings__c.object
```diff
@@ -390,6 +390,16 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>Validate_Program_Plan_for_Nested_PR__c</fullName>
+        <defaultValue>true</defaultValue>
+        <description>When selected, Plan Requirements can specify either a Program Plan or a parent Plan Requirement, but not both. For example, a Plan Requirement that has a parent Plan Requirement can’t also specify a Program Plan, or vice versa. If you specify both a Program Plan and a parent Plan Requirement, validation fails and the Plan Requirement isn’t saved.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>When selected, Plan Requirements can specify either a Program Plan or a parent Plan Requirement, but not both. For example, a Plan Requirement that has a parent Plan Requirement can’t also specify a Program Plan, or vice versa. If you specify both a Progr</inlineHelpText>
```
lifewithryan: @xinnanlisf the inline help here is getting cut off. @juliesnow do we want to do anything about this?

src/classes/PREQ_PreventPPlanParent_TDTM.cls
```diff
@@ -0,0 +1,72 @@
...
+    ********************************************************************************************************/
+    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist,
+            TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+
+        if (triggerAction == TDTM_Runnable.Action.AfterInsert
+            || triggerAction == TDTM_Runnable.Action.AfterUpdate) {
+            for (Plan_Requirement__c planReq : [SELECT ID,
+                                                    Program_Plan__c,
+                                                    Plan_Requirement__c,
+                                                    Plan_Requirement__r.Program_Plan__c
```
lifewithryan: @xinnanlisf do we need to retrieve this field any longer? I think we only need Program Plan and Plan Requirement since those are the only two fields we are checking now.

robertbwatson:
@xinnanlisf  @lifewithryan I added a few comments to the original PR that I think got lost when it was closed - working on adding them back here now.

@xinnanlisf Does this resolve a Github issue #?

src/aura/STG_CMP_Tabs/STG_CMP_Tabs.cmp
```diff
@@ -76,6 +76,9 @@
             <li aura:id="courseConTab" id="courseConTab" class="slds-tabs__item slds-text-heading--label" title="Course Connections" role="tab">
                 <ui:outputURL aura:id="courseConTabBtn" value="#" label="{!$Label.c.stgTabCourseConn}" click="{!c.tabNavigationClick}" class="courseConTab course-con-menu-item" />
             </li>
+            <li aura:id="programPlanTab" id="programPlanTab" class="slds-tabs__item slds-text-heading--label" title="Program Plan" role="tab">
```
robertbwatson: @xinnanlisf @lifewithryan I don't think we need to edit this now (as our other tabs are the same way), but I wonder if we should switch to using labels for these tab titles in a future tech debt WI. I expect this is perhaps used by screenreaders and such (I might be wrong though)

src/classes/PREQ_PreventPPlanParent_TEST.cls
```diff
@@ -0,0 +1,144 @@
...
+* @date 2019
+* @group Plan Requirement
+* @group-content ../../ApexDocContent/PlanRequirements.htm
+* @description Tests for PREQ_PreventPPlanParent_TDTM.
+*/
+@isTest
+private class PREQ_PreventPPlanParent_TEST {
+
+    @isTest
+    public static void pReqInsertChildPRWithSamePP() {
```
robertbwatson: @xinnanlisf I think we should add ApexDocs to all of the unit tests so we know exactly what each is testing.

robertbwatson: @xinnanlisf I also think we might want to consider adding a few additional unit tests to test additional configurations:
- A unit test that has valid data (to confirm that no exception is thrown)
- A unit test that has invalid data, but the custom setting is disabled (to confirm that no exception is thrown)

src/objects/Hierarchy_Settings__c.object
```diff
@@ -390,6 +390,16 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>Validate_Program_Plan_for_Nested_PR__c</fullName>
+        <defaultValue>true</defaultValue>
```
robertbwatson: @xinnanlisf @lifewithryan  Did we decide that this should be disabled by default for all current customers, but enabled for new installs? If so, I think we need additional changes here. Either this should be defaulted to false and set to true on new installs, or set to true by default and set to false when upgrading from 1.77 to 1.78... thoughts? (Or maybe this works as expected?)

src/aura/STG_CMP_Tabs/STG_CMP_Tabs.cmp
lifewithryan: @robertbwatson I like that idea. We should create a separate WI under the Tech Debt Theme for this.

src/classes/PREQ_PreventPPlanParent_TEST.cls
lifewithryan: That's probably a good idea. We should get in the habit of doing this for all of our functionality. Testing good and bad data with settings toggled on/off.

xinnanlisf: I am hesitating about the first one. Seems a bit overtesting to me. We might end up having duplicate test methods eventually.
The second one sounds good to me!

src/objects/Hierarchy_Settings__c.object
xinnanlisf: If I remember it correctly, upgrading EDA won't populate the custom setting value for existing users.

src/aura/STG_CMP_Tabs/STG_CMP_Tabs.cmp
xinnanlisf: We might have an item already. Let me see if I can find it. fyi @aligeez

src/classes/PREQ_PreventPPlanParent_TEST.cls
xinnanlisf: You mean adding comments?

src/objects/Hierarchy_Settings__c.object
lifewithryan: I thought that was only on existing fields; don't brand new fields come over with a value? Since its a checkbox, I thought we had an issue where it is neither true nor false until it's toggled. (I could be wrong though).
I will ask since I clearly can't remeber :)

src/classes/PREQ_PreventPPlanParent_TEST.cls
lifewithryan: When it comes to testing, some would say that you can't "over test" -- that said, it does complete the happy path of things working as expected. I'd say if you have time, to add it. I am curious though what you mean when you talk about having duplicate test methods though. Care to elaborate a bit?

robertbwatson: @xinnanlisf  Yep, I think we talked about that during a recent agile processes meeting. It will help us with our current effort in understanding what we have for unit tests vs automated regression vs manual testing

src/objects/Hierarchy_Settings__c.object
robertbwatson: I can't quite remember either - I'm leaning towards @xinnanlisf being correct. We may need to wait for the beta to know for certain. Either way, it should definitely be part of the test plan.

src/classes/PREQ_PreventPPlanParent_TEST.cls
```diff
@@ -0,0 +1,207 @@
...
+        List<Plan_Requirement__c> assertPReqs = [select ID from Plan_Requirement__c];
+        System.assertEquals(1, assertPReqs.size());
+
+        System.assertEquals(false, results.success);
+        System.assertEquals(Label.nestedPlanRequirementPPExp, results.errors[0].message);
+    }
+
+    //Make sure an error is thrown when creating a Plan Requirement who is a child of another plan requirement,
+    //linking to a different program plan and the setting of Validate_Program_Plan_for_Nested_PR__c is on
+    public static void pReqInsertChildPRWithDifferentPP() {
```
robertbwatson: @xinnanlisf I just noticed this isn't marked as isTest, so it's not being run as a unit test.

xinnanlisf: huh, good catch!

src/classes/PREQ_PreventPPlanParent_TEST.cls
xinnanlisf: @lifewithryan yeah I will add it to unit tests then.

I feel like I confuse myself a bit here :). My initial thought was: The first test method Bobby purposed is in fact testing if a nested plan requirement can have look up to a program plan. (Or, if a plan requirement can have both plan requirement and program plan lookup populated) Isn't it testing against core functionality?

src/objects/Hierarchy_Settings__c.object
xinnanlisf: Just confirmed in my own packaging test org. Checkbox will be unchecked when upgrading even the default value is "checked".

xinnanlisf:
@lifewithryan should be back to ready for review now! Let me know if I missed anything.

src/objects/Hierarchy_Settings__c.object
juliesnow: Yeah, @lifewithryan - Xinnan will remove "For example, a Plan Requirement that has a parent Plan Requirement can’t also specify a Program Plan, or vice versa," which pulls it well below the 255 character limit.

src/classes/PREQ_PreventPPlanParent_TEST.cls
```diff
@@ -0,0 +1,239 @@
...
+        prs[0].Program_Plan__c = pPlans[0].Id;
+        insert prs[0];
+
+        Test.startTest();
+        prs[1].Plan_Requirement__c = prs[0].Id;
+        prs[1].Program_Plan__c = pPlans[0].Id;
+        Database.saveResult results = Database.insert(prs[1], false);
+        Test.stopTest();
+
+        //Verify new plan requirement is not saved
```
robertbwatson: @xinnanlisf Super minor, but we're verifying that it IS saved :)

src/classes/PREQ_PreventPPlanParent_TEST.cls
```diff
@@ -0,0 +1,239 @@
...
+        insert prs[0];
+        prs[1].Plan_Requirement__c = prs[0].Id;
+        insert prs[1];
+
+        Test.startTest();
+        prs[1].Program_Plan__c = pPlans[1].Id;
+        Database.saveResult results = Database.update(prs[1], false);
+        Test.stopTest();
+
+        //Verify new plan requirement is not saved
```
robertbwatson: @xinnanlisf  Same here - we're verifying that it is saved

src/classes/PREQ_PreventPPlanParent_TEST.cls
```diff
@@ -0,0 +1,239 @@
...
+        List<Plan_Requirement__c> prs = UTIL_UnitTestData_TEST.getMultipleTestPlanRequirements(2);
+        prs[0].Program_Plan__c = pPlans[0].Id;
+        insert prs[0];
+
+        Test.startTest();
+        prs[1].Program_Plan__c = pPlans[1].Id;
+        Database.saveResult results = Database.insert(prs[1], false);
+        Test.stopTest();
+
+        //Verify new plan requirement is not saved
```
robertbwatson: @xinnanlisf  and one final one :)

src/classes/PREQ_PreventPPlanParent_TEST.cls
```diff
@@ -0,0 +1,239 @@
...
+    //Make sure no error is thrown when inserting a Plan Requirement who is not a child of another plan requirement,
+    //linking to program plan and the setting of Validate_Program_Plan_for_Nested_PR__c is on
+    @isTest
+    public static void pReqInsertNonChildPRWithPPSettingOn() {
+        UTIL_CustomSettingsFacade.getSettingsForTests(new Hierarchy_Settings__c(Validate_Program_Plan_for_Nested_PR__c = true));
+
+        List<Program_Plan__c> pPlans = UTIL_UnitTestData_TEST.getMultipleTestProgramPlans(2);
+        insert pPlans;
+
+        List<Plan_Requirement__c> prs = UTIL_UnitTestData_TEST.getMultipleTestPlanRequirements(2);
```
robertbwatson: I'm not entirely sure how lines 221-223 are different than the other test block.

src/objects/Hierarchy_Settings__c.object
robertbwatson: @xinnanlisf Perfect!!!

src/classes/PREQ_PreventPPlanParent_TEST.cls
lifewithryan: Yes it's testing against core functionality simply because we're just trying to save the record and not throw an exception. It seems simple enough and this would likely never fail, but because we are adding some logic around it, its more of a "cover our butts" kind of thing. If we ever see that fail we know we really screwed something up. I feel like you can't over test, some would argue but I say "nay nay" -- I'm fine with such tests.

src/classes/PREQ_PreventPPlanParent_TEST.cls
```diff
@@ -0,0 +1,237 @@
...
+
+        Test.startTest();
+        prs[0].Program_Plan__c = pPlans[0].Id;
+        Database.saveResult results = Database.insert(prs[0], false);
+        Test.stopTest();
+
+        //Verify new plan requirement is saved
+        List<Plan_Requirement__c> assertPReqs = [SELECT ID,Program_Plan__c
+                                                FROM Plan_Requirement__c
+                                                WHERE ID = :prs[1].Id
```
lifewithryan: I think we want pr[0] here instead of pr[1]

src/classes/ERR_ExceptionHandler.cls
```diff
@@ -112,4 +113,5 @@ public with sharing class ERR_ExceptionHandler {
 
     public class AfflAccountNoRecordTypeException extends Exception {}
     public class AffAccountRecordTypeMappingsException extends Exception {}
+    public class nestedPlanRequirementPPException extends Exception {}
```
robertbwatson: @xinnanlisf @lifewithryan  This might be me being super nit-picky, and I know we want to move this PR forward, but shouldn't class names begin with a capital letter?

lifewithryan: Yes, it should be. I'll do a run through and change it where it needs to be changed since Xinnan is on PTO.

lifewithryan: done

src/classes/PREQ_PreventPPlanParent_TDTM.cls
```diff
@@ -0,0 +1,71 @@
...
+            for (Plan_Requirement__c planReq : [SELECT ID,
+                                                    Program_Plan__c,
+                                                    Plan_Requirement__c
+                                                FROM Plan_Requirement__c
+                                                WHERE ID in :newlist]) {
+                if (stgUniqueProgramPlanEnabled == true
+                        && planReq.Plan_Requirement__c != null
+                        && planReq.Program_Plan__c != null) {
+                    //throw an error
+                    throw new ERR_ExceptionHandler.NestedPlanRequirementPPException(Label.nestedPlanRequirementPPExp);
```
robertbwatson: @xinnanlisf @lifewithryan I think we're getting exception emails here because we're throwing an exception. I think if we add an error like planReq.addError(Label.whateverTheLabelIs), it should be handled differently.

cc @ceiroa

xinnanlisf: @robertbwatson 
I don't think this would work since planReq is not available in trigger context.

xinnanlisf: @lifewithryan too

robertbwatson: @xinnanlisf Actually, I think I missed this but why are we querying the Plan Requirement? We should be able to just loop over the newlist since we're only accessing fields on the Plan Requirement.

robertbwatson: I think we had to query the object before when we were checking for proper nesting, but it seems like we can get away with not needing to query here.

xinnanlisf: @robertbwatson that's a fair point.

xinnanlisf: For future reference, we can even map it to the old list. e.g. Contact contactInContext = (Contact)oldmap.get(c.ID)

robertbwatson: @xinnanlisf  @lifewithryan Depending what we decide here may also impact my PR on Time Fields, as I chose to use the addError() method: https://github.com/SalesforceFoundation/EDA/pull/837/files#diff-c9996d678f52cbc57cee482c01f4d94cR140

robertbwatson:
> looks good here

@lifewithryan  @xinnanlisf  I agree! I think we should have Dina run through this again - at least the basic test cases.

Issue #781, opened 2019-04-30
Duplicate Affiliation records are created when an Account record type is changed


robertbwatson:
* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the current behavior?**
If we update an Account record type from a record type that has Affiliation mapping to a record type that also has Affiliation mapping and this Account has an Affiliation record with a Contact, then the system will create a duplicate Affiliation record.

Example: 
1. Create a Contact
2. Create an Account with Education Institution record type
3. Create a Primary Affiliation between Account and Contact
4. Update Account record type to Business Organization.
5. Duplicate Affiliation records now exist - one as the Primary.

* **What is the expected behavior?**
Duplicate Affiliation records should not be created.

* **What is the motivation / use case for changing the behavior?**
This bug can cause storage issues and data issues in an organization, especially if an Account's record type is changed and that Account has an Affiliation to hundreds/thousands of Contact records.

robertbwatson:
**lurch: attach W-021767

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x800000097R0mAAE" target="_blank">W-021767</a>

jacebryan:
Issue identified is no longer reproducible. Closing Issue.

dgamboni:
Work Item was missing the step to check the "primary" box when creating the affiliation. Work Item has updated steps in Chatter and has been re-opened.

mrbelvedere:
Included in beta release 1.78 (Beta 6)

mrbelvedere:
Included in production release 1.78

PR #811, opened 2019-05-29
Added reentrancy flag check to CON_PrimaryAffls_TDTM to prevent running during account record type updates

skosgi1:
# Critical Changes

# Changes

 - We've added a flag to prevent duplicate Affiliations from being created when the record type of a primary affiliated account is changed. 

# Issues Closed
Fixes   <a href="https://github.com/SalesforceFoundation/EDA/issues/781/">781</a>
# New Metadata

# Deleted Metadata

# Testing Notes


robertbwatson:
@skosgi1 Can you edit the Issues Closed section to say this Fixes #781? Tying the PR back to the original Github issue will mean that the issue gets automatically closed once we merge this PR.

src/classes/AFFL_AccRecordType_TDTM.cls
```diff
@@ -156,13 +157,15 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
...
-            if(contactIDs.size() > 0 && accountIDs.size() > 0) {
+            if( contactIDs.size() > 0 && accountIDs.size() > 0) {
                 List<Affiliation__c> afflsMadePrimary = [select Contact__c, Account__c, Affiliation_Type__c
                           from Affiliation__c where Contact__c IN :contactIDs and Account__c IN :accountIDs];
                 AFFL_MultiRecordType_TDTM afflMulti = new AFFL_MultiRecordType_TDTM();
                 afflMulti.uncheckOtherPrimariesSameType(afflsMadePrimary, dmlWrapper);
             }
         }
+
+        TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.CON_PrimaryAffls_TDTM_After_Update, true);
```
robertbwatson: @skosgi1 @lifewithryan  This might fix the bug, but it's not clear to me why. I'm a little concerned that we're setting another classes's recursion flag here, as we're in the AFFL_AccRecordType_TDTM class, not the CON_PrimaryAffls_TDTM class.

Is the bug happening in the CON_PrimaryAffls_TDTM class as a result of that trigger logic being called when this logic here executes? Maybe we need logic in the CON_PrimaryAffls_TDTM class to detect duplicates somehow?

src/classes/AFFL_AccRecordType_TDTM.cls
```diff
@@ -70,17 +70,17 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
...
          //         --> the key affl lookup field for that type in Contacts affiliated to it needs to be populated
-    	 if(newlist != null && newlist.size() > 0 && oldlist != null && oldlist.size() > 0) {
+    	 if( newlist != null && newlist.size() > 0 && oldlist != null && oldlist.size() > 0) {
     	 	
-             for (Integer i = 0; i < newlist.size(); i++) {
+             for ( Integer i = 0; i < newlist.size(); i++) {
     	 	    SObject so = newlist[i];
     	 	    //AFTER UPDATE
-    	 		if(triggerAction == TDTM_Runnable.Action.AfterUpdate) {
+    	 		if( triggerAction == TDTM_Runnable.Action.AfterUpdate) {
```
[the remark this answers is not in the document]
skosgi1: @lifewithryan Looks like it is just github. It looks good on the editor.

src/classes/AFFL_AccRecordType_TDTM.cls
robertbwatson: @skosgi1 @lifewithryan I'd also like to see either a new unit test that tests that duplicate primary affiliation records aren't created when an account record type is updated (or perhaps an existing test). The fact that our unit tests didn't catch this bug to begin with is a pretty good indication to me that we should have one.

src/classes/AFFL_AccRecordType_TDTM.cls
```diff
@@ -52,7 +52,7 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
 
     	 DmlWrapper dmlWrapper = new DmlWrapper();
 
-         if (TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.AFFL_AccRecordType_TDTM)) {
+         if ( TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.AFFL_AccRecordType_TDTM)) {
```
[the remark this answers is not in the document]
skosgi1: @lifewithryan  Fixed the indentation. Can you please give it a second review? Thanks.

robertbwatson:
> @skosgi1 Can you edit the Issues Closed section to say this Fixes #781? Tying the PR back to the original Github issue will mean that the issue gets automatically closed once we merge this PR.

Future reference: https://help.github.com/en/articles/closing-issues-using-keywords

src/classes/AFFL_AccRecordType_TDTM.cls
skosgi1: @robertbwatson @lifewithryan  When the record type on the account changes the AFFL_AccRecordType_TDTM class updates the primary affiliation field on the Contact Object which would then fire CON_PrimaryAffls_TDTM to create another affiliation. So I am trying to stop firing CON_PrimaryAffls_TDTM trigger logic at the end of AFFL_AccRecordType_TDTM execution. Please let me know if the solution sounds ok.

skosgi1: @robertbwatson On it!

robertbwatson: I think I might need to pull this down and do some testing of it tomorrow or Friday. The CON_PrimaryAffls_TDTM trigger has a lot of logic so I'd be concerned we're skipping it when we shouldn't necessarily be always skipping it.

I think there are other scenarios we might need to test here too, for example if we change an account's record type to a record type that isn't mapped to a Primary field on the Contact.

skosgi1: @robertbwatson @lifewithryan Added a new test method for duplicate check.

lifewithryan: Yea, let's deep dive into this one a bit. It seems like we shouldn't be turning off that trigger from within this one. If turning off that trigger works, then maybe there's logic in there that's to blame. If we have the re-entrancy code in the CON_PrimaryAffls_TDTM class already, then I would think that should kick in for us when it needs to.

src/classes/AFFL_AccRecordType_TEST.cls
```diff
@@ -331,4 +331,42 @@ public with sharing class AFFL_AccRecordType_TEST {
         system.assertEquals(hh.Id, markUpdated.Primary_Household__c);
         system.assertEquals(null, markUpdated.Primary_Organization__c);
     }
+
+    @isTest
+    public static void duplicateAffiliationCheck() {
```
robertbwatson: @skosgi1 This is a great unit test - I think you might also be able to update the existing changePrimaryAffiliationRecordType() unit test on line 54 to check that duplicates aren't being created there as well, as the test is very similar.

src/classes/AFFL_AccRecordType_TEST.cls
```diff
@@ -331,4 +331,42 @@ public with sharing class AFFL_AccRecordType_TEST {
...
+
+        // Insert a business account
+        Account bizAcc = UTIL_UnitTestData_API.getMultipleTestAccounts(1, orgRecTypeID)[0];
+        insert bizAcc;
+
+        // Insert a contact with business account as Primary business organization
+        Contact contact = UTIL_UnitTestData_API.getContact();
+        insert contact;
+        
+        //Verify default household account has been created for the contact
```
robertbwatson: @skosgi1 This comment isn't really valid and can be removed, I think -- we're inserting a contact and then setting the account as the Primary business organization (the comment on line 343 applies to what we're doing here)

src/classes/AFFL_AccRecordType_TEST.cls
```diff
@@ -331,4 +331,42 @@ public with sharing class AFFL_AccRecordType_TEST {
...
+        contact = [SELECT AccountId,  Primary_Organization__c FROM Contact WHERE Id = :contact.Id];
+        contact.Primary_Organization__c = bizAcc.id;
+        update contact;
+        
+        //Verify 1 affiliation is created
+        List<Affiliation__c> affiliationList = new List<Affiliation__c>();
```
robertbwatson: @skosgi1 Minor comment, but lines 353 and 354 could be combined to just `List<Affiliation__c> affiliationList = [SELECT ...];`

src/classes/AFFL_AccRecordType_TEST.cls
```diff
@@ -331,4 +331,42 @@ public with sharing class AFFL_AccRecordType_TEST {
...
+        Test.startTest();
+
+        Account acc = [select id, RecordTypeId from account WHERE id = :bizAcc.Id];
+        acc.RecordTypeId = householdRecTypeID;
+        update acc;
+
+        Test.stopTest();
+
+        //Verify no duplicate affiliations are created
+        List<Affiliation__c> allAffiliationList = new List<Affiliation__c>();
```
robertbwatson: Same here - we could just do `affiliationList = [SELECT ...];`

skosgi1:
@robertbwatson Thanks for the suggestions. I have updated the code accordingly.

src/classes/AFFL_AccRecordType_TDTM.cls
```diff
@@ -70,17 +70,17 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
          //
          //    B. If that account did not have a record type:
          //         --> the key affl lookup field for that type in Contacts affiliated to it needs to be populated
-    	 if(newlist != null && newlist.size() > 0 && oldlist != null && oldlist.size() > 0) {
```
robertbwatson: Nothing really changes in this class at the moment, except for minor formatting updates.

src/classes/CON_PrimaryAffls_TDTM.cls
```diff
@@ -84,18 +84,22 @@ public class CON_PrimaryAffls_TDTM extends TDTM_Runnable {
...
 
-        // BEFORE UPDATE
-        if ( newlist != null && triggerAction == TDTM_Runnable.Action.AfterUpdate && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.CON_PrimaryAffls_TDTM_After_Update)) {
-            for (integer i=0; i<newlist.size(); i++) {
+        // AFTER UPDATE
+        // Do not run if the reentrant flag is set OR if the contacts are being updated as a result of the AFFL_AccRecordType_TDTM trigger, as
+        // that TDTM class already handles the Affiliation record updates.
+        if (newlist != null && triggerAction == TDTM_Runnable.Action.AfterUpdate 
+                && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.CON_PrimaryAffls_TDTM_After_Update)
+                && !TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.AFFL_AccRecordType_TDTM)) {
```
robertbwatson: @skosgi1 @lifewithryan I think this might be the better fix - still not sure if it's the "correct" fix but I think it's safe to ignore the logic here if we're in the middle of updating contacts & affiliation records as a result of an account record type change, since the updates are all handled in that class instead.

src/classes/CON_PrimaryAffls_TDTM.cls
```diff
@@ -105,21 +109,21 @@ public class CON_PrimaryAffls_TDTM extends TDTM_Runnable {
                         //         was his intention)
                         //  --> Make the old Affiliation non-primary. Although this could be AFFL_MultiRecordType_TDTM's responsibility, since the Affl type
                         //        is not yet defined (and the formula will run later), we'll do it here.
-                        else if(newContact.get(afflMapper.primaryAfflFieldNames[j]) != null && newContact.get(afflMapper.primaryAfflFieldNames[j])
-                        != oldContact.get(afflMapper.primaryAfflFieldNames[j])) {
+                        else if (newContact.get(afflMapper.primaryAfflFieldNames[j]) != null && newContact.get(afflMapper.primaryAfflFieldNames[j])
```
lifewithryan: @robertbwatson while you're in there, care to cuddle the else statements? they need love. or are they that way because of that comment? If so can we move the comments inside the else? I'm just having a really hard time reading that.

src/classes/CON_PrimaryAffls_TDTM.cls
lifewithryan: We just have to watch out for side effects here that we may not be aware of, but seems like it'd be okay.

lifewithryan:
I will say this PR could still use some formatting love with some of the legacy stuff but I don't want to hold anything up as they're not blatantly bad, just my nit pickiness :)

src/classes/CON_PrimaryAffls_TDTM.cls
robertbwatson: @lifewithryan Yep, I'll make another quick pass at the formatting (cc @skosgi1 )

robertbwatson: Done

robertbwatson:
@juliesnow Just pinging you here to let you know you have release notes to review on this one.

dgamboni:
@juliesnow Do we want to let users know in the release notes about the issue uncovered in testing? (see : https://salesforce.quip.com/s3imAOxbOlgv) 

For example, _"We've added a flag to prevent duplicate Affiliations from being created on the Contact record when the record type of a primary affiliated account is changed. If the Account's record type is changed to one that already has an Account listed in a Contact record's Primary Affiliation field, this Account will replace the existing one."_

robertbwatson:
> @juliesnow Do we want to let users know in the release notes about the issue uncovered in testing? (see : https://salesforce.quip.com/s3imAOxbOlgv)
> 
> For example, _"We've added a flag to prevent duplicate Affiliations from being created on the Contact record when the record type of a primary affiliated account is changed. If the Account's record type is changed to one that already has an Account listed in a Contact record's Primary Affiliation field, this Account will replace the existing one."_

@dgamboni @juliesnow My only feedback here is that this is how it has always worked. So we need to make sure we're clear that we didn't add this functionality - if the Account's record type is changed to one that already has an Account listed in a Contact record's Primary Affiliation field, it has always done this.

dgamboni:
> > @juliesnow Do we want to let users know in the release notes about the issue uncovered in testing? (see : https://salesforce.quip.com/s3imAOxbOlgv)
> > For example, _"We've added a flag to prevent duplicate Affiliations from being created on the Contact record when the record type of a primary affiliated account is changed. If the Account's record type is changed to one that already has an Account listed in a Contact record's Primary Affiliation field, this Account will replace the existing one."_
> 
> @dgamboni @juliesnow My only feedback here is that this is how it has always worked. So we need to make sure we're clear that we didn't add this functionality - if the Account's record type is changed to one that already has an Account listed in a Contact record's Primary Affiliation field, it has always done this.

Good point, @robertbwatson! Mostly I would hate for a Customer to change a Record Type on an Account and suddenly have a lot of Primary Affiliations get updated without the Customer being aware that could happen. However, since this already exists, perhaps Release Notes isn't the best place to call this out, then, but just the docs?

PR #813, opened 2019-05-30
Adding reentrancy flag to CCON_Faculty_TDTM & improved COFF_Affiliation_TDTM efficiency

### LINKED ISSUE #534: Add reentrancy flags to CCON_Faculty_TDTM
The After Insert/Update events make changes to the related Course Offering (setting primary faculty field) which has the potential to make updates back to the Course Enrollment table. A reentrancy flag that controls access to the "courseOfferingsToUpdate" logic would save some additional processing

robertbwatson:
# Critical Changes

# Changes

* We've updated the CCON_Faculty_TDTM class to prevent reentrancy issues. We've also improved the COFF_Affiliation_TDTM class to perform only DML operations on Course Connection records that need to be updated.

# Issues Closed
Fixes #534 

# New Metadata

# Deleted Metadata

# Testing Notes


src/classes/CCON_Faculty_TDTM.cls
```diff
@@ -65,169 +68,179 @@ public with sharing class CCON_Faculty_TDTM extends TDTM_Runnable {
...
+                    );
+                    courseOfferingsToUpdate.add(cOff);
+                }
             }
+        }
 
-            if(affiliationsToDelete.size() > 0) {
-                dmlWrapper.objectsToDelete.addAll((List<SObject>)affiliationsToDelete);
-            }
+        TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.CCON_Faculty_TDTM, true);
```
robertbwatson: There's quite a bit of formatting changes in this class, but these few lines are the biggest changes (as well as the reentrancy check on line 77)

src/classes/COFF_Affiliation_TDTM.cls
```diff
@@ -357,21 +360,25 @@ public class COFF_Affiliation_TDTM extends TDTM_Runnable {
...
+        for (Integer i = 0; i < oldConns.size(); i++) {
             if (offeringFacultyConnIds.containsKey(oldConns[i].Course_Offering__c)
                     && offeringFacultyConnIds.get(oldConns[i].Course_Offering__c) != null
                     && offeringFacultyConnIds.get(oldConns[i].Course_Offering__c).containsKey(updatedFacultyIDs[i])) {
+                
                 Course_Enrollment__c existingConn = offeringFacultyConnIds.get(oldConns[i].Course_Offering__c).get(updatedFacultyIDs[i]);
-                existingConn.Primary__c = true;
-                // Current Course Connection exists on Course Offering with Contact, mark as Primary
-                objectsToUpdate.add((SObject)existingConn);
+                if (!existingConn.Primary__c) {
```
robertbwatson: I've made some efficiency improvements here and further down in other methods. When testing the reentrancy prevention, I realized that this class was updating course connections even when the Primary field was already set to the correct value.

src/classes/COFF_Affiliation_TDTM.cls
```diff
@@ -462,6 +474,8 @@ public class COFF_Affiliation_TDTM extends TDTM_Runnable {
     ********************************************************************************************************/
     @testVisible
     private void updateConnectionsOnFacultyRemove(List<ID> removedFacultyIDs, Set<Id> offsWithRemovedFaculty, List<SObject> objectsToUpdate) {
+        // TODO: This doesn't delete anything, just updates course connections - should it actually delete affiliations and course connections,
```
robertbwatson: I found this to be confusing -- the ApexDocs above contradict what this method actually does. I think we need to either update the comments or update the logic -- probably best to create a tech debt work item, perhaps?

src/classes/COFF_Affiliation_TEST.cls
```diff
@@ -735,14 +735,16 @@ public with sharing class COFF_Affiliation_TEST {
...
+
         TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.COFF_Affiliation_TDTM, true);
 
         //Create Course Offering with Faculty member
         Course_Offering__c offering = new Course_Offering__c(Section_ID__c = 'BIO-101 Spring 16', Course__c = course.ID,
         Faculty__c = faculty.Id, Term__c = term.Id, Capacity__c = 200);
         insert offering;
 
-        Course_Enrollment__c conn = new Course_Enrollment__c(Contact__c = faculty.Id, Course_Offering__c = offering.Id);
+        //Create Primary Course Connection for the Faculty member
```
robertbwatson: I had to fix a few unit tests in this class because of the efficiency improvements. If the record didn't actually need to be updated, the unit tests were failing as they were not included in the objectsToInsert or objectsToUpdate lists.

src/classes/COFF_Affiliation_TDTM.cls
lifewithryan: I need to review this further. It looks like you had to add the connection to the list to be updated when it suggests this method was doing that already, so I can see why you were confused. I need to take a look at what was going on here before (outside of githubs UI because that's just difficult to read). 
In any case, if the ApexDocs are saying one thing and we are doing another we should certainly create a Tech Debt WI for this.

src/classes/COFF_Affiliation_TDTM.cls
```diff
@@ -71,19 +71,28 @@ public class COFF_Affiliation_TDTM extends TDTM_Runnable {
             Map<Id, Course_Offering__c> mapNewFacultyIdToNewCourseOffering = new Map<Id, Course_Offering__c>();
 
             if(newlist != null) {
+                
                 if (triggerAction == TDTM_Runnable.Action.AfterInsert || triggerAction == TDTM_Runnable.Action.AfterUpdate) {
                     //Query fields not available in the trigger
-                    newList = [select Course__c, Faculty__c, Start_Date__c, End_Date__c, Course__r.Account__c, Term__r.Start_Date__c,
-                                Term__r.End_Date__c from Course_Offering__c where ID in :newlist];
+                    newList = [SELECT Course__c
+                                      , Faculty__c
```
robertbwatson: @lifewithryan Commas at the beginning of a line? 🙀

src/classes/COFF_Affiliation_TDTM.cls
robertbwatson: The only code change I made here was to only add the object to objectsToUpdate if it actually needed to be updated.

I was just really confused because the ApexDocs say "Deletes the Affiliations that had been automatically created from the Course Enrollments with assigned Faculty members. Also deletes the Course Enrollments themselves." -- when it does literally none of that.

src/classes/COFF_Affiliation_TDTM.cls
lifewithryan: haha, yea, it's hold over from what I've done for years. Actually learned it from Kevin Swiggum :)

We don't *have* to do it that way, but I definitely don't want the wrapping. If it's going to wrap I'd rather line break it and line up the keywords at a minimum.

src/classes/COFF_Affiliation_TDTM.cls
lifewithryan: Yea we really need to fix the apex docs. We may want to involved Ali with this and find out what it is really supposed to be doing and either fix the docs or the code.

src/classes/COFF_Affiliation_TDTM.cls
robertbwatson: I agree about the wrapping! I'm not sure I'll adopt this way, but it's fine. I've just never seen it like this before.

src/classes/COFF_Affiliation_TDTM.cls
robertbwatson: I'm playing a bit of git blame / history tracking to see if I can figure out when this was done. I'll reach out to him next week once he's back.

src/classes/COFF_Affiliation_TDTM.cls
lifewithryan: No worries. As long as we avoid wrapping/scrolling right and we use some formatting to keep it legible, I'm good. We have three very usable options in the style guide.

src/classes/COFF_Affiliation_TDTM.cls
robertbwatson: @lifewithryan  I did some digging and this conflicting ApexDocs / functionality has been around since 2016. I did some testing and it works like this;
- This method is only called when course connections is enabled and the Primary Faculty lookup is cleared on the Course Offering. So it makes sense to update the Course Connection to no longer be primary, and to keep the data.
- When a Course Offering is deleted from the system, then the "deleteAfflsEnrolls" method (another method in this class) is called, which actually does delete Affiliations and Course Connection records.

In my opinion, I think it's safe to say that the ApexDocs are wrong and need to be updated. The functionality is correct, unless we don't want to keep historical data. But that seems to go against how EDA functions everywhere else.

robertbwatson: @lifewithryan If you agree with this, I can go ahead and update the ApexDocs here now. We can circle back with Ali & Dina next week.

lifewithryan: mighty fine detective work. If we can update the docs now, I say we do so.

robertbwatson: Done, resolving

robertbwatson:
@juliesnow Just pinging you here to let you know you have release notes to review on this one.

dgamboni:
Tested / Passes QE

jacebryan:

"We've updated the CCON_Faculty_TDTM class to prevent reentrancy issues. We've also improved the COFF_Affiliation_TDTM class to perform only DML operations on Course Connection records that need to be updated."

- What value did the changes to CCON_Faculty_TDTM class bring to EDA?
- What value did the changes to COFF_Affiliation_TDTM class bring to EDA?
- Who would find the most value in knowing about these changes/improvements?

PR #814, opened 2019-05-30
Add regression and QA flows

jstvz:
This PR adds metadata and configuration to support both feature testing and regression testing in scratch orgs.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
# Testing Notes
* Course Connection Object > Add Term Grades Related List with columns Term Grade Name, Term, Term Grade Type, Letter Grade
* Account Object > HEDA Organization Page Layout > Facilities Related List > Columns = Facility Name, Display Name, Facility Type, Capacity, Parent Facility; Sort by Display Name, Ascending
* List View for  “All Accounts” > Fields = Account Name, Account Record Type, Parent Account, Last Modified Date
* Program Enrollment Object : Page Layout : Add “Program Plan” field 
* Contact Object : Page Layout > Add “Primary Language” field to Contact Details; Add all “Primary” fields except for Primary Language under the "Primary Affiliations" section
* Course Offering Object : Page Layout > Add “Facility” field
* Facility Object : Page Layout > Add “Display Name” field


dashattuck:
reviewing right now...

jstvz:
@dashattuck The flow to run is `regression_org`, as in `cci flow run regression_org --org foo`. I'll add a README to document this.

jstvz:
Regarding the `HEDA` references, these need to remain since they're referencing the packaged metadata, e.g. src/applications/HEDA.app

juliesnow:
@jstvz - Can you pls include a brief description of the changes under "Changes?" TIA!

jstvz:
> @jstvz - Can you pls include a brief description of the changes under "Changes?" TIA!

@juliesnow Updated the PR description so it isn't blank, but my understanding was that only customer-facing changes are included under that heading. I'll update if my understanding is incorrect.

unpackaged/config/readme.md
```diff
@@ -0,0 +1,47 @@
...
+
+Key Flows
+------------------
+
+### `edaectomy`
+  Removes configuration and unmanaged metadata.
+
+
+### `dev_org`
+  Modified - Deploys Dev configiguration and runs some post-install methods as the final tasks.
```
robertbwatson: @jstvz Minor, but you've got some misspelling in a couple of places here. I don't think "configiguration" is a word.

dgamboni:
@jstvz I tested the Regression flow, and looks like it works well! -- it appears to have successfully installed the previous version and then upgraded to the current beta version. There are some additional configurations that are needed, but if we'd like to ship now I believe we can easily update those incrementally.

In terms of the **qa_org** flow -- could we have the same configurations in the **qa_org** flow as we do in the **regression_org** flow?  They are not quite the same right now, and I believe we'll want to have separate flows (or tasks?) in order to create pre-configured namespaced and non-namespaced testing orgs for qa, but without the newest beta version installed. Thank you!

jstvz:
@dgamboni I've set these up to be identical, so you shouldn't see any difference (in configuration) between the configuration `regression_org` and `qa_org`. If you do, let me know. I'll go ahead and merge tomorrow if there's no objection.

dgamboni:
Hi @jstvz I was seeing differences in the org configurations between `qa_org` and `regression_org` flows when I tested these and left the comment yesterday.

robertbwatson:
@jstvz @dgamboni I suppose this should probably be discussed in a separate WI now that this has been merged, but should the qa_org flow also run the Apex that enables Course Connections? (Maybe that's the difference that Dina saw?)

jstvz:
@robertbwatson Thanks, that's it. I'll add it and submit another PR.

PR #817, opened 2019-06-04
Adding reentrancy flag to TERM_CourseOff_TDTM

### LINKED ISSUE #536: Add reentrancy flags to TERM_CourseOff_TDTM
It seems unlikely this would experience reentrancy problems since the Term updating the related Course Offerings won't circle back to update the Term again...but this would be a good safeguard in case a customer would put any Triggers on Course Offerings that update the Term (and kick off unnecessary code cycles)

robertbwatson:
# Critical Changes

# Changes
* We've updated the TERM_CourseOff_TDTM class to prevent reentrancy issues. 

# Issues Closed
Fixes #536 

# New Metadata

# Deleted Metadata

# Testing Notes
- Create a few terms for the upcoming school year. Do not set the start date and end date fields.
- Create several hundred courses.
- For a given term, use the bulk data loader to upload ~1000 or more course offerings - multiple course offerings for each course.
  - Some of these course offerings should not have a start date or an end date.
  - Some of these course offerings should have a start date, but not an end date.
  - Some of these course offerings should have an end date, but not a start date.
  - Some of these course offerings should have the start date and end date populated.
- Navigate to the term and set the start date and end date.
- Expectations: The term should update successfully. All course offering records without a start date or end date should be updated successfully. If only one of those fields was blank, that field should be updated but the field that already had a value should not be updated.


jacebryan:

"We've updated the TERM_CourseOff_TDTM class to prevent reentrancy issues."

What value did the changes to TERM_CourseOff_TDTM class bring to EDA?
Who would find the most value in knowing about these changes/improvements?

PR #790, opened 2019-05-07
Adding reentrancy checks to REL_Relationships_Cm_TDTM & REL_Relationships_Con_TDTM classes.

robertbwatson:
# Critical Changes

# Changes

## Updates to Prevent Reentrancy Issues
* We've updated the REL_Relationships_Cm_TDTM & REL_Relationships_Con_TDTM classes to prevent reentrancy issues.

# Issues Closed
Fixes #533 

# New Metadata

# Deleted Metadata

# Testing Notes

PR #819, opened 2019-06-06
Adding reentrancy flag to REL_Relationships_TDTM 

### LINKED ISSUE #535: Add reentrancy flags to REL_Relationships_TDTM
Since this trigger handler creates reciprocal relationship records, there's a good possibility that there's some extra back-and-forth processing on the two records...even though the code is currently protected against creating extra records in error. 

Need to be careful about Reentrancy here. Inserting a relationship record will auto-create a second relationship record (Reciprocal). This class needs to process on BOTH records performing different functions. So a flag in Insert that disables the Insert operation will break the features on the Reciprocal record.

robertbwatson:
# Critical Changes

# Changes

* We've updated the REL_Relationships_TDTM class to prevent reentrancy issues on Relationship record updates.

# Issues Closed
Fixes #535 

# New Metadata

# Deleted Metadata

# Testing Notes
- Relationship Creation:
  - Create two new Contacts.
  - Create a new Relationship record, tying the two Contacts together.
  - Expectation: The Relationship record is successfully created, as well as its reciprocal Relationship record.

- Relationship Update:
  - Create another Contact.
  - Update the existing Relationship record from the first test to be related to the new Contact instead.
  - Expectation: The reciprocal Relationship record is updated to be for the new Contact instead of the old Contact.

- Relationship Update:
  - Update the existing Relationship record from the first test and specify a different Type and Description.
  - Expectation: The reciprocal Relationship record's data is also updated.

- Relationship Delete:
  - Delete one of the Relationship records.
  - Expectation: The other Relationship record is also deleted.


robertbwatson:
Keeping this as a draft PR until #790 is merged

src/classes/REL_Relationships_TDTM.cls
```diff
@@ -72,40 +87,34 @@ public class REL_Relationships_TDTM extends TDTM_Runnable {
...
                     reciprocalsToUpdate.add(r);
                 }
             }
             
             // AFTER UPDATE
-            if (triggerAction == TDTM_Runnable.Action.afterUpdate) {   
+            if (triggerAction == TDTM_Runnable.Action.AfterUpdate) {   
                 if (r.ReciprocalRelationship__c == oldList[i].ReciprocalRelationship__c 
-                && r.RelatedContact__c != null) {
+                        && r.RelatedContact__c != null) {
```
lifewithryan: how do you feel about this format:
```
if (r.ReciprocalRelationship__c == oldList[i].ReciprocalRelationship__c &&
    r.RelatedContact__c != null) {
```

no big deal and probably all about the same in the end. I've usually put my && at the end of a line so when I see them hanging, I know there's more. Either works though :)

src/classes/REL_Relationships_TDTM.cls
```diff
@@ -72,40 +87,34 @@ public class REL_Relationships_TDTM extends TDTM_Runnable {
...
                     relationshipsToUpdate.add(r);
                     origRelationships.add(oldList[i]);
-                }
-                if (r.RelatedContact__c == null) {
+                } if (r.RelatedContact__c == null) {
```
lifewithryan: Is there supposed to be a carriage return here (separate if?) or is there supposed to be an "else if" here?

src/classes/REL_Relationships_TEST.cls
```diff
@@ -817,9 +823,9 @@ public class REL_Relationships_TEST {
     * @description Method to test the UTIL_CustomSettingsFacade class as a different user
     */
     @isTest
-    public static void testRelationshipUTILS(){
+    public static void testRelationshipUTILS() {
         //to test the Relationship_Utils class
-        System.runAs(new User(Id = UserInfo.getUserId())){
+        System.runAs(new User(Id = UserInfo.getUserId())) {
```
lifewithryan: reminds me of LISP...lol

src/classes/REL_Relationships_TDTM.cls
robertbwatson: I don't want to mess with the existing logic too much here, so I've just added the carriage returns back.

src/classes/REL_Relationships_TDTM.cls
xinnanlisf: I can't say I like the idea of putting logical operators at the end of the line. It is probably fine here but once the logic getting more complicated, it will be really hard to read...

robertbwatson: I moved it back to the end of the line. I don't really have a preference either way.

jacebryan:

"We've updated the REL_Relationships_TDTM class to prevent reentrancy issues on Relationship record updates."

What value did the changes to REL_Relationships_TDTM class bring to EDA?
Who would find the most value in knowing about these changes/improvements?

Issue #820, opened 2019-06-06
Account Record Type change to non-mapped record type results in null pointer exception error


robertbwatson:
* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the current behavior?**
If we update an account record type from a record type that has an affiliation mapping to a record type that does not have an affiliation mapping and this account has a primary affiliation connection with a contact, then the system will throw a null pointer exception when saving.

* **What is the expected behavior?**
A null pointer exception should not be thrown.

* **What is the motivation / use case for changing the behavior?**
Users should be able to change the account record type successfully

* **Acceptance Criteria**
1. Create a Contact
2. Create an Account with Department record type - this record type should exist in the Affiliation Mappings in the EDA Settings
3. Create a Primary Affiliation between Account and Contact
4. Update Account record type to Business Organization - this record type should NOT exist in the Affiliation Mappings in the EDA Settings
5. System should NOT throw an exception

robertbwatson:
**lurch: attach W-021765

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x800000097QzqAAE" target="_blank">W-021765</a>

jstvz:
Included in beta release 1.79 (Beta 8)

mrbelvedere:
Included in production release 1.79

PR #823, opened 2019-06-20
Fixed NPE and Added a New Unit Test

skosgi1:
# Critical Changes

# Changes
We've fixed a null pointer exception that occurs when the record type of an Account with a primary affiliation is changed to a record type with no affiliated mapping under EDA Settings. With this change, EDA now updates the Account's record type as specified when this type of change occurs.

# Issues Closed
Fixes <a href="https://github.com/SalesforceFoundation/EDA/issues/820#issuecomment-499705630">#820</a>

# New Metadata

# Deleted Metadata

# Testing Notes
1. Create a Contact
2. Create an Account with Department record type - this record type should exist in the Affiliation Mappings in the EDA Settings
3. Create a Primary Affiliation between Account and Contact
4. Update Account record type to Business Organization - this record type should NOT exist in the Affiliation Mappings in the EDA Settings
5. System should not throw an exception and "Primary Department" field should be cleared.


robertbwatson:
@skosgi1 Can you help Julie by supplying an initial description of the issue that was fixed here under "Changes"?

skosgi1:
@robertbwatson @juliesnow Added description of the fix in "Changes"

src/classes/AFFL_AccRecordType_TDTM.cls
```diff
@@ -146,9 +146,10 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
                     if (!containsNew(accsTypeChanged, accNewOldType.oldType)) {
                         contact.put(accNewOldType.oldType, null);
```
robertbwatson: I'm not exactly sure how this line of code is reached, but I'd expect a NPE to be potentially thrown here as well. How do we enter this logic? Does it get called when switching from a record type that wasn't mapped to a record type that is mapped?

src/classes/AFFL_AccRecordType_TEST.cls
```diff
@@ -378,4 +378,53 @@ public with sharing class AFFL_AccRecordType_TEST {
...
         System.assertEquals(true, allAffiliationList[0].Primary__c);
     }
+
+    @isTest
+    public static void checkNPE() {        
+        
+        setup();
+
+        //delete an affiliation mapping
+        List<Affl_Mappings__c> mappings = [select ID from Affl_Mappings__c where Name = 'Household Account'];
```
robertbwatson: This is minor because the rest of the class might be written like this as well, but for the SOQL keywords we prefer them to be capitalized - like "SELECT", "FROM", "WHERE", etc

robertbwatson: Honestly, lines 388-391 could just be consolidated into one line: `delete [SELECT Id FROM Affl_Mappings__c WHERE Name = 'Household Account'];` (no need to check if there are rows returned)

src/classes/AFFL_AccRecordType_TDTM.cls
skosgi1: @robertbwatson Not sure I am following your question, the updatePrimaryAfflsOnChange method gets called from the run method, if the account already had a record type those would get stored in accTypesChanged, else they would get stored in accTypesAdded so  accNewOldType.oldType wouldn't throw an exception since that is a record from the list accTypesChanged.

src/classes/AFFL_AccRecordType_TEST.cls
skosgi1: @robertbwatson updated the test class.

src/classes/AFFL_AccRecordType_TDTM.cls
lifewithryan: I'm having a heck of time deciphering what this class as a hole is doing so I have to grok it a bit. Will add my comments as soon as I can. The GH interface by only showing me the changes is making it tough to absorb so I will review the file as a whole and come back and make comments.

robertbwatson: @skosgi1 Yeah, I was just concerned if accNewOldType.oldType could ever be null because if so, we'd get a NPE here.

It looks like we're checking for null on line 97 while we're building this list (https://github.com/SalesforceFoundation/EDA/blob/fa783e6f7d965aa2b615920950678752868dad76/src/classes/AFFL_AccRecordType_TDTM.cls#L97), so I think this is safe to leave as-is.

robertbwatson: Ooh, nice - thanks Github for formatting this so nicely :)

robertbwatson:
@skosgi1 FYI for future - I noticed you did a force-push here (https://github.com/SalesforceFoundation/EDA/compare/113ea78589f57b06fc5ee3af29390ea0fca77373..c942327f3fda4e6d24b6de4a667468244efc6014) - be sure to always make sure to pull down the latest from the remote branch, otherwise you might overwrite others commits. We have an automatic process that commits the latest changes from master into each feature branch every time we merge a new feature.

src/classes/AFFL_AccRecordType_TDTM.cls
```diff
@@ -146,9 +146,10 @@ public class AFFL_AccRecordType_TDTM extends TDTM_Runnable {
...
                     }
-                    contact.put(accNewOldType.newType, accNewOldType.accId);
-                    dmlWrapper.objectsToUpdate.add(contact);
-
+                    if (accNewOldType.newType!=NULL) {
+                        contact.put(accNewOldType.newType, accNewOldType.accId);
+                        dmlWrapper.objectsToUpdate.add(contact);
```
robertbwatson: @skosgi1 I just did a double-check on this and need to assign it back to you. Because the contact could be updated on line 147, we need to add the contact record to the dmlWrapper outside of this null check as was being done previously.

(This might affect W-033954?)

robertbwatson:
@skosgi1 Additionally, we realized during our release review meeting today that you need the # in the issue number in order for it to be automatically picked up in the release notes. So the "Issues Closed" section should actually read: "Fixes #820"

src/classes/AFFL_AccRecordType_TDTM.cls
lifewithryan: This could be a candidate for a very specific null check unit test, no?

src/classes/AFFL_AccRecordType_TDTM.cls
skosgi1: Thanks for catching that @robertbwatson. You're right.

skosgi1:
@robertbwatson @lifewithryan Updated the logic to reflect contact update and updated the test class.

src/classes/AFFL_AccRecordType_TDTM.cls
skosgi1: I think this was scenario was captured in changePrimaryAffiliationRecordTypeReentrancyCheck().

src/classes/AFFL_AccRecordType_TEST.cls
```diff
@@ -378,4 +378,55 @@ public with sharing class AFFL_AccRecordType_TEST {
...
+
+        Test.stopTest();
+
+        //Verify no duplicate affiliations are created and no NPE
+        List<Affiliation__c> allAffiliationList = [SELECT Id, Name, Primary__c FROM Affiliation__c 
+                                                   WHERE Contact__c = :contact.Id 
+                                                   AND Account__c = :bizAcc.Id] ;
+        
+        //Verify Primary Organization is Null
+        Contact updatedContact = [SELECT id, Primary_Organization__c, Primary_Household__c FROM contact WHERE id = :contact.Id];
```
robertbwatson: @skosgi1 Interesting - does that mean that perhaps W-033954 isn't actually a bug? Or is that a different scenario?

skosgi1: @robertbwatson You're right. It isn't a bug. It should be a part of this fix. I will comment on W-033945 that it is going to be a part of W-021765.

robertbwatson: We can probably just close it & the Github issue if it wasn't ever a bug to begin with

dgamboni:
@juliesnow Another suggestion for release notes to add some more detail : "Previously, if an Account had a Primary Affiliation with a Contact, and that affiliated Account’s record type was updated from one that had an affiliation mapping in EDA Settings to a record type without an affiliation mapping in EDA Settings, a null pointer exception was thrown. That NPE has been fixed in this release.” (I realize this is passive voice which isn't always ideal, but I thought it could help give you a little more insight into the issue that was fixed to wordsmith it for our customers.) Please let us know when the text is approved so we can merge & close. Thanks!!!

PR #826, opened 2019-06-20
Localization First Delivery June 2019

GloryChoe:


robertbwatson:
@lifewithryan FYI, I noticed there are a lot of merge conflicts here - likely from all of our merges last week. Should someone look into this soon? (I think someone not on our team is already testing this?)

GloryChoe:
Will it help if I update to the latest and then regenerate translation files? @lifewithryan @robertbwatson

lifewithryan:
> Will it help if I update to the latest and then regenerate translation files? @lifewithryan @robertbwatson

@GloryChoe -- that might help. For clarity -- doing so will NOT add to the cost of translation correct? The "new" fields that have since been added will remain un-translated?

lifewithryan:
> Will it help if I update to the latest and then regenerate translation files? @lifewithryan @robertbwatson

@GloryChoe part of the issue is that you all were so efficient in getting this done and we've had numerous large pushes since you've submitted the PR. Is it possible to not re-pull/re-generate until this first pass has been approved and ready for merge. That might cut down on the number of conflicts since hopefully it wouldn't be sitting out as long. I'm sure we'd still do another pass at the end, but hopefully that would go faster.

GloryChoe:
> > Will it help if I update to the latest and then regenerate translation files? @lifewithryan @robertbwatson
> 
> @GloryChoe part of the issue is that you all were so efficient in getting this done and we've had numerous large pushes since you've submitted the PR. Is it possible to not re-pull/re-generate until this first pass has been approved and ready for merge. That might cut down on the number of conflicts since hopefully it wouldn't be sitting out as long. I'm sure we'd still do another pass at the end, but hopefully that would go faster.

No problem, could you please ping me when you are all ready for me to re-pull/re-generate?

PR #830, opened 2019-06-24
Use namespace tokens for packaged attribute record types

jstvz:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/unpackaged/config/dev/profiles/Admin.profile
+++ b/unpackaged/config/dev/profiles/Admin.profile
@@ -38,11 +38,11 @@
     </layoutAssignments>
     <layoutAssignments>
         <layout>%%%NAMESPACE%%%Attribute__c-%%%NAMESPACE%%%EDA Credential Layout</layout>
-        <recordType>%%%NAMESPACE%%%Attribute__c.%%%NAMESPACED_ORG%%%Credential</recordType>
+        <recordType>%%%NAMESPACE%%%Attribute__c.%%%NAMESPACE%%%Credential</recordType>
     </layoutAssignments>
     <layoutAssignments>
         <layout>%%%NAMESPACE%%%Attribute__c-%%%NAMESPACE%%%EDA Student Characteristic Layout</layout>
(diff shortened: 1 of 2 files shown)


unpackaged/config/dev/profiles/Admin.profile
```diff
@@ -38,11 +38,11 @@
     </layoutAssignments>
     <layoutAssignments>
         <layout>%%%NAMESPACE%%%Attribute__c-%%%NAMESPACE%%%EDA Credential Layout</layout>
-        <recordType>%%%NAMESPACE%%%Attribute__c.%%%NAMESPACED_ORG%%%Credential</recordType>
+        <recordType>%%%NAMESPACE%%%Attribute__c.%%%NAMESPACE%%%Credential</recordType>
     </layoutAssignments>
     <layoutAssignments>
         <layout>%%%NAMESPACE%%%Attribute__c-%%%NAMESPACE%%%EDA Student Characteristic Layout</layout>
-        <recordType>%%%NAMESPACE%%%Attribute__c.%%%NAMESPACED_ORG%%%Student_Characteristic</recordType>
+        <recordType>%%%NAMESPACE%%%Attribute__c.%%%NAMESPACE%%%Student_Characteristic</recordType>
```
robertbwatson: @jstvz On line 55 & in other places, we have NAMESPACED_ORG - is that inaccurate? (Perhaps it works because those record types are unmanaged, so therefore they don't ever have a namespace?)

lifewithryan: beat me by 24 seconds!

lifewithryan: @jstvz I've always wondered what the difference was here...if you have some time someday to elaborate, I'd be all ears :)

jstvz: So, you're correct. The distinction is between orgs that have a namespace applied (Packaging and namespaced scratch orgs) because those orgs treat any metadata as having a namespace, whether it is deployed as unpackaged or packaged.

lifewithryan:
@davidmreed @davisagli -- can one of y'all look at this? we need to get QE on this for regression.

PR #833, opened 2019-06-27
NPE Check on Affl deletion and new test method

skosgi1:
# Critical Changes

# Changes
We've fixed a null pointer exception that occurs when a primary Affiliation with no Contact is deleted. With this change, EDA now deletes the Affiliation.

# Issues Closed
Fixes <a href="https://github.com/SalesforceFoundation/EDA/issues/834">#834</a>

# New Metadata

# Deleted Metadata

# Testing Notes
1) Create an account
2) Create a Contact
3) Create a primary affiliation between the account and the contact
4) Update the Affiliation Page layout by making the field 'Contact' not required
5) Clear the value in 'Contact' field
6) Delete the affiliation
The record should be deleted and no error messages like "There's a problem saving this record. You might not have permission to edit it, or it might have been deleted or archived. Contact your administrator for help." should be seen.


ceiroa:
@skosgi1 Pease don't forget to list the issue closed in the Issues Closed section (with pound so the release notes script picks it up). Thanks!

skosgi1:
@ceiroa Got it! Thank you. But this one does not have an issue this is a part of W-032893.

ceiroa:
@skosgi1 Ah, ok. Then please reference that one (the equivalent in Github) under issues closed. Github should link that to both PRs. Thanks!

src/classes/AFFL_MultiRecordType_TDTM.cls
```diff
@@ -141,7 +141,7 @@ public class AFFL_MultiRecordType_TDTM extends TDTM_Runnable {
 					Contact relatedContact = relatedContactsMap.get(afflOld.Contact__c);
 
 					//AFTER DELETE - delete lookup relationship, if necessary
-					if (triggerAction == TDTM_Runnable.Action.AfterDelete && lookupFieldName != null) {
+					if (triggerAction == TDTM_Runnable.Action.AfterDelete && lookupFieldName != null && relatedContact != null) {
```
robertbwatson: Especially because the processAfflDeleted() method is a public method, I would prefer we do the null check in the method itself rather than before calling the method (or in addition to)

src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -416,4 +416,49 @@ public with sharing class AFFL_MultiRecordType_TEST {
 		System.assertEquals(false, result.success);
 		System.assertEquals(Label.afflAccoutMappingError, result.errors[0].message);
 	}
+
+	@isTest
```
robertbwatson: I know we're not consistent in this class, but we should be adding ApexDoc-style comments to every new unit test so that it's clear what each unit test is testing.

skosgi1:
@robertbwatson Updated the classes with requested changes, can you give it another shot when you get a chance? Thank you.

dgamboni:
Tested as part of https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E0000087vZ4QAI/view / Passes QE.

Issue #807, opened 2019-05-23
New Field on Contact "Preferred Name"


JasonFitzy:
As an educational organization I would like the option to use a "Preferred Name" field on Contact so that we can recognize Contacts as they choose, but still maintain records of their legal name.

Acceptance Criteria

1. There is a "Preferred Name" field on Contact
2. The field is not added to a page layout

JasonFitzy:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087v2FQAQ" target="_blank">W-033313</a>

PR #850, opened 2019-07-03
Adding Chosen Full Name to Contact

lifewithryan:
# Critical Changes

# Changes
- We've added a Chosen Full Name field to allow Contacts to have themselves referred to by a name chosen by them.
# Issues Closed
Fixes: #807
# New Metadata
- Contact.Chosen_Full_Name__c
# Deleted Metadata

# Testing Notes


lifewithryan:
@juliesnow -- could use some assistance on the Chosen Full Name release notes. I feel like we want to emphasize the reasoning behind it, but maybe not?

src/objects/Contact.object
```diff
@@ -19,6 +19,18 @@
         <type>Email</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>Chosen_Full_Name__c</fullName>
+        <description>The full name that the Contact prefers to use instead of a legal or official name.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The full name that the Contact prefers to use instead of a legal or official name.</inlineHelpText>
```
robertbwatson: If we retrieve the metadata from the org, there should be corresponding translation files with changes as well.

lifewithryan: Yep good call. I added this manually via the XML...I bet I can add that manually as well to the necessary files.

lifewithryan: If not, I will pull.

lifewithryan:
added the translations (decided to pull)

src/objectTranslations/Contact-ca.objectTranslation
```diff
@@ -4422,6 +4427,11 @@
             <translation>Feina</translation>
         </picklistValues>
     </fields>
+    <fields>
```
robertbwatson: @lifewithryan Unfortunately, we have to make sure we only include Chosen Full Name on all of these because the Primary fields are unmanaged.

lifewithryan: DOH! Thanks...

robertbwatson:
@juliesnow  I'm merging this in today in order to hit the merge freeze, but we'll need your review here on Monday before we get the 👍 on the overall 1.79 release notes.

PR #861, opened 2019-07-10
Sync Course Connection Contact and Contact lookup field on Attendance Event

### LINKED ISSUE #744: Attendance Event: Triggers and TDTM Enabled
New feature, related to #723 

N/A

Triggers and TDTM class setup for Attendance Events

* **Acceptance Criteria**
- Triggers are visible from TDTM area/tab in HEDA
- Ability to Enable/Disable Triggers

xinnanlisf:
# Critical Changes

# Changes
## New Trigger Handlers
We’ve added a new trigger handler (Attendance_Event__c) that enables the following behaviors:

- If you specify a Contact for an Attendance Event that is different than what's specified for the associated Course Connection, an error message appears and the record isn't saved.
- If you specify an Attendance Event that has an associated Course Connection, but you don't specify a Contact, the Attendance Event is populated with the Contact from the Course Connection.
- If you change the Contact on a Course Connection after an Attendance event has been created, a Contact mismatch might occur between the two records. This will be addressed in a future release.

This trigger handler is enabled by default, but you can deactivate it from the “Trigger Handlers” tab.

# Issues Closed
Fixes #744
# New Metadata
Custom Label: AttendanceCourseConnectionContactMismatch
Class: ATTD_CourseConnectionContact_TDTM
Class: ATTD_CourseConnectionContact_TEST
# Deleted Metadata

# Testing Notes
1. Create two contacts.
2. Create a course connection who is linked to the first contact. 
3. Create an Attendance Event who is linked to the first contact and the course connection.
4. Make sure the attendance event can be saved.
5. Go to the newly created attendance event and update the contact to the second contact.
6. Make sure an error is thrown
7. Create an Attendance Event who is linked to the first contact and not the course connection.
8. Make sure the attendance event can be saved.
9. Create an Attendance Event who is linked to the the course connection not the contact.
10. Make sure the contact on Attendance Event is the same to the contact on course connection.


xinnanlisf:
@juliesnow would you mind take a review on this PR please? Thank you!

src/classes/ATTD_CourseConnectionContact_TDTM.cls
```diff
@@ -0,0 +1,90 @@
...
+            if (!recursionFlag) {
+                for (Attendance_Event__c attd : [SELECT Id,
+                                                        Contact__c,
+                                                        Course_Connection__c,
+                                                        Course_Connection__r.Contact__c
+                                                FROM Attendance_Event__c 
+                                                WHERE Id IN :newlist]) {
+                    if (attd.Course_Connection__c != null
+                        && attd.Course_Connection__r.Contact__c != null
+                        && attd.Course_Connection__r.Contact__c != attd.Contact__c) {
```
skosgi1: @xinnanlisf As per our conversation we could move attd.Course_Connection__r.Contact__c != attd.Contact__c to line 68.

src/classes/ATTD_CourseConnectionContact_TEST.cls
```diff
@@ -0,0 +1,260 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Attendance Events
+* @group-content ../../ApexDocContent/AttendanceEvents.htm
+* @description Tests specific to Attendance Events
+*/
+@isTest
+public class ATTD_CourseConnectionContact_TEST {
```
skosgi1: @xinnanlisf As per convo we could change the access modifiers to private.

src/package.xml
```diff
@@ -141,6 +143,7 @@
         <members>TDTM_Account</members>
         <members>TDTM_Address</members>
         <members>TDTM_Affiliation</members>
+        <members>TDTM_Attendance</members>
```
skosgi1: @xinnanlisf As per our convo we could change the name to TDTM_AttendanceEvent

src/classes/ATTD_CourseConnectionContact_TDTM.cls
```diff
@@ -0,0 +1,93 @@
...
+                    }
+                }
+            }     
+        }
+        
+        TDTM_TriggerHandler.processDML(dmlWrapper, true);
+        
+        //Reset recursion flag
+        dmlWrapper = null;
+        TDTM_ProcessControl.setRecursionFlag(TDTM_ProcessControl.registeredTrigger.ATTD_CourseConnectionContact_TDTM, false);
```
robertbwatson: We need to set it to true before line 85 or it will never be true, I think?

src/classes/ATTD_CourseConnectionContact_TDTM.cls
```diff
@@ -0,0 +1,93 @@
...
+        DmlWrapper dmlWrapper = new DmlWrapper();
+
+        boolean recursionFlag = TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.ATTD_CourseConnectionContact_TDTM);
+
+        Map<ID, SObject> newmap = new Map<ID, SObject>(newlist);
+        
+        if (triggerAction == TDTM_Runnable.Action.AfterInsert
+                || triggerAction == TDTM_Runnable.Action.AfterUpdate) {
+                    
+            if (!recursionFlag) {
```
robertbwatson: There's a lot of nested if statements here - can we check if recursionFlag is true after line 51 and if it is, just exit this method entirely?

src/classes/ATTD_CourseConnectionContact_TDTM.cls
```diff
@@ -0,0 +1,93 @@
...
+* @author Salesforce.org
+* @date 2019
+* @group Attendance Events
+* @group-content ../../ApexDocContent/AttendanceEvents.htm
+* @description Makes sure that linked course connection's contact matches contact lookup field
+*/
+public with sharing class ATTD_CourseConnectionContact_TDTM extends TDTM_Runnable {
+    /*******************************************************************************************************
+    * @description Makes sure that linked course connection's contact matches contact lookup field
+    * @param listNew the list of Attendance Events from trigger new.
```
robertbwatson: Minor, but the params are actually newlist and oldlist

src/classes/ATTD_CourseConnectionContact_TDTM.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>46.0</apiVersion>
```
robertbwatson: Do we want all new classes and trigger on 46.0 or do we want to wait until we move all metadata to 46.0?

robertbwatson: (Probably a larger question that we need to agree on as an entire team)

src/labels/CustomLabels.labels
```diff
@@ -16,6 +16,14 @@
         <shortDescription>AfflProgEnrollDeleted</shortDescription>
         <value>If selected, when deleting a Program Enrollment, delete its related Affiliation. Otherwise, keep the Affiliation and set its Status to the value defined in Status Specified for Affiliations Not Deleted.</value>
     </labels>
+    <labels>
```
robertbwatson: Minor, but I think there should be changes to other translation files as a result of this new label - if we retrieve the metadata from the scratch org, they should get pulled down automatically.

src/classes/ATTD_CourseConnectionContact_TDTM.cls
xinnanlisf: Good catch

src/classes/ATTD_CourseConnectionContact_TDTM.cls
xinnanlisf: Well I don't think it will run those nested if statements if recursionflag is true anway

src/classes/ATTD_CourseConnectionContact_TDTM.cls
xinnanlisf: Updated.

src/classes/ATTD_CourseConnectionContact_TDTM.cls-meta.xml
xinnanlisf: Updated to 45.0 for now.

xinnanlisf:
@robertbwatson should be good now.

src/classes/ATTD_CourseConnectionContact_TDTM.cls
robertbwatson: That's fine, I was just hoping we could get rid of so many nested if statements here. I'm a big fan of exiting earlier and the ApexPMD plugin would probably flag this for having too many nested levels, but for sake of getting this through I'm okay with leaving this as-is.

src/triggers/TDTM_AttendanceEvent.trigger-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexTrigger xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>46.0</apiVersion>
```
robertbwatson: This is still 46.0 but I talked to @lifewithryan about this today and I think it's okay to leave as-is.

jacebryan:
CCE feedback on release note usability (post merge freeze):

“Require Matching Contacts for Course Connections and Attendance Events”

- What value did the introduction of Attendance_Event__c trigger bring to EDA?
- Are there any anticipated impacts that should be tested based on the introduction of the Attendance_Event__c trigger (especially since its enabled by default)?
- Who is the intended audience for this new trigger?

PR #873, opened 2019-07-25
EDA Telemetry

robertbwatson:
# Critical Changes

# Changes
## EDA Feature Telemetry
* EDA Feature Telemetry collects feature and usage metrics when customers use our products. We use this information to understand the level of usage and adoption of various EDA features. No personally identifiable information is collected through EDA Feature Telemetry. Read more about EDA Feature Telemetry here: [link to EDA Feature Telemetry]. We added the following to EDA Feature Telemetry:
  * Automatic Household Naming Enabled
  * Count of Active Course Connection Record Types
  * Count of Administrative Accounts
  * Count of Auto-Create Campaign Member Relationship Rules
  * Count of Auto-Create Contact Relationship Rules
  * Count of Course Connections
  * Count of Course Offering Schedules
  * Count of Error Logs
  * Count of Facilities
  * Count of Household Accounts
  * Count of Plan Requirements
  * Count of Program Plans
  * Count of Term Grades
  * Course Connections Enabled
  * Default Account Model = Administrative
  * Default Account Model = Household
  * Error Handling Enabled
  * Has Custom Administrative Account Naming
  * Has Custom Household Account Naming
  * Has Inactivated EDA TDTM Triggers
  * Has User-Managed TDTM Triggers
  * Relationship Reciprocal Method = Value Inversion
  * Storing Errors Enabled
  * Using Old Contact Ethnicity Values
  * Using Old Course Description Field

# Issues Closed

# New Metadata
New Apex Classes:
* UTIL_FeatureManagement.cls
* UTIL_FeatureManagement_TEST.cls
* UTIL_OrgTelemetry.cls
* UTIL_OrgTelemetry_BATCH.cls
* UTIL_OrgTelemetry_TEST.cls

New Feature Parameter Booleans:
* HasCustomAdministrativeAccountNaming
* HasCustomHouseholdAccountNaming
* HasDefaultAdministrativeAccountModel
* HasDefaultHouseholdAccountModel
* HasEDAInactiveTDTM
* HasUserManagedTDTM
* HasValueInversionReciprocalMethod
* IsEnabled_AutomaticHouseholdNaming
* IsEnabled_CourseConnections
* IsEnabled_ErrorHandling
* IsEnabled_StoreErrors
* UsingOldContactEthnicity
* UsingOldCourseDescription

New Feature Parameter Integers:
* Data_CountAdministrativeAccounts
* Data_CountCampaignMemberRelationshipRules
* Data_CountContactRelationshipRules
* Data_CountCourseConnections
* Data_CountCourseOfferingSchedules
* Data_CountErrorLog
* Data_CountFacilities
* Data_CountHouseholdAccounts
* Data_CountPlanRequirements
* Data_CountProgramPlans
* Data_CountTermGrades
* Org_CountActiveCourseConnectionRecordTypes

# Deleted Metadata

# Testing Notes
While no telemetry data can be sent to the FMA org with scratch orgs, we can test that the batch job executes successfully by running the following command via anonymous Apex & monitoring the status of the Apex Jobs: `UTIL_OrgTelemetry.dispatchTelemetryBatchJob();`


src/classes/STG_InstallScript.cls
```diff
@@ -181,32 +184,6 @@ global without sharing class STG_InstallScript implements InstallHandler {
         }
     }
 
-    /*******************************************************************************************************
```
robertbwatson: The changes in this file, besides the new sendTelemetryData() method, are just a result of me re-arranging the methods to go in order of global -> public -> private

robertbwatson:
@lifewithryan I'm keeping this in Draft mode so that we don't merge it early, but if you have time you can start looking over the telemetry architecture here.

src/classes/STG_InstallScript.cls
```diff
@@ -284,4 +237,69 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+
+        for(TDTM_Global_API.TdtmToken token : TDTM_Global_API.getDefaultTdtmConfig() ) {
+            objectTypes.add(token.targetObject);
+            classNames.add(token.className);
+        }
+
+        // Get EDA trigger Handlers from the database
+        List<Trigger_Handler__c> hedaHandlers = [SELECT Id, Name, Owned_by_Namespace__c FROM Trigger_Handler__c WHERE Class__c in: classNames AND Object__c in: objectTypes LIMIT 5000];
+
+        for(Trigger_Handler__c t : hedaHandlers) {
```
lifewithryan: Although they're hard to read, should we consider using the SOQL for loop here? I see we are retrieving 5000 records as a limit (though I'd be very shocked if we had anyone even that close). However if someone were to have 5000 records actually returned, the loop would run 5000 times versus using the SOQL for loop with a list which would run that for loop in chunks of 200 which would run the loop 25 times. (Granted the loop isn't doing much). Just thinking "out loud" :)

src/classes/STG_InstallScript.cls
```diff
@@ -284,4 +237,69 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+    * understanding of how our customers use our application.
+    */
+    private static void sendTelemetryData() {
+        sendTelemetryDataMethodCalled = true;
+        if (!Test.isRunningTest()) {
+            UTIL_OrgTelemetry.dispatchTelemetryBatchJob();
+        }
+    }
+
+    @testVisible
```
lifewithryan: I see the private static variable here but it appears below the method that is using it. Not something I am used to seeing. Thoughts?

src/classes/UTIL_FeatureManagement_TEST.cls
```diff
@@ -0,0 +1,90 @@
...
+
+        public Object handleMethodCall(
+            Object stubbedObject,
+            String stubbedMethodName,
+            Type returnType,
+            List<Type> listOfParamTypes,
+            List<String> listOfParamNames,
+            List<Object> listOfArgs
+        ) {
+            switch on stubbedMethodName {
```
lifewithryan: is this our very first EDA switch statement?? :)

src/classes/STG_InstallScript.cls
robertbwatson: It looks like this only runs when upgrading from a version lower than 1.38, so I don't think it's worth the effort to change. (I could update the ApexDocs to reflect that though, as it's probably something we can remove eventually...)

src/classes/STG_InstallScript.cls
robertbwatson: Where should it go? Should it go above the method? I don't really have a preference, though I don't necessarily like sticking it at the top of the class when it's a larger class like this, since it's only used by this one specific method.

src/classes/UTIL_FeatureManagement_TEST.cls
robertbwatson: UTIL_OrgTelemetry.cls also has one! This is thanks to NPSP and Michael Smith, though -- I copy/pasted a lot of this directly from the Cumulus repo...

src/classes/STG_InstallScript.cls
lifewithryan: I would stick it above the method...but that's just me, I'm not gonna force ya to do it :)

robertbwatson: Sounds good! I've made the change in my edaTelemetry__v1Features child branch that I'm actively working in.

src/classes/STG_InstallScript.cls
```diff
@@ -284,4 +237,69 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+        //Type '7' is for scheduled Apex
+        List<CronTrigger> scheduledJobs = [select Id, CronJobDetail.Id from CronTrigger
+                    where CronJobDetail.Name = :JobName and CronJobDetail.JobType = '7'];
+
+        if(scheduledJobs.size() == 0) {
+            Type classType = Type.forName(className);
+
+            if(classType != null) {
+               Object instance = classType.newInstance();
+               if(instance instanceof Schedulable)
```
lifewithryan: friends don't let friends skip the curly braces..

robertbwatson: Oh goodness! This is pre-existing code but I'll change it for sure!

lifewithryan: no problem...it was showing up as all green for me otherwise I'd probably would have just noted.

robertbwatson:
**lurch: attach W-033620

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E0000087vQlQAI" target="_blank">W-033620</a>

robertbwatson:
@lifewithryan I've moved this from draft mode to ready for review - feel free to do another once-over when you get a chance :)

src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -451,7 +451,7 @@ public class UTIL_UnitTestData_TEST {
     public static Course_Offering__c createCourseOffering(Id courseId, Id termId) {
         Account acct;
         if (courseId == null || termId == null) {
-            acct = UTIL_UnitTestData_API.getMultipleTestAccounts(1, UTIL_Describe_API.getAdminAccRecTypeID())[0];
+            acct = UTIL_UnitTestData_API.getMultipleTestAccounts(1, UTIL_Describe_API.getAcademicAccRecTypeID())[0];
```
lifewithryan: @robertbwatson does this have the possibility of being null/empty if the customer org has deleted the Academic record type? I think our UTIL describe stuff takes care of this for us but thought I'd ask.

robertbwatson: Hmm - it's possible that null could be returned - but I think in customer orgs, we can't ever guarantee that unit tests will always pass if they make modifications to their configuration. Technically the same could be said if the customer deleted/disabled the administrative record type as well.

src/classes/UTIL_OrgTelemetry.cls
```diff
@@ -0,0 +1,255 @@
...
+        for (Schema.PicklistEntry pe : picklistEntries) {
+            if (pe.isActive()) {
+                activeEthnicityPicklistValues++;
+            }
+        }
+
+        // Determine if the customer is using the new Extended Description field on the Course object.
+        // For this, we look at the 500 most recently modified Course records.
+        List<Course__c> courses = [SELECT Description__c, Extended_Description__c FROM Course__c ORDER BY LastModifiedDate DESC LIMIT 500];
+        for (Course__c course : courses) {
```
robertbwatson: @lifewithryan Thanks for calling out this -- from the link you sent me, "Developers should always use a SOQL for loop to process query results that return many records, to avoid the limit on heap size." -- I hadn't really considered heap size, and since Extended Description is a long text area, maybe we would be better off doing the SOQL in the for loop.

Honestly I almost /never/ do SOQL in a for loop -- I don't really think I ever understood the reasoning behind it before. So thanks!

robertbwatson: Updated.

robertbwatson:
@lifewithryan I spent a few hours testing this and all looks good to me. Are we ready to merge this?

cc @jstvz as a heads up - the parameters should be headed over to the packaging org soon...

PR #876, opened 2019-07-26
EDA Telemetry V1 Features merge to feature/rwatson/edaTelemetry

robertbwatson:
Note: This is just merging the child edaTelemetry__v1Features branch into edaTelemetry, so no release notes are required. Full release notes will be added to #873 

This child branch contains the 25 Feature Parameters for V1 of EDA Feature Telemetry and the associated logic & unit tests.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


src/classes/UTIL_OrgTelemetry.cls
```diff
@@ -85,10 +111,145 @@ public without sharing class UTIL_OrgTelemetry {
...
+        featureManager.setPackageBooleanValue(TelemetryParameterName.UsingOldContactEthnicity.name(), activeEthnicityPicklistValues > 2);
+
+        featureManager.setPackageBooleanValue(TelemetryParameterName.UsingOldCourseDescription.name(), usingOldCourseDescription);
+    }
+
+    /**
+     * @description Handles the calculation of all EDA feature telemetry that involves counting
+     * custom object record count.
+     */
+    private void handleObjectUsageTelemetry() {
```
robertbwatson: My only concern with this is that we may be trying process too many records at once and run into CPU limits.

I'm not concerned about hitting row limits, as the count() query only counts as one row queried. But for objects like Accounts, Course Connections, etc -- do we need to be worried about a high amount of data?

The nice thing is, since we're building this into separate batches already, it's really easy to break this out into additional batches if we need to. And we can implement counting similar to NPSP did for counting Opportunities. I just don't know if the queries against some of these objects will cause timeouts or not...

robertbwatson: If orgs do run into issues, the default value of these parameters is -1 (not 0), so we should be able to run a report in WLMA to determine if there are issues here.

robertbwatson: @lifewithryan I think yes, with additional work required. NPSP added an additional batch class that implements Database.Stateful to count with limits: https://github.com/SalesforceFoundation/Cumulus/blob/master/src/classes/UTIL_OrgTelemetry_SObject_BATCH.cls

robertbwatson:
@lifewithryan If you're good with this then, can I merge this into the edaTelemetry parent branch & close this? Then we'll just have one final PR to review & QA.

lifewithryan:
> @lifewithryan If you're good with this then, can I merge this into the edaTelemetry parent branch & close this? Then we'll just have one final PR to review & QA.

Yep let's do it

PR #877, opened 2019-07-30
Error Message on Relationship object

### LINKED ISSUE #865: EDA : Fix typos in error message when attempting to edit the Related Contact in a Relationship
**Do you want to request a *feature* or report a *bug*?**
-BUG

**What is the current behavior?**
-There are two typos in the error message given when the user attempts to change the Related Contact in a Relationship. (The word "on" should be "one" and the sentence is missing a period at the end of the sentence)

**What is the expected behavior?**
-No typos in the error message. (change "on" to "one" and add a period at the end of the sentence.)

**Steps to Reproduce:**
1. Create 3 Contacts : Giovanna Gamberini, Louisa Gamberini, and Diana Gamberini
2. Go to Giovanna Gamberini's Contact Record. 
3. From Giovanna Gamberini's Relationship Related List, click "New". 
-- In "Related Contact" add "Diana Gamberini"
-- In "Type" choose "Daughter"
-- Click Save
4. Ensure you're still on Giovanna's Gamberini's Contact Record.
5. From Giovanna Gamberini's Relationship Related List, click on the Relationship # just created (It will look similar to "R-000000")
-- In "Related Contact", click the Pencil to enable edit mode
-- Click the "X" to remove "Diana Gamberini" from the "Related Contact" field
-- Add "Louisa Gamberini" to the "Related Contact" field
-- Click Save
**Actual Result** : Error message is: *Rather than change the contacts in an existing relationship, delete this relationship and create a new on between the new contacts*
**Expected Result** : Error message should be: *Rather than change the contacts in an existing relationship, delete this relationship and create a new one between the new contacts.*

What is the motivation / use case for changing the behavior?
Quality!

skosgi1:
# Critical Changes

# Changes

## Updates to Relationship Objects
We've made some minor editorial changes to the error message that displays when users try to change the Related Contact on a Relationship record.

# Issues Closed
Fixes #865 

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/objectTranslations/Relationship__c-en_US.objectTranslation
+++ b/src/objectTranslations/Relationship__c-en_US.objectTranslation
@@ -168,7 +168,7 @@
     </layouts>
     <startsWith>Consonant</startsWith>
     <validationRules>
-        <errorMessage><!-- Rather than change the contacts in an existing relationship, delete this relationship and create a new on between the new contacts --></errorMessage>
+        <errorMessage><!-- Rather than change the Contacts in an existing Relationship, delete this Relationship and create a new one.--></errorMessage>
         <name>Related_Contact_Do_Not_Change</name>
     </validationRules>
 </CustomObjectTranslation>
--- a/src/objects/Relationship__c.object
+++ b/src/objects/Relationship__c.object
@@ -336,6 +336,6 @@ TEXT(Contact__r.Salutation) + &quot; &quot; + Contact__r.FirstName + &quot; &quo
         <description>Do not allow user to change Related Contact value as it damages the paired relationship data</description>
         <errorConditionFormula>and(not( ISNEW()), ISCHANGED( RelatedContact__c ))</errorConditionFormula>
         <errorDisplayField>RelatedContact__c</errorDisplayField>
-        <errorMessage>Rather than change the contacts in an existing relationship, delete this relationship and create a new on between the new contacts</errorMessage>
+        <errorMessage>Rather than change the Contacts in an existing Relationship, delete this Relationship and create a new one.</errorMessage>
     </validationRules>
 </CustomObject>


xinnanlisf:
Also, @juliesnow would you mind reviewing the release note for this PR? Thank you!

skosgi1:
> @skosgi1 The changes look good. One quick question, do we need to update translation for other languages too?
> Also, could you change the fixed issue number to "Fixes #XXX"?

Oh good catch. I am not entirely sure how to handle the translations for different languages.. I think  someone from the other team manages the translations, I will bring this to team tomorrow.

skosgi1:
> > @skosgi1 The changes look good. One quick question, do we need to update translation for other languages too?
> > Also, could you change the fixed issue number to "Fixes #XXX"?
> 
> @xinnanlisf Oh good catch. I am not entirely sure how to handle the translations for different languages.. I think someone from the other team manages the translations, I will bring this to team tomorrow.

@xinnanlisf This is good to go. We are not updating other languages. Confirmed with @lifewithryan and @robertbwatson.

xinnanlisf:
@skosgi1 since this is the translation is not accurate any more. Should we temporarily remove the related translation and put them into next round of translation work? fyi @lifewithryan

lifewithryan:
> @skosgi1 since these translations are not accurate any more, should we temporarily remove the related translation and put them into next round of translation work? fyi @lifewithryan

@xinnanlisf that's a good question, suggestion -- however I'm not sure we need to do that. Given how fast things change and are added, I think translations are always going to lag behind. I'm not sure how scaleable it is going to be if we try to manage that manually and keep track of it. However, perhaps @ceiroa might have some more insight here as well.

xinnanlisf:
> > @skosgi1 since these translations are not accurate any more, should we temporarily remove the related translation and put them into next round of translation work? fyi @lifewithryan
> 
> @xinnanlisf that's a good question, suggestion -- however I'm not sure we need to do that. Given how fast things change and are added, I think translations are always going to lag behind. I'm not sure how scaleable it is going to be if we try to manage that manually and keep track of it. However, perhaps @ceiroa might have some more insight here as well.

My only concern is that I believe our translation effort is on those metadata that haven't been translated. Correct me I am wrong for sure :)

PR #882, opened 2019-08-02
Fix nested Program robot error

Safiyab:
Just a workaround that seems to be more stable than the previous version. Might still need more investigation in the future.
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/robot/EDA/resources/EDA.py
+++ b/robot/EDA/resources/EDA.py
@@ -59,10 +59,8 @@ class EDA(object):
         locator = eda_lex_locators["tab"]
         locator = self.format_all(locator, title)
         self.selenium.capture_page_screenshot()
-        self.selenium.log_source()
         self.selenium.wait_until_page_contains_element(locator, timeout=60,
-                                                       error=title + " list header is not available on the page")
-        self.selenium.capture_page_screenshot()
+                                                       error="'" + title + "' list header is not available on the page")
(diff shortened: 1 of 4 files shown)


Safiyab:
@jstvz Can you please approve this PR ? It's not a great solution, but for the moment, I think its more stable than the current master version.

robot/EDA/resources/EDA.py
```diff
@@ -329,9 +332,22 @@ def populate_placeholder(self, loc, value):
         field = self.selenium.get_webelement(xpath)
         field.send_keys(value)
         time.sleep(1)
-#         if loc == ("Search Contacts" or "Search Accounts"):
         field.send_keys(Keys.ARROW_DOWN + Keys.ENTER)
 
+    def format_all(self, loc, value):
+        """ Formats the given locator with the value for all {} occurrences """
+        count = 0
```
jstvz: Lines 339-342 can be replaced with:

```suggestion
        count = loc.count('{')
```

Safiyab:
@jstvz Never mind this PR. I created this before I knew the issue was a mismatch between local and MetaCI org. So this solution is anyway going to fail at one of the locators. I'm closing this PR and will create another one just to mark the scripts as unstable (until we resolve this issue).

PR #884, opened 2019-08-02
EDA Behavior Involvement and Behavior Response

### LINKED ISSUE #875: Hybrid - Behavior Mgmt - Creation of Behavior Involvement and Behavior Response to handle light behavior management
Spec: https://salesforce.quip.com/QLeJAxYpxBWE
Spike: https://salesforce.quip.com/h7L7AxaR3jXM

Purpose:

Keep track of an behavior case, who was involved (handle multiple people), when it happened, what happened, where it happened, and what actions were taken by the educational institution in response to the event. 

Keep track of a student's disciplinary marks through a school year, term, or their entire duration attending that education institution. An behavior case can involve multiple parties, multiple cases, and stay with the 'Contact' for the entire duration of the term or school year or time at the educational institution. 

The "Behavior Involvement" will the the junction object between a 'Contact' and a 'Case', which allows an organization to record a case for one incident and associate multiple people and their role played in said incident.

The "Behavior Response" will be an object linked to Behavior Involvement object through look up relationship, which allows an organization to record an action/response on a behavior involvement.

xinnanlisf:
# Critical Changes

# Changes

## New Behavior Management Objects
- We've added two new objects to support your behavior management programs.
   - Behavior Involvement lets you associate individuals with a behavioral event that's tracked as a Case.
   - Behavior Response helps you track corrective actions and other responses. 
- We've also added fields to the Case object as part of tracking behavior management.
   - Category tracks the type of event, such as a code of conduct violation or other type of school violation.
   - Location tracks where the event occurred, such as on campus or off campus.
- For new EDA installations, the Behavior Involvement related list is automatically added to the Contact page layout and Case page layout (for the Incident record type). The Behavior Response related list is automatically added to the Behavior Involvement page layout. 
If your organization has an existing EDA installation, you'll want to manually add the these related lists to your page layouts to take advantage of these new objects. For information, check out [Display a New Custom Object After a Product Release](https://powerofus.force.com/s/article/EDA-Display-Cust-Obj-Post-Rel).

# Issues Closed
Fixes #875
# New Metadata
Layouts:
Behavior_Involvement__c-EDA Behavior Involvement Layout.layout
Behavior_Response__c-EDA Behavior Response Layout.layout
Case-EDA Incident Layout.layout

Translation objects:
Behavior_Involvement__c-ca.objectTranslation
Behavior_Involvement__c-en_US.objectTranslation
Behavior_Involvement__c-es.objectTranslation
Behavior_Involvement__c-fr.objectTranslation
Behavior_Response__c-ca.objectTranslation 
Behavior_Response__c-en_US.objectTranslation 
Behavior_Response__c-es.objectTranslation
Behavior_Response__c-fr.objectTranslation 

Objects:
Behavior_Involvement__c.object
Behavior_Response__c.object

BusinessProcess:
Case.Behavior Management

RecordType:
Case.Incident

Tabs:
Behavior_Involvement__c.tab
Behavior_Response__c.tab

# Deleted Metadata

# Testing Notes


src/objects/Behavior_Response__c.object
```diff
@@ -0,0 +1,193 @@
...
+        <relationshipLabel>Behavior Responses</relationshipLabel>
+        <relationshipName>Behavior_Responses</relationshipName>
+        <required>false</required>
+        <trackTrending>false</trackTrending>
+        <type>Lookup</type>
+    </fields>
+    <fields>
+        <fullName>Contact__c</fullName>
+        <externalId>false</externalId>
+        <formula>Behavior_Involvement__r.Contact__r.FirstName + &apos; &apos; +  Behavior_Involvement__r.Contact__r.LastName</formula>
```
robertbwatson: Did we ever officially decide if we're going to show a hyperlink instead?

unpackaged/config/dev/profiles/Admin.profile
```diff
@@ -75,7 +79,11 @@
         <visibility>DefaultOn</visibility>
     </tabVisibilities>
     <tabVisibilities>
-        <tab>%%%NAMESPACE%%%Time_Block__c</tab>
```
robertbwatson: Was this intentionally removed?

src/package.xml
```diff
@@ -932,6 +965,7 @@
     <types>
         <members>Attribute__c.Credential</members>
         <members>Attribute__c.Student_Characteristic</members>
+        <members>Case.Incident</members>
```
robertbwatson: I don't know exactly what our process is on introducing record types on standard objects. Similar to account, does this maybe need to be unmanaged? Not sure if all of our existing customers should get this by default or not, if they're already heavily using the Case object and don't plan on using Behavior Management...

src/package.xml
```diff
@@ -196,6 +196,10 @@
         <members>svgIcon_EVT_Press</members>
         <name>AuraDefinitionBundle</name>
     </types>
+    <types>
+        <members>Case.Behavior Management</members>
```
robertbwatson: Similar to my question below - do we want this to be packaged or should it be unmanaged?

src/objects/Behavior_Response__c.object
lifewithryan: You've beaten me to it, was going to ask the same thing. We can leverage the HYPERLINK formula function here and it will link to the contact.

src/package.xml
lifewithryan: I was going to recommend we move these to unmanaged as well for that reason. I was reviewing the spec and didn't see where we specified one way or the other. I think perhaps we want to go with unmanaged here if possible. That may mean that the page layouts associated will likely have to be unmanaged as well. (worth considering)

src/package.xml
lifewithryan: ahhh yes the business process will likely want to be unmanaged as well just in case people don't want to use it, they can delete/user their own, etc.

tasks/salesforce.py
```diff
@@ -126,6 +126,12 @@ def _process_metadata(self):
             'false',
         )
 
+        # Set record type visibilities for Case
+        self._set_record_type(
+            'Case.{managed}Incident'.format(**namespace_args),
```
lifewithryan: if/when we move the record types for case to unmanaged, remember to check this and modify if needed. I can't remember how that bit works, there's managed and namespaced, etc...we may not need the {managed} bit.

unpackaged/config/dev/profiles/Admin.profile
lifewithryan: looks like an accident

src/package.xml
robertbwatson: On the fence about the page layout being unmanaged. I think it can be managed like the account page layouts (consistency?), but I'd like to know how it gets assigned to the correct record type. 

It looks like we assign page layouts to the admin.profile in unpackaged/config/dev/profiles/Admin.profile (and perhaps elsewhere?) so we might want to do something similar here.

robertbwatson: And for existing customers they'd have to assign it themselves, presumably...

lifewithryan: yea you're correct. I'm stuck in K12 land at the moment where we can't do stuff with page layouts because the fields won't be there. Not the case (no pun intended) here :)

src/objects/Behavior_Response__c.object
xinnanlisf: Ah, for some reason this change is not committed. I will work on it!

unpackaged/config/dev/profiles/Admin.profile
xinnanlisf: This is some duplicate code I am trying to clean up.

xinnanlisf:
@lifewithryan @robertbwatson thanks for your feedback! I will go ahead and fix:
1. Contact formula field on behavior response
2. Unmanaged Case record types and business process

Let me know if I missed anything!

unpackaged/config/dev/profiles/Admin.profile
robertbwatson: Ah, I see now. Perfect!

src/package.xml
xinnanlisf: Trying to clarify here, we do want to keep case record type and page layout as managed and change the business process to unmanaged?

robertbwatson: The Case record type should be unmanaged, but the page layout can remain in the package? I think?

cumulusci.yml
```diff
@@ -212,6 +212,8 @@ plans:
                             name: EDA - Account Record Types
                         contact_key_affl_fields:
                             name: EDA - Contact Key Affiliation Fields
+                        case_behavior_record_types:
+                            name: EDA - Case Behavior Record Types
```
robertbwatson: This seems like a weird name, maybe. Perhaps "EDA - Case Incident Record Type"?

src/package.xml
```diff
@@ -757,6 +771,14 @@
         <members>Attribute__c-en_US</members>
         <members>Attribute__c-es</members>
         <members>Attribute__c-fr</members>
+        <members>Behavior_Involvement__c-ca</members>
```
robertbwatson: I think there should also be translation files for Case?

unpackaged/pre/case_behavior_record_types/package.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Package xmlns="http://soap.sforce.com/2006/04/metadata">
+  	<types>
```
robertbwatson: Minor, but it seems like the formatting got messed up a bit here.

unpackaged/pre/case_behavior_record_types/objects/Case.object
```diff
@@ -0,0 +1,30 @@
...
+            <fullName>New</fullName>
+            <default>true</default>
+        </values>
+        <values>
+            <fullName>Working</fullName>
+            <default>false</default>
+        </values>
+    </businessProcesses>
+    <recordTypes>
+        <fullName>Incident</fullName>
```
robertbwatson: When this was moved out of the package, it looks like the Location and Category picklist values available to the record type got removed. Should they be added back here? (See https://github.com/SalesforceFoundation/EDA/pull/884/commits/99c930658857aa3187ad99bf5bd4617fecdbc419#diff-e0dc43ce6e02eabdfe8f745916f6385bL89)

cumulusci.yml
xinnanlisf: Well Incident is one of the record types and I am wondering if we "might" (although not likely) add more unmanaged record types to this file?

unpackaged/pre/case_behavior_record_types/objects/Case.object
xinnanlisf: I am wondering if we should leave those picklist values in src/objects/Case.object?

unpackaged/pre/case_behavior_record_types/package.xml
xinnanlisf: Should be good now!

src/package.xml
xinnanlisf: Added.

robertbwatson: @xinnanlisf I think they need to be added to the package.xml too?

unpackaged/pre/case_behavior_record_types/objects/Case.object
robertbwatson: @xinnanlisf @lifewithryan I'm talking about the picklist values available for each record type that was removed in this commit: https://github.com/SalesforceFoundation/EDA/commit/99c930658857aa3187ad99bf5bd4617fecdbc419#diff-e0dc43ce6e02eabdfe8f745916f6385bL89 -- the picklist values still exist in src/objects/Case.object, but we've lost the record type assignment.

lifewithryan: ooooohhhhh, gotcha good catch @robertbwatson -- yea we should probably specify the the picklist values in the recordtype metadata as well, otherwise they won't be available to said recordtype -- or maybe they're available for all because I did see them in my test.

xinnanlisf:
Should be good now! @lifewithryan

unpackaged/pre/case_behavior_record_types/objects/Case.object
xinnanlisf: We want those picklist values only unmanaged? We can't have them both in managed and unmanaged otherwise the build will fail..

src/package.xml
lifewithryan: correct and I apologize if I've steered you wrong somewhere. Balancing the odd unmanaged/managed K12 stuff and the EDA stuff is messing with my head. For this feature (which is on EDA) Record Types should be unmanaged and page layouts can be managed.

unpackaged/pre/case_behavior_record_types/objects/Case.object
lifewithryan: I thought we could have them be managed but the unmanaged bit just assigns the values to the correct record type? Meaning in src/objects/Case.object we have the "field" and the valueset defined, but in unpackaged/.../recordtypes/Case.object we have the recordType metadata and the values listed for that? Are we saying that makes the build fail? Could have sworn we did that elsewhere.

xinnanlisf: Yeah once a record type is created all the picklist fields assigned will be moved under the new record type.

robertbwatson: @xinnanlisf  @lifewithryan  The build failure: "Update of RecordType Case.Incident: Error on line 29, col 25: Global value set 'Category__c' in record type 'Incident' on the entity 'Case' cannot be resolved" suggests that it's looking for a global value set. I think this is because this is in /pre/ - so it's running *before* the fields actually exist on the Case object.

robertbwatson: The record type picklist assignment probably needs to happen in /post/

unpackaged/pre/case_behavior_record_types/objects/Case.object
```diff
@@ -0,0 +1,144 @@
...
+                <fullName>Off campus</fullName>
+                <default>false</default>
+            </values>
+            <values>
+                <fullName>On campus</fullName>
+                <default>false</default>
+            </values>
+        </picklistValues>
+        <picklistValues>
+            <picklist>Origin</picklist>
```
robertbwatson: Probably only need Category & Location above and not the rest of these?

xinnanlisf:
@lifewithryan @robertbwatson should be good to go!

cumulusci.yml
```diff
@@ -221,5 +221,7 @@ plans:
                         name: Course Connection Record Types for EDA
                     facility_display_name:
                         name: Facility Display Name Formula Field
+                    case_behavior_record_types:
+                            name: EDA - Case Behavior Record Types
```
robertbwatson: Super minor, but now that we've moved this to the post-install, it bothers me that "Course Connection Record Types for EDA" is inconsistent with our naming for all other record types...

robertbwatson:
@xinnanlisf  Great work on this! I've approved but kept this assigned to @lifewithryan  since he also requested changes.

xinnanlisf:
Hi @lianeyee we might need your help on the release note :)

cumulusci.yml
```suggestion
                        name: EDA - Case Behavior Record Types
```

cumulusci.yml
```diff
@@ -221,5 +221,7 @@ plans:
                         name: Course Connection Record Types for EDA
                     facility_display_name:
                         name: Facility Display Name Formula Field
+                    case_behavior_record_types:
```
robertbwatson: Super minor, but now that we've moved this to the post-install, it bothers me that "Course Connection Record Types for EDA" is inconsistent with our naming for all other record types...

jstvz: I'm okay with renaming it in this PR if there's consensus on that. Otherwise we can revisit this next week.

lianeyee: Are we talking about changing it to "EDA - Course Connection Record Types" to make it consistent? If so, I'm in favor.

cumulusci.yml
```diff
@@ -218,8 +218,10 @@ plans:
                 task: deploy_post
                 ui_options:
                     course_connection_record_types:
-                        name: Course Connection Record Types for EDA
+                        name: EDA - Course Connection Record Types for EDA
```
```suggestion
                        name: EDA - Course Connection Record Types
```

src/package.xml
```diff
@@ -203,6 +203,8 @@
         <members>Account.HEDA_Account_Compact_Layout</members>
         <members>Address__c.HEDA_Address_Compact_Layout</members>
         <members>Affiliation__c.HEDA_Affiliation_Compact_Layout</members>
+        <members>Behavior_Involvement__c.Behavior_Involvement_Compact_Layout</members>
```
robertbwatson: @xinnanlisf Should we fix the naming here to be consistent? "EDA_"?

lifewithryan: Since these are on legacy objects I'm gonna say we leave it. Pretty sure we decided that only net new page layouts would get the EDA prefix.

ceiroa: Well, the Behavior one seems new. I'd consider giving that one a prefix. The others can't really be changed at this point, in case we were thinking moving from HEDA to EDA, since that would create brand new ones.

lifewithryan: Oh yea, I was looking at the HEDA stuff and made an assumption...yes we should add the EDA prefix to the Behavior one...my coffee isn't working apparently.

ceiroa: No worries. I had the same initial thought when I got the email notification, since you can't see the colors there. Only when I came to Github I realized. :-)

robertbwatson: Sorry, I'll be more clear next time!

cc @xinnanlisf in case you didn't see this :)

xinnanlisf: should be good now!

dgamboni:
This passes QA; pending PM/UAT review.

robertbwatson:
@xinnanlisf  Minor, but the latest round of description updates (https://github.com/SalesforceFoundation/EDA/pull/884/commits/d5f72636e48f0fef33430b8fec4e4dbf9b585504) didn't include the translation updates

xinnanlisf:
@robertbwatson should be good now!

PR #890, opened 2019-08-09
Account Look up Auto-Population

### LINKED ISSUE #557: When we remove the Account lookup on a Contact, system should fill in the Account lookup if it can find an Account that Primary Contact = Current Contact
1. Create a Contact with either Admin or HH account model. System will automatically create an Account for the Contact.  And mark current Contact as the Primary Contact of the new Created Account.
2. Now, try to remove the Account lookup of the Contact and click Save. You will see that now the Account Lookup is empty of the Current Contact. 
3. The behavior is wrong here since every Contact must associate with an Account. 
4. The correct behavior is that system should fill in the Account lookup if it can find an Account that Primary Contact = Current Contact

skosgi1:
# Critical Changes

# Changes

## General Updates
We've fixed an issue that occurs when the Account lookup on a Contact is cleared. With this fix, the lookup is repopulated with the Account (if the Account is the default record type), or is populated with a new Account record of the default record type.

# Issues Closed
Fixes #557 

# New Metadata

# Deleted Metadata

# Testing Notes

- Create a contact. The Account lookup is auto-populated with the Account with default account model record type.
- Clear the Account lookup field. The system should re-populate the Account lookup with the same Account.

- Create a contact and manually populate the Account lookup to an Account that is *not* the default account model record type.
- Clear the Account lookup field. The system should create a new Account of the default account model record type and populate the Account lookup with the new Account.


src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -760,12 +763,14 @@ private class ACCT_IndividualAccounts_TEST {
...
         Test.startTest();
-        con.AccountId = null;
+        con.AccountId = NULL;
         update con;
         Test.stopTest();
 
-        Account assertAccount = [Select Primary_Contact__c FROM Account LIMIT 1];
-        system.assertEquals(null, assertAccount.Primary_Contact__c);
+        Account assertAccount = [SELECT Primary_Contact__c FROM Account LIMIT 1];
+        Contact[] relatedCOntact = [SELECT AccountId FROM Contact];
```
robertbwatson: Minor -- I believe we prefer <List>'s over [] notation. Consider updating all references in this class to use Lists

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -760,12 +763,14 @@ private class ACCT_IndividualAccounts_TEST {
...
+        System.assertNotEquals(NULL, assertAccount.Primary_Contact__c);
```
robertbwatson: We should know for both of these asserts what the expected value is, correct?

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -921,4 +926,35 @@ private class ACCT_IndividualAccounts_TEST {
...
+
+        con.AccountId = NULL;
+        Test.startTest();
+        update con;
+        Test.stopTest();
+
+        Contact[] updatedContacts = [SELECT Account.Name, AccountId FROM Contact where id=:con.id];
+
+        //the account should not be the one we started with
+        System.assertNotEquals(acct.id, updatedContacts[0].AccountId);
```
robertbwatson: We should also assert that it's not null, and possibly assert that the account's RecordTypeId = UTIL_Describe_API.getAdminAccRecTypeID().

robertbwatson:
@skosgi1 I made a few changes to the release notes to highlight the two scenarios that happen here (re-populating the lookup or creating a new account & populating with the new value) - can you look it over and let me know if it makes sense?

src/classes/ACCT_IndividualAccounts_TEST.cls
skosgi1: Hmm we are checking if the ids are equal so no need for assert that it is not null, right? The contact/account is not deleted. Same reason for recordtype, the ids are same any way and there is no code that changes the recordtype.

skosgi1:
> @skosgi1 I made a few changes to the release notes to highlight the two scenarios that happen here (re-populating the lookup or creating a new account & populating with the new value) - can you look it over and let me know if it makes sense?

@robertbwatson Looks much better. Thank you.

src/classes/ACCT_IndividualAccounts_TEST.cls
skosgi1: oops discard this please, I was looking at the other method.

skosgi1:
@robertbwatson I have addressed the changes, can you please give it another shot?

skosgi1:
> @skosgi1 Look good! I'm going to assign to Dina for when she's back next week.

Thank you!

PR #892, opened 2019-08-14
add .vscode settings to gitignore

Julian88Tex:
Added VSCode folder to .gitignore so it is ignored by git.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


Julian88Tex:
@robertbwatson hopefully I'm following the right procedure but thought I'd put this PR in for anyone who's using VScode

robertbwatson:
> @robertbwatson hopefully I'm following the right procedure but thought I'd put this PR in for anyone who's using VScode

@Julian88Tex I noticed there are some Robot changes here as well - is that intentional? If so, can you work with Doug on getting those into EDA? I'd prefer this PR to just be for the .gitignore.

Also, thanks for doing this! I've been using VS Code since I've been here and not sure why I've never done this...

lifewithryan:
> > @robertbwatson hopefully I'm following the right procedure but thought I'd put this PR in for anyone who's using VScode
> 
> @Julian88Tex I noticed there are some Robot changes here as well - is that intentional? If so, can you work with Doug on getting those into EDA? I'd prefer this PR to just be for the .gitignore.
> 
> Also, thanks for doing this! I've been using VS Code since I've been here and not sure why I've never done this...

@robertbwatson @Julian88Tex -- agreed, it's never showed up for me as a file to be added until just recently. Thanks for doing this @Julian88Tex. Also I agree, lets just keep this change to .gitignore here.

Julian88Tex:
@robertbwatson @lifewithryan Thanks so much for checking! Sorry about adding in that robots test. I am still getting used to working with forks and merging back in. I remember checking the PR to make sure it didn't get added to it but I think there was just a longer delay that I wasn't expecting.

Yeah, I'm not sure if I had a vscode setting that just adds this file into every repo, but it was getting on my nerves.

I've updated PR now so it should just be that one ignore change.

.gitignore
```diff
@@ -9,6 +9,7 @@ build.properties
 salesforce.schema
 Referenced Packages
 /.sfdx
+**/.vscode/
```
lifewithryan: @Julian88Tex  -- what is this ** thing I'm seeing here. I've not seen that before.

.vscode/ should be enough I think

lifewithryan: or is this just in case .vscode appears in subdirectories as well? (even so, in that case I'd think one * would do it...but I could be wrong.

Julian88Tex: So not positive but I believe it's all directories. That being said I literally copy and pasted from Cumulus.

Julian88Tex: So not positive but I believe it's all directories. That being said I literally copy and pasted from Cumulus.

lifewithryan: no worries. assuming it work (i'll test locally here but sounds good to me)

robertbwatson:
@lifewithryan should this go to Dina or are we okay to just merge this? I'm good if you've tested it locally.

PR #926, opened 2019-09-12
W-035996 Updating the Case "Behavior Management" business process

robertbwatson:
We updated the Case "Behavior Management" business process so that we can succesfully deploy EDA to Enterprise Orgs.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/orgs/beta.json
+++ b/orgs/beta.json
@@ -1,6 +1,6 @@
 {
   "orgName": "EDA - Beta Test Org",
-  "edition": "Developer",
+  "edition": "Enterprise",
   "settings" : {
     "orgPreferenceSettings": {
         "s1DesktopEnabled": true,
--- a/src/package.xml
+++ b/src/package.xml
@@ -203,8 +203,6 @@
         <members>Account.HEDA_Account_Compact_Layout</members>
         <members>Address__c.HEDA_Address_Compact_Layout</members>
         <members>Affiliation__c.HEDA_Affiliation_Compact_Layout</members>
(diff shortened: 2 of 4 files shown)


robertbwatson:
@juliesnow  Note I didn't add release notes here since this is fixing a bug with a feature that we're also releasing in EDA 1.83 (#884) -- however, I do think we need to revisit the release notes for Behavior Management to make sure we call out that existing customers will need to manually add this new support process and record type if they want to use the "Incident" record type on the Case.

orgs/beta.json
```diff
@@ -1,6 +1,6 @@
 {
   "orgName": "EDA - Beta Test Org",
-  "edition": "Developer",
+  "edition": "Enterprise",
```
robertbwatson: I also updated the beta and release org configurations to be Enterprise orgs, per @jstvz suggestion.

src/package.xml
```diff
@@ -203,8 +203,6 @@
         <members>Account.HEDA_Account_Compact_Layout</members>
         <members>Address__c.HEDA_Address_Compact_Layout</members>
         <members>Affiliation__c.HEDA_Affiliation_Compact_Layout</members>
-        <members>Behavior_Involvement__c.EDA_Behavior_Involvement_Compact_Layout</members>
```
robertbwatson: This was done automatically... I think maybe they were previously removed but didn't get removed from the package.xml?

PR #930, opened 2019-09-13
Fixed build failure due to missing GB translation files

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


xinnanlisf:
@ryanguest this is the fix. As you can see, these files are just placeholders and not translated yet. We can have them translated in next round.

PR #936, opened 2019-09-19
Prevent Term Deletion

### LINKED ISSUE #905: Block Deletion of Terms when Term Grades Exist

Feature 

Terms can be deleted even if there are Term Grade records associated with that Term.

Terms should not be able to be deleted if Term Grade records exist.

Data integrity

skosgi1:
# Critical Changes

# Changes
We've added a new trigger handler on Term (TERM_CannotDelete_TDTM) that prevents a Tern record from being deleted if any Term Grade records are associated with it.
 
# Issues Closed
Closes #905

# New Metadata

- TERM_CannotDelete_TDTM

- TERM_CannotDelete_TEST

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/TERM_CannotDelete_TDTM.cls
+++ b/src/classes/TERM_CannotDelete_TDTM.cls
@@ -37,8 +37,8 @@
 public with sharing class TERM_CannotDelete_TDTM extends TDTM_Runnable {
     /*******************************************************************************************************
     * @description Prevents Term records from being deleted if it has any Term Grade associations.   
-    * @param listNew the list of Terms from trigger new. 
-    * @param listOld the list of Terms from trigger old. 
+    * @param newList the list of Terms from trigger new. 
+    * @param oldList the list of Terms from trigger old. 
     * @param triggerAction which trigger event (BeforeInsert, AfterInsert, etc.).  
     * @param objResult the describe for Terms
     * @return dmlWrapper.  
--- a/src/classes/TERM_CannotDelete_TDTM.cls
+++ b/src/classes/TERM_CannotDelete_TDTM.cls
@@ -1,0 +1,61 @@
+/*
+    Copyright (c) 2019, Salesforce.org
+    All rights reserved.
+    
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+    
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of Salesforce.org nor the names of
+      its contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+ 
+    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT 
+    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS 
+    FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE 
+    COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, 
+    INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, 
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; 
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER 
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT 
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN 
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE 
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Terms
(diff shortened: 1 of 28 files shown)


src/classes/TERM_CannotDelete_TDTM.cls
```diff
@@ -0,0 +1,61 @@
...
+* @author Salesforce.org
+* @date 2019
+* @group Terms
+* @group-content ../../ApexDocContent/Terms.htm
+* @description This class prevents Term records from being deleted if it has any Term Grade associations. 
+*/
+public with sharing class TERM_CannotDelete_TDTM extends TDTM_Runnable {
+    /*******************************************************************************************************
+    * @description Prevents Term records from being deleted if it has any Term Grade associations.   
+    * @param listNew the list of Terms from trigger new.
```
robertbwatson: I noticed this before in other files, so there must have been a bad copy/paste started a long time ago. But these params are actually "newList" and "oldList" below.

src/classes/TERM_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,102 @@
...
+
+    @isTest 
+    static void testDeleteTermsWithTermGrades() {
+        
+        // Insert Course Offering
+        Course_Offering__c courseOffering = UTIL_UnitTestData_TEST.createCourseOffering(NULL, NULL);
+        
+        // Insert Contact
+        Contact student = UTIL_UnitTestData_TEST.getContact();
+        student.LastName = 'Term Grade';
```
robertbwatson: I don't think we need lines 62-64 -- it looks like getContact() already specifies these fields? Since we only have one contact in this unit test, that should be fine.

src/classes/TERM_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,102 @@
...
+        insert term;
+    }
+    
+    // Test deletion of Terms with Term Grades
+
+    @isTest 
+    static void testDeleteTermsWithTermGrades() {
+        
+        // Insert Course Offering
+        Course_Offering__c courseOffering = UTIL_UnitTestData_TEST.createCourseOffering(NULL, NULL);
```
robertbwatson: Can we pass in the Term ID here? Otherwise we have data integrity issues as another Term would be created for this.

src/classes/TERM_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,102 @@
...
+        insert student;
+        
+        // Insert Course Connection
+        Course_Enrollment__c courseConnection = UTIL_UnitTestData_TEST.getCourseConnection(student.Id, courseOffering.Id);
+        insert courseConnection;
+        
+        Term__c term = [SELECT Id FROM Term__c LIMIT 1];
+        
+        // Insert Term Grade
+        Term_Grade__c termGrade = new Term_Grade__c();
```
robertbwatson: Can we add a new getTermGrade() method to UTIL_UnitTestData_TEST that accepts these two parameters?

(Actually, it looks like we're adding a getTermGrade() method in one of the other child PRs, but that one has different parameters so I think having both methods would be fine.)

src/labels/CustomLabels.labels
```diff
@@ -56,6 +56,14 @@
         <shortDescription>CannotDelete</shortDescription>
         <value>Record cannot be deleted because children records exist. Delete children first.</value>
     </labels>
+    <labels>
```
robertbwatson: Was this verified by Julie? I'm curious why we're adding a more specific label here and not using the "CannotDelete" label. I think being specific is better, but it's not consistent with what we've done in the past so just want to make sure everyone is in agreement this is what we want to do here.

@lifewithryan thoughts?

skosgi1:
@robertbwatson  update the code, can you please take another quick look at it? Thank you.

src/classes/TERM_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,97 @@
...
+        Course_Enrollment__c courseConnection = UTIL_UnitTestData_TEST.getCourseConnection(student.Id, courseOffering.Id);
+        insert courseConnection;
+        
+        // Insert Term Grade
+        Term_Grade__c termGrade = UTIL_UnitTestData_TEST.getTermGradeWTermCourseConn(term.Id, courseConnection.Id);
+        insert termGrade;
+        
+        try {
+            Test.startTest();
+                delete term;
```
robertbwatson: Just wanted to say I've never actually seen anyone add indentation between Test.startTest() and Test.stopTest() before but I actually really like this idea. It makes it much easier to see what's being done during the separate test context.

src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -440,6 +440,32 @@ public class UTIL_UnitTestData_TEST {
         return contactLanguageToInsert;
     }
 
+    /*********************************************************************************************************
+    * @description Gets a Term for testing.
+    * @param accountId Account to assocaite Term with.
```
robertbwatson: Minor misspelling, "associate"

src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -440,6 +440,32 @@ public class UTIL_UnitTestData_TEST {
...
+    public static Term__c getTerm(Id accountID, String termName) {
+        Term__c term = new Term__c();
+        term.Account__c = accountID;
+        term.Name = termName;
+        return term;
+    }
+
+    /*********************************************************************************************************
+    * @description Gets a TermGrade with Term and Course Connection for testing.
+    * @param termID Account to assocaite Term Grade with.
```
robertbwatson: Minor misspelling, "associate", and it's actually the Term, not the Account

src/labels/CustomLabels.labels
robertbwatson: @skosgi1 You removed the label everywhere but forgot to do it here :)

skosgi1:
> @skosgi1 More minor word changes and you forgot to remove the label from the actual labels file :)
> 
> Otherwise this is great and I'm approving it. Feel free to assign and move to Dina once you make the word edits and remove the label.

src/labels/CustomLabels.labels
skosgi1: @robertbwatson  😂

aligeez:
Closing this for now until we reach a decision on a EDA Settings option for prevent deletion logic.

PR #944, opened 2019-09-26
W-035967 - Update descriptions in cumulusci.yml and add additional qa metadata & configuration

xinnanlisf:
This internal task adds page layouts, search layouts, and methods to enable custom settings for the EDA qa_org flow. This enables our QE team to use specific customizations when spinning up a testing org. This task also organizes the related descriptions in the cumulusci.yml file. 

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


cumulusci.yml
```diff
@@ -87,6 +87,7 @@ tasks:
             output: robot/EDA/doc/Keywords.html
 
     update_admin_profile:
+        description: Sets record type visibilities for System Administrative profile
```
jstvz: I'd like to not overwrite the built-in description. Can we add to it instead? Suggested:

```suggestion
        description: Retrieves, edits, and redeploys the System Administrator profile with full FLS perms for all objects/fields. Also sets record type visibilities.
        group: Custom Tasks
```

Also adding `group:` here to make clear that we've overridden the default task.

cumulusci.yml
```diff
@@ -126,37 +131,25 @@ flows:
...
-                    namespaced_org: True
-                    skip_record_types: True
-
     config_qa:
+        descripton: Configure an org for use as a QA org after package metadata is deployed
         steps:
             3:
                 task: execute_install_apex
             4:
-                task: deploy_dev_config
```
xinnanlisf: @jstvz do you think it is ok for us to remove this line? config_qa should only deploy qa config?

cumulusci.yml
```diff
@@ -37,8 +37,8 @@ tasks:
         options:
             path: unpackaged/config/dev_delete
 
-    execute_install_apex:
-        description: Runs most of the install script methods from STG_InstallScript class
+    execute_dev_apex:
```
xinnanlisf: @jstvz I renamed it to dev_apex since it feels like it is running apex for dev orgs. We can change it back if you prefer install_apex thou.

cumulusci.yml
```diff
@@ -167,21 +162,6 @@ flows:
                 options:
                     unmanaged: False
 
-    deploy_unmanaged:
```
xinnanlisf: @jstvz i have removed these flows too since we are not adding tasks to them.

scripts/setup.cls
```diff
@@ -1,17 +1,46 @@
 public static void qaSetup() {
-  enableCourseConnections();
+    configQACustomSettings();
```
xinnanlisf: @robertbwatson not sure if you are doing the same work to enable settings for dev/qa configs. but fyi..

robertbwatson: @xinnanlisf  No worries - I modified the execute_install_apex task (or execute_dev_apex) in #945 so should be fine here.

cumulusci.yml
robertbwatson: It is doing most of the install script though, of a new install, so the old name does sort of make sense. It's doing lines 44-63 of STG_InstallScript.cls, minus the scheduling of the scheduled jobs and running the batch email script.

(I don't have a strong opinion either way)

cumulusci.yml
```diff
@@ -167,21 +162,6 @@ flows:
...
-        steps:
-            4:
-                # Disable uninstall_packaged_incremental
-                task: None
-
-    dev_org_namespaced:
```
robertbwatson: Wait - I use this one all of the time. This is the flow we call to set up a scratch org with a namespace. I don't think all of the steps are defined here because it's a core CCI flow.

cumulusci.yml
xinnanlisf: Fair enough, qa config is also running this thus I don't think it makes sense to call it dev_apex either...Ok I will change it back.

cumulusci.yml
xinnanlisf: Right, many of the flows in this yml file is basically just adding more tasks or simply overriding. For example, dev_org does not even exist in this yml file but it is defined in sfdo tooling yml file.

robertbwatson: But why is this being removed? We're modifying step 3 to make sure to pass in namespaced_org: true to the update_admin_profile task - does that not matter? (I don't quite understand this but I also don't want the flow to suddenly break)

xinnanlisf: I see so it is overriding the options for #3 task. I will add them back then!

robertbwatson: I think so but I don't know for certain... would be something that @jstvz could help explain :)

xinnanlisf:
@robertbwatson this PR should be ready for review. Since you are ready looking into this, would you mind taking a review?

Also, @jstvz I added a few things after you approved this PR. Would you mind giving a sanity check too?

Thanks!

cumulusci.yml
jstvz: There's some overlap here, but I lean towards deploying it as part of the QA config. This is for two reasons: first, because new schema will often be updated here first, and second, to avoid having to duplicate sets of metadata.

That said, why did you want to remove it?

cumulusci.yml
```diff
@@ -126,37 +133,25 @@ flows:
...
                     unmanaged: False
 
-    config_packaging:
-        steps:
-            1:
-                # Configure update_admin_profile
-                options:
-                    managed: True
-                    namespaced_org: True
-                    skip_record_types: True
```
jstvz: Can we restore this block, please? Recall that we're overriding `update_admin_profile` with a custom task that includes the `skip_record_types` option.

cumulusci.yml
```diff
@@ -126,37 +133,25 @@ flows:
                 task: deploy_dev_config
 
     config_managed:
+        description: Configure an org for use as a dev org after package metadata is deployed.
         steps:
-            2:
-                # Configure update_admin_profile
-                options:
-                    managed: True
```
jstvz: This one is okay to remove, since it's duplicating the built-in option.

cumulusci.yml
```diff
@@ -48,69 +48,76 @@ tasks:
...
                 task: uninstall_packaged
 
     config_apextest:
+        description: Configure an org to run apex tests after package metadata is deployed.
         steps:
             3:
                 task: execute_install_apex
 
     config_dev:
+        description: Configure an org for use as a dev org after package metadata is deployed.
```
jstvz: I'm guessing that you're adding descriptions here so that the configuration file is easier to understand. I think that makes sense, but should we consider that by doing so we run the risk of them going out of sync? Perhaps for those built-in tasks/flows that you're adding descriptions for we could add comments instead?

robot/EDA/resources/EDA.py
```diff
@@ -1,12 +1,16 @@
 import logging
 import time
+import datetime
```
jstvz: These imports are unused. Others below on L13 are duplicated. Unless you intended to update these files, they should be reverted to the version on `master`.

robot/EDA/tests/browser/pe/program_enrollments.robot
```diff
@@ -2,6 +2,7 @@
 
 Resource        robot/EDA/resources/EDA.robot
 Library         robot/EDA/resources/EDA.py
+Library         SeleniumLibrary
```
jstvz: This change is unrelated and should be included in a separate PR.

scripts/setup.cls
```diff
@@ -1,17 +1,46 @@
...
+    settings.put('%%%NAMESPACE%%%Enable_Course_Connections__c', true);
+    settings.put('%%%NAMESPACE%%%Student_RecType__c', rtInfosByName.get('Student').getRecordTypeId());
+    settings.put('%%%NAMESPACE%%%Faculty_RecType__c', rtInfosByName.get('Faculty').getRecordTypeId());
+}
 
-  %%%NAMESPACE%%%Hierarchy_Settings__c settings = %%%NAMESPACE%%%Hierarchy_Settings__c.getOrgDefaults();
+//Enable Specify Role For Created Affiliations
+private static void enableSpecifyRoleForCreatedAffiliations(%%%NAMESPACE%%%Hierarchy_Settings__c settings) {   
+    if (settings.%%%NAMESPACE%%%Affl_ProgEnroll_Set_Role__c == null 
+        || settings.%%%NAMESPACE%%%Enable_New_Preferred_Phone_Sync__c != true) {
```
jstvz: This is a different field from the null check in this conditional. Is that intentional?

cumulusci.yml
xinnanlisf: @jstvz that makes a lot of sense. I was under the impression that dev config is only used for dev org and qa org should use its own config. I will go ahead and add it back.

robot/EDA/resources/EDA.py
xinnanlisf: Yeah i think these changes are due to relEng fix on robot tests. I will remove them. Thanks!

cumulusci.yml
xinnanlisf: That is a great idea!

cumulusci.yml
```diff
@@ -38,7 +38,7 @@ tasks:
             path: unpackaged/config/dev_delete
 
     execute_install_apex:
-        description: Runs most of the install script methods from STG_InstallScript class
+        # description: Runs anonymous apex for dev to config settings, insert Affiliation mappings, insert Relationship lookups and insert default tdtm trigger handlers.
```
```suggestion
        description: Runs anonymous apex for dev to config settings, insert Affiliation mappings, insert Relationship lookups and insert default tdtm trigger handlers.
```

cumulusci.yml
```diff
@@ -48,69 +48,76 @@ tasks:
                 TDTM_Global_API.setTdtmConfig( TDTM_Global_API.getTdtmConfig() );
 
     execute_qa_apex:
-        description: Runs Apex setup for QA orgs.
+        # description: Runs anonymous apex for qa to enable course connections.
```
```suggestion
        description: Runs anonymous apex for qa to enable course connections.
```

cumulusci.yml
```diff
@@ -146,17 +150,17 @@ flows:
                     skip_record_types: True
 
     config_qa:
+        # descripton: Configure an org for use as a QA org after package metadata is deployed.
         steps:
             3:
                 task: execute_install_apex
             4:
-                task: deploy_dev_config
```
jstvz: I don't think this made it back in yet

cumulusci.yml
robertbwatson: I think the proposed new description here is invalid. It does other things now besides enabling course connections, right?

unpackaged/config/qa/objects/___NAMESPACE___Course_Offering__c.object
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
+    <searchLayouts>
+        <searchResultsAdditionalFields>Course__c</searchResultsAdditionalFields>
```
robertbwatson: Do these need the %%%namespace%%%?

unpackaged/config/qa/objects/___NAMESPACE___Plan_Requirement__c.object
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
+    <searchLayouts>
+        <searchResultsAdditionalFields>Plan_Requirement__c</searchResultsAdditionalFields>
```
robertbwatson: Do these need the %%%namespace%%%?

unpackaged/config/qa/objects/___NAMESPACE___Term_Grade__c.object
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
+    <searchLayouts>
+        <searchResultsAdditionalFields>Course_Connection__c</searchResultsAdditionalFields>
```
robertbwatson: Do these need the %%%namespace%%%?

robertbwatson:
@xinnanlisf  Also, can you add a description to this PR? It can go above Critical Changes so that it doesn't get into the release notes, but will be good for our reference later.

xinnanlisf:
@jstvz @robertbwatson should be good now!

cumulusci.yml
```diff
@@ -48,69 +48,76 @@ tasks:
...
 
     robot_libdoc:
+        # description: Generates documentation for project keyword files.
         options:
             path: robot/EDA/resources/EDA.py,robot/EDA/resources/EDA.robot
             output: robot/EDA/doc/Keywords.html
 
     update_admin_profile:
+        description: Retrieves, edits, and redeploys the System Administrator profile with full FLS perms for all objects/fields. Also sets record type visibilities.
+        group: Custom Tasks
```
robertbwatson: This is already listed under the "Salesforce Metadata" group - are we sure we want to move it to a new group?

cumulusci.yml
```diff
@@ -48,69 +48,76 @@ tasks:
...
         class_path: tasks.salesforce.UpdateAdminProfile
```
robertbwatson: Also, this isn't a custom task, I don't think - should we actually be overriding the description here? I think we're just overriding the default behavior in tasks/salesforce.py

unpackaged/config/qa/objects/Contact.object
```diff
@@ -19,4 +19,27 @@
...
+        <lookupPhoneDialogsAdditionalFields>CONTACT.PHONE3</lookupPhoneDialogsAdditionalFields>
+        <lookupPhoneDialogsAdditionalFields>CONTACT.PHONE4</lookupPhoneDialogsAdditionalFields>
+        <lookupPhoneDialogsAdditionalFields>CONTACT.PHONE5</lookupPhoneDialogsAdditionalFields>
+        <lookupPhoneDialogsAdditionalFields>CONTACT.PHONE6</lookupPhoneDialogsAdditionalFields>
+        <lookupPhoneDialogsAdditionalFields>WorkPhone__c</lookupPhoneDialogsAdditionalFields>
+        <searchResultsAdditionalFields>FULL_NAME</searchResultsAdditionalFields>
+        <searchResultsAdditionalFields>ACCOUNT.NAME</searchResultsAdditionalFields>
+        <searchResultsAdditionalFields>CONTACT.EMAIL</searchResultsAdditionalFields>
+        <searchResultsAdditionalFields>Primary_Academic_Program__c</searchResultsAdditionalFields>
+        <searchResultsAdditionalFields>Primary_Language__c</searchResultsAdditionalFields>
```
robertbwatson: Here's another one missing %%%NAMESPACE%%%, I think -- this is packaged (Primary_Academic_Program__c is unpackaged, though)

unpackaged/config/qa/objects/___NAMESPACE___Course_Offering__c.object
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
+    <searchLayouts>
+        <searchResultsAdditionalFields>%%%namespace%%%Course__c</searchResultsAdditionalFields>
```
robertbwatson: @xinnanlisf @jstvz For this & the other files - does casing matter? (Does it need to be %%%NAMESPACE%%%?)

davisagli: It does need to be in all caps -- the code that replaces the tokens is here: https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/utils.py#L184

cumulusci.yml
xinnanlisf: I think this was @jstvz suggested?

cumulusci.yml
xinnanlisf: @jstvz too.

unpackaged/config/qa/objects/___NAMESPACE___Course_Offering__c.object
xinnanlisf: Thanks @davisagli! We will just capitalize them so that they look consistent!

cumulusci.yml
jstvz: Yeah, I suggested this to make it clearer that we're overriding the built-in task with a custom one.

cumulusci.yml
jstvz: Yeah, my view is that if any task depends on a python class in `tasks/` then it's a custom task and should be highlighted as such. I think it's helpful to keep that in mind when doing any troubleshooting or debugging.

cumulusci.yml
robertbwatson: Thanks for the explanation!

xinnanlisf:
@jstvz i think this PR is ready. Let me know if any other suggested changes!

cumulusci.yml
```diff
@@ -49,69 +49,76 @@ tasks:
...
 
     robot_libdoc:
+        # description: Generates documentation for project keyword files.
         options:
             path: robot/EDA/resources/EDA.py,robot/EDA/resources/EDA.robot
             output: robot/EDA/doc/Keywords.html
 
     update_admin_profile:
+        description: Retrieves, edits, and redeploys the System Administrator profile with full FLS perms for all objects/fields. Also sets record type visibilities.
+        group: Custom Tasks
```
```suggestion
        group: "EDA: Custom Tasks"
```

dgamboni:
@juliesnow This is an internal task only; are these release notes OK?

spinkelman:
Merging per @dgamboni.

PR #960, opened 2019-10-10
Adding triggers for new objects.

### LINKED ISSUE #959: Add triggers for recently added objects
We need triggers for:

- Behavior Involvement
- Behavior Response
- Case
- Attribute
- Language (seems like one already exists for ContactLanguage)
- Facility

The goal is that all orgs with EDA follow the best practice of a single trigger per object. We want to provide them so customers use ours instead of theirs if the write TDTM classes.

We need to consider that some customers may have already created their own triggers for some of these objects. I *think* (fairly certain) this should be fine, but we need to think it through.

ceiroa:
# Critical Changes

# Changes
- We've added triggers for the following objects: Attribute, Behavior Involvement, Behavior Response, Case, Facility and Language. Use them, as needed, to allow other packages and custom code in your org to plug TDTM classes into EDA's TDTM Framework without needing to provide your own triggers.

# Issues Closed
Fixes #959 

# New Metadata
- Triggers:
  - Attribute
  - Behavior Involvement
  - Behavior Response
  - Case
  - Facility
  - Language

# Deleted Metadata

# Testing Notes
No manual testing necessary. As far as the build passes the PR should be good. We are just adding triggers for other packages and custom code to be able to plug into the TDTM framework, but there is no functionality currently running for any of these in EDA.


ceiroa:
I still need to get the PR text reviewed by Julie, but feel free to look at the code whenever you have the chance @robertbwatson. Thanks!

ceiroa:
@robertbwatson Actually, I see that you already have the Account Naming Refactoring assigned to you for review, so I'm going to change this to Xinnan.

xinnanlisf:
@dashattuck it looks like the robot test for this PR is failing after second attempt. Could you take a look at it please? Thank you! :)

robertbwatson:
> @ceiroa this looks good except I noticed there were some "fqn" in Trigger meta xml files. Are they added on purpose? I need to do a quick research on that.

@xinnanlisf  I think that gets added automatically if the file was created locally in VS Code as I've seen it before if I create a new class locally instead of in Salesforce. If the metadata is retrieved from Salesforce, I think it gets removed from the XML files.

xinnanlisf:
@robertbwatson so i think it is fine then?

robertbwatson:
> @robertbwatson so i think it is fine then?

@xinnanlisf I've commited up https://github.com/SalesforceFoundation/EDA/pull/960/commits/5e207963860c2d8940e0745f35693a3be321bbf3 after pushing to Salesforce and pulling back down. So the XML files should be good now.

juliesnow:
@ceiroa - pls see what you think about minor edits above...

ceiroa:
@juliesnow - Looks good. Maybe I'd say "without having to provide" instead of "without providing"? Up to you...

ceiroa:
Thanks @robertbwatson for those updates to the XML files! (Sorry, somehow I missed some of the notifications in this PR.)

robertbwatson:
@ceiroa The builds pass here, so can this just be merged in to get it off Dina's full plate? :)

dgamboni:
As per Bobby & Carlos, I'll merge this PR without any manual QA. Thanks for your help!

PR #964, opened 2019-10-14
Preferred Email/Phone Translation Issue Fix

### LINKED ISSUE #946: Preferred Phone functionality does not work in languages other than English
**Summary**

EDA 1.82 added a trigger handler for Preferred Phone which will automatically update the standard Phone based on the selection in the Preferred Phone picklist, and will ensure that a preferred phone is selected if more than one phone field is populated.

This functionality does not work as expected when the language is not English. 

**Workaround**

In EDA Settings (as of EDA 1.84), make sure Enable Enhanced Preferred Phone Functionality is unchecked. Or, remove translations on Preferred Phone picklist and work with this functionality using the English selections only.

**Steps to Repeat**

In English:

1. Create a Contact. Set Preferred Phone to Home and populate Home Phone and Other Phone. Save.
2. Confirm that the value in the Home Phone is now in the Phone standard field.
3. Change Preferred Phone to Other. Save.
4. Confirm that the value in the Other Phone is now in the Phone standard field.

Confirm that Translations are enabled for the organization in Translation Workbench. Make sure French is an active language. Switch User to use French Language (User record > Change Language to French - no other changes)

1. Go to the same Contact as used in the English example above.
2. Change Preferred Phone to Domicle (Home)

image

**Expected Behavior**

Record should save and update Phone with the value in the Téléphone (domicile) field.

**Actual Behavior**

Error on Save.

image

Translates to: The phone specified in Preferred Phone was not found. Be sure to type the label of an existing custom phone field.

It appears that we are hard-coding the English values [here](https://github.com/SalesforceFoundation/EDA/blob/master/src/classes/CON_Preferred_TDTM.cls#L61).

skosgi1:
# Critical Changes

# Changes
## Updates to the Preferred Phone Functionality
The enhanced Preferred Phone functionality released in update 1.84 can now be enabled in orgs that translate Catalan, Spanish, or French.
- If you are an existing EDA customer who wants to enable the feature in an org that translates any of these languages, go to EDA Settings | Accounts and Contacts and check the "Enable Enhanced Preferred Phone Functionality" checkbox.
- If you're installing in a new org that translates any of these languages, the enhanced Preferred Phone logic is enabled by default, so you don't need to change the new setting. However, we do recommend that you uncheck "Disable Preferred Phone enforcement" if you want to ensure that Contacts have a Preferred Phone value when more than one Phone field is specified on the Contact record.
- If you're translating in any other languages and want to use this feature, use the translation workbench to update Preferred Phone and Preferred Email picklist values to match the translated field labels.

**Note:** Existing customers might notice that some of EDA's unit tests fail after this upgrade. It's not required that EDA unit tests pass in customer orgs, but if you want them to pass, updating the Contact's Preferred Phone picklist values in your org to "Home Phone," "Mobile Phone," "Other Phone," and "Work Phone" should fix failures related to the upgrade.

# Issues Closed
Fixes #946 

# New Metadata

# Deleted Metadata

# Testing Notes

Testing Preferred Phone:

1. Make sure Enhanced Preferred Phone Functionality is checked : EDA Settings → Accounts & Contacts → Enable Enhanced Preferred Phone Functionality:
2. Make sure "CON_PreferredPhone_TDTM" trigger handler is active in "Trigger handlers" Tab.
3. Create a Contact with Preferred Phone as ‘Home’ and enter a value in the field ‘Home Phone’ and try saving the record. The record saves.
4. Change the language from user → User record →  Change Language to Spanish OR  
    click your avatar on top right → Settings → Language and Time zone → Español
5. Now create a contact with ‘Teléfono elegido’ as ‘Teléfono particular’  and enter a value for the field ‘Teléfono particular’  and try to save the record, the record should be saved.

Testing Preferred Email:

1.  Make sure "CON_PreferredEmail_TDTM" trigger handler is active in "Trigger handlers" Tab.
2.Create a contact with Preferred Email as "Work" and enter a value in the field "Work Email: and try saving the record. The record saves.
3.  Change the language from user → User record →  Change Language to Spanish OR  
    click your avatar on top right → Settings → Language and Time zone → Español
4. Create a Contact with "Correo electrónico elegido" as "Correo electrónco laboral" and enter a value in "Correo electrónico laboral" and try to save the record, the record should be saved.

Test Push Upgrade for Old Customers vs New Install (New Customers)
Test Two+ Phones ("Tell us which is preferred.")
Test Two Emails ("Tell us which is preferred.")
Test Run Cleanup

Note : For old customers, the picklist values must match the field label values (except for in English.) For new customers, the picklist values must match the field label values for all 4 languages (English, French, Spanish, Catalan.)


src/classes/CON_Phone.cls
```diff
@@ -314,11 +314,28 @@ public with sharing class CON_Phone {
...
+            String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
+            List<Schema.PicklistEntry> preferredPhonePicklist = UTIL_Describe.getFieldPicklistEntries('Contact', prefix+'PreferredPhone__c');
+            List<String> preferredPhoneValues = new List<String>();
+            for (Schema.PicklistEntry eachValue: preferredPhonePicklist) {
+                preferredPhoneValues.add(eachValue.getValue());
+            }
+             
+            if (!preferredPhoneValues.contains(lbl)) {
+                this.prefLabel = (preferredPhoneMappings.containsKey(lbl)) ? preferredPhoneMappings.get(lbl) : lbl;
+                System.debug('11111111 lbl'+this.prefLabel);
```
ceiroa: Remove debug statements?

src/classes/CON_Email.cls
```diff
@@ -281,11 +281,24 @@ public class CON_Email {
                 'Work Email' => 'Work',
                 'Alternate Email' => 'Alternate'
             };
-
+                     
+            String namespace = UTIL_Namespace.getNamespace();
+            String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
ceiroa: We also support custom fields (user would add a value to the picklist and the field to the layout). Would that functionality be affected?

src/classes/CON_Phone.cls
```diff
@@ -314,11 +314,28 @@ public with sharing class CON_Phone {
...
             };
-
+            
             this.value = val;
             this.label = lbl;
             this.apiName = api;
-            this.prefLabel = (preferredPhoneMappings.containsKey(lbl)) ? preferredPhoneMappings.get(lbl) : lbl;
+            
+            String namespace = UTIL_Namespace.getNamespace();
+            String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
ceiroa: Same question as above regarding prefix.

src/objectTranslations/Contact-ca.objectTranslation
```diff
@@ -3473,36 +3473,64 @@
         <name>PreferredPhone__c</name>
         <picklistValues>
             <masterLabel>Home</masterLabel>
-            <translation>Llar</translation>
+            <translation>Home Phone</translation>
```
ceiroa: We should use the Catalán translation for all these, not English (available in the fields with the same name).

skosgi1: I think since this is a Platform Only language some fields are not being translated, so I updated the picklist translations corresponding to those fields to English.

src/classes/CON_Phone.cls
skosgi1: Oops missed that. Removed it now!

src/classes/CON_Email.cls
skosgi1: That functionality would work if the label and picklist have an exact match.

src/classes/CON_Phone.cls
skosgi1: That functionality would work if the label and picklist have an exact match.

src/classes/CON_Email.cls
ceiroa: So the prefix won't be added for fields that are not in the package? 

Ideally we would test this still works, but I guess the only way to test it is after merging (with a beta). Unfortunately namespaced scratch orgs don't help here...

skosgi1: Hmm we are appending the prefix for only "PreferredPhone__c" and "Preferred_Email__c" fields which are packaged but we are not appending it to the actual phone/email fields. Please let me know if I that was your question.

ceiroa: Ah, sorry. My bad. I looked at this too quickly. Yes, we are applying the prefix only to the field, not to the picklist values (which I guess don't need it anyway). Thanks for bearing with me, Samsheeth!

src/classes/CON_Email.cls
```diff
@@ -281,11 +281,24 @@ public class CON_Email {
...
+            
+            for (Schema.PicklistEntry eachValue: preferredEmailPicklist) {
+                preferredEmailValues.add(eachValue.getValue());
+            }
+             
             this.value = val;
             this.label = lbl;
             this.apiName = api;
-            this.prefLabel = (preferredEmailMappings.containsKey(lbl)) ? preferredEmailMappings.get(lbl) : lbl;
+            if (!preferredEmailValues.contains(lbl)) {
```
robertbwatson: @skosgi1 Some comments explaining this might be ideal. Something like: "Use the old preferred email mappings if the picklist values do not contain the field label"?

Or, maybe update the comment on line 278 "Store default email field mappings to preferred email picklist values prior to EDA 1.86."

src/classes/CON_Phone.cls
```diff
@@ -314,11 +314,25 @@ public with sharing class CON_Phone {
...
+            List<Schema.PicklistEntry> preferredPhonePicklist = UTIL_Describe.getFieldPicklistEntries('Contact', prefix+'PreferredPhone__c');
+            List<String> preferredPhoneValues = new List<String>();
+            for (Schema.PicklistEntry eachValue: preferredPhonePicklist) {
+                preferredPhoneValues.add(eachValue.getValue());
+            }
+             
+            if (!preferredPhoneValues.contains(lbl)) {
```
robertbwatson: Same comment as above re: adding an explanation here or updating the comment on line 310.

src/objectTranslations/Contact-ca.objectTranslation
robertbwatson: They were previously translated though, right? So that might be a bit confusing. Is it because the Home, Mobile, and Other are not translated in the UI, and so if we translated the picklist, the functionality would still not work for them?

src/objectTranslations/Contact-ca.objectTranslation
```diff
@@ -3473,36 +3473,64 @@
...
+        </picklistValues> 
     </fields>
     <fields>
         <help>Quin correu electrònic s&apos;hauria d&apos;utilitzar per a la majoria de comunicacions que incloguin aquest contacte?</help>
         <label>Correu preferit</label>
         <name>Preferred_Email__c</name>
         <picklistValues>
             <masterLabel>Alternate</masterLabel>
-            <translation>Alternatiu</translation>
+            <translation>orreu electrònic alternatiu</translation>
```
robertbwatson: Missing a C at the beginning, I think...

src/objectTranslations/Contact-es.objectTranslation
```diff
@@ -3473,19 +3473,35 @@
...
         </picklistValues>
         <picklistValues>
             <masterLabel>Mobile</masterLabel>
-            <translation>Móvil</translation>
+            <translation>Teléfono Móvil</translation>
         </picklistValues>
         <picklistValues>
             <masterLabel>Other</masterLabel>
-            <translation>Otros</translation>
+            <translation>Otro teléfono</translation>
```
robertbwatson: I'm worried in all of these translations that maybe we're guessing and we should actually leave this up to the translations team or at least run it by them first. 

For example, why is Other / Other Phone "Otro telefono", whereas all of the others have "Telefono ..." first?

(I don't know any other languages so it's possible that this is correct, but I'm questioning it...)

Thoughts, @skosgi1 @ceiroa?

src/objectTranslations/Contact-fr.objectTranslation
```diff
@@ -3494,15 +3510,27 @@
...
         </picklistValues>
         <picklistValues>
             <masterLabel>University</masterLabel>
-            <translation>Université</translation>
+            <translation>Adresse e-mail de l&apos;université</translation>
         </picklistValues>
         <picklistValues>
             <masterLabel>Work</masterLabel>
-            <translation>Professionnel</translation>
+            <translation>E-mail alternatif</translation>
```
robertbwatson: This seems wrong also.

src/objectTranslations/Contact-fr.objectTranslation
```diff
@@ -3494,15 +3510,27 @@
         <name>Preferred_Email__c</name>
         <picklistValues>
             <masterLabel>Alternate</masterLabel>
-            <translation>Autre</translation>
+            <translation>E-mail alternatif</translation>
```
robertbwatson: Why was the translation changed completely here?

src/objectTranslations/Contact-ca.objectTranslation
skosgi1: Correct!

src/classes/CON_Email.cls
skosgi1: Good point. Updated it.

src/classes/CON_Phone.cls
skosgi1: Updated it.

src/objectTranslations/Contact-ca.objectTranslation
skosgi1: Good catch! Updated it.

src/objectTranslations/Contact-fr.objectTranslation
skosgi1: Good catch! Updated it.

src/objectTranslations/Contact-fr.objectTranslation
skosgi1: I think "Autre" translate to "Other" and the change translates to "Alternate Email". It matches the field label also.

src/objectTranslations/Contact-es.objectTranslation
skosgi1: Yeah looks like one is "other" and the other one translates to "another" ? but they have to match the field labels anyway for the functionality to work, so not sure about the inconsistency here...

ceiroa: Well, we don't have to guess if we are taking the value from the label. 

Also, yeah, it is correct that in this case "teléfono" goes after and the others have it first.

robot/EDA/resources/locators.py
```diff
@@ -144,7 +144,7 @@
     "eda_tile": "//div[@class='slds-app-launcher__tile-body']//a[contains(text(),'EDA')]/../..",
     "new_account": "//span[@title='New Account']",
     "new_account_next_button": "//button[contains(@class, 'slds-button')]//span[@class=' label bBody' and text()='Next']",
-    "new_account_name": "//input[@class='input uiInput uiInputText uiInput--default uiInput--input']",
+    "new_account_name": "//label/span[text()='Account Name']/following-sibling::span/following::input[1]",
```
robertbwatson: @dashattuck Same comment that I made earlier in another open PR - if this is being handled in #968, then let's remove it from here. I'd prefer to keep Robot changes separate from other work items.

src/objectTranslations/Contact-ca.objectTranslation
robertbwatson: Interesting. Thoughts on this one, @ceiroa? We're essentially taking away translations... but... this wouldn't be working for old or new customers anyway, since the actual fields themselves are not translated by Salesforce.

So, Catalan customers should translate both the Preferred Phone picklist and the standard field values if they need this to work in Catalan.

src/objectTranslations/Contact-es.objectTranslation
robertbwatson: @skosgi1 haha! I'm sorry. I forgot that we were mapping these to the actual field labels, which *are* translated already by Salesforce. Nevermind!

src/objectTranslations/Contact-fr.objectTranslation
robertbwatson: Nevermind - I forgot that you were just mapping them to the actual field labels

lifewithryan:
> @robertbwatson @skosgi1 Sorry, for some reason I cannot reply inline - I still don't get how exactly Platform-only languages work. The [documentation](https://help.salesforce.com/articleView?id=faq_getstart_what_languages_does.htm&type=5) says "_[...]when you select a platform-only language, all standard Salesforce labels default to English or, in select cases, to an end-user or fully supported language._" I guess that means we should leave it in English? I don't get how they are able to provide a translation in Catalán, but it does sound like having it in English is the right thing to do...

@ceiroa this is something we came across in testing when they translated EDA. Some of the bits just simply don't translate. However I remember it being very unclear as to why/what/when.

ceiroa:
Alrighty. English it is then.

skosgi1:
> Alrighty. English it is then.

In this case only work phone is translated and the rest of the phone fields/ picklist values are in English. So 1 picklist is translated into Catalan and 3 are not.

ceiroa:
Scratching my head on this one again, @skosgi1. Why is "work phone" translated? How is it different from the others? Should we just not translate any of them? It seems strange that that one is translated and not the others...

skosgi1:
> Scratching my head on this one again, @skosgi1. Why is "work phone" translated? How is it different from the others? Should we just not translate any of them? It seems strange that that one is translated and not the others...

@ceiroa In this case only "Work Phone" is custom fields so we have it in the translation file, other phone fields are standard fields and this language being platform only - as per what I read not all standard fields get translated.

ceiroa:
@skosgi1 Ah, got you. I didn't realize the other phone fields were standard. Ok, then we should be good to keep work phone in English. Thanks!

robertbwatson:
> > Scratching my head on this one again, @skosgi1. Why is "work phone" translated? How is it different from the others? Should we just not translate any of them? It seems strange that that one is translated and not the others...
> 
> @ceiroa In this case only "Work Phone" is custom field so we have it in the translation file, other phone fields are standard fields and this language being platform only - as per what I read not all standard fields get translated.

Right - and for email functionality, Alternate Email, University Email, and Work Email are all custom fields that we've added to EDA, so we've translated them ourselves. Which means that picklist is correctly translated to Catalan. But until Salesforce supports Catalan as a core language, Home, Other, and Mobile phone fields remain in English.

skosgi1:
> @skosgi1 Ah, got you. I didn't realize the other phone fields were standard. Ok, then we should be good to keep work phone in English. Thanks!

Hmm all the other picklists would be in English, but "Work Phone" would be in Catalan. Since only the field "Work Phone" is translated into Catalan. "Home Phone" , "Mobile Phone", "Other Phone" picklist values will be in English and "Work Phone" value will be in Catalan. If we keep "Work Phone" in English the functionality would not work for Catalan since the field Work Phone is now in Catalan.

robertbwatson:
@ceiroa FYI I've assigned this to you - since you requested changes it needs to be approved by you before it's moved to Dina.

robot/EDA/resources/locators.py
robertbwatson: Reverted, resolving.

skosgi1:
@robertbwatson @ceiroa I have made the changes to the code to use API values instead of labels, can you give it a quick review when you get a moment? Thank you!

src/classes/CON_Phone.cls
```diff
@@ -324,14 +324,17 @@ public with sharing class CON_Phone {
             String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
             List<Schema.PicklistEntry> preferredPhonePicklist = UTIL_Describe.getFieldPicklistEntries('Contact', prefix+'PreferredPhone__c');
             List<String> preferredPhoneValues = new List<String>();
+            Map<String, String> preferredLabelToApi = new Map<String, String>();  
             for (Schema.PicklistEntry eachValue: preferredPhonePicklist) {
-                preferredPhoneValues.add(eachValue.getValue());
+                PreferredLabelToApi.put(eachValue.getLabel(),eachValue.getValue());
             }
```
ceiroa: @skosgi1 ~ Just to confirm, will `eachValue.getValue()` get the picklist's element API value?

src/classes/CON_Phone.cls
```diff
@@ -324,14 +324,17 @@ public with sharing class CON_Phone {
...
-             
-            if (!preferredPhoneValues.contains(lbl)) {
+            
+            /* If the field label is not in Preferrred Phone picklist API values then check if it is in PreferredPhoneMappings, if yes then take the value from mappings
+             else keep the label as label. If the field label is in the Preferrred Phone picklist API values then assign the picklist API value that matches the field label. */
+            if (!PreferredLabelToApi.keySet().contains(lbl)) {
                 this.prefLabel = (preferredPhoneMappings.containsKey(lbl)) ? preferredPhoneMappings.get(lbl) : lbl;
             } else {            
-                 this.prefLabel = lbl;
+                 this.prefLabel = PreferredLabelToApi.get(lbl);
```
ceiroa: Also for my understanding; calling `PreferredLabelToApi.get(lbl)` will return the API value, not the label, right?

src/classes/CON_Phone.cls
lifewithryan: That's the way I read this as well.

skosgi1: correct! getValue() is the API and getLabel() is the field label.

src/classes/CON_Phone.cls
skosgi1: Correct!

ceiroa: Do we care that the value is getting assigned to `prefLabel` even though it's the API name, or are we ok with that?

skosgi1: Hmm not sure if I understood your question correctly @ceiroa, We're assigning the value of the picklist API corresponding to picklist label that matches with the field label. Please let me know if it doesn't answer your question.

robertbwatson: If I understand Carlos correctly, he's just concerned that the variable name "prefLabel" is a little misleading since it might actually be the API name.

I agree it is a little confusing. The comments are good here and in CON_Email.cls but I wonder if it would be nice to explain the "why" in the comments. The comments explain the code but don't necessarily explain why it is done this way.

skosgi1:
@robertbwatson I updated the code as we discussed earlier. Can you take a quick pass it again? thx.

lifewithryan:
> @robertbwatson I updated the code as we discussed earlier. Can you take a quick pass it again? thx.

@robertbwatson @skosgi1 -- I've pulled this down too as I want to run through some things but was in a meeting so won't get a chance to finish until tomorrow. First skim looks good. Looks like there may be potential in the future to make the new method you created slightly more generic/flexible so that both email and phone could use it, but that's not anything I would hold up a merge for. more like a tech debt future task.

skosgi1:
> > @robertbwatson I updated the code as we discussed earlier. Can you take a quick pass it again? thx.
> 
> @robertbwatson @skosgi1 -- I've pulled this down too as I want to run through some things but was in a meeting so won't get a chance to finish until tomorrow. First skim looks good. Looks like there may be potential in the future to make the new method you created slightly more generic/flexible so that both email and phone could use it, but that's not anything I would hold up a merge for. more like a tech debt future task.

@lifewithryan Thanks for the quick feedback. I added the new generic method to UTIL_Describe.cls now.

src/classes/CON_Phone.cls
```diff
@@ -183,6 +183,16 @@ public with sharing class CON_Phone {
                 contact.OtherPhone = contact.Phone;          
         }
     }
+
+    /*******************************************************************************************************
+    * @description Returns a map with Preferred Phone picklist entry's label as key and API as value
+    * @retun Map<String, String> A key value pair of Preferred Phone picklist entrys's label and API respectively
```
robertbwatson: Nit: return

src/classes/CON_Email.cls
```diff
@@ -162,6 +162,16 @@ public class CON_Email {
         }
     }
 
+    /*******************************************************************************************************
+    * @description Returns a map with Preferred Email picklist entry's label as key and API as value
+    * @retun Map<String, String> A key value pair of Preferred Email picklist entry's label and API respectively
```
robertbwatson: Nit: return

src/classes/CON_Phone.cls
```diff
@@ -183,6 +183,16 @@ public with sharing class CON_Phone {
...
+    */
+    private static Map<String, String> createPreferredLabelToAPIMap() {
```
robertbwatson: I'm sorry for holding up the QA on this and adding a last minute suggestion (& also sorry for going AFK immediately after our meeting today!)

When I had mentioned caching performance enhancements here, I meant that it would be nice to only generate the map once. This will be especially useful if we have a large number of contacts that we're iterating over at once.

I won't hold up the review for this, but a future enhancement here could be something like:
```
private static Map<String, String> labelToAPIMap;
private static Map<String, String> getPreferredLabelToAPIMap() {
    if (labelToAPIMap == null) {
        // code goes here
        labelToAPIMap = UTIL_Describe...
    }
    return labelToAPIMap;
}
```

In this example, the map is only generated once and subsequent calls to `getPreferredLabelToAPIMap` return the stored value.

skosgi1: Thanks for the clarification @robertbwatson. I made the change, can you take another quick look? Thx.

robertbwatson: Perfect @skosgi1 -- thanks!

PR #967, opened 2019-10-14
[EDA]: Ensure that Account Naming for HouseHold is in Alphabetical Order

### LINKED ISSUE #966: Ensure that Account Naming for HouseHold is in Alphabetical Order

Bug 

When the Account name format of the org contains either {!LastName} ({!FirstName}), 
{!{!FirstName} {!LastName}, {!LastName} ({!{!FirstName}}), the Account renaming isn't correctly renaming the Account name in the correct alphabetical order. 

The Account Name process should update the Account Name in the correct alphabetical order regardless of which format is set in the org. 

This helps ensure that user experience is consistent regardless of the Account name format selected.

kdy618:
# Critical Changes

# Changes
We have fixed the issue with the inconsistent Account naming in alphabetical order. 

# Issues Closed
Closes #966

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/UTIL_ACCT_Naming.cls
+++ b/src/classes/UTIL_ACCT_Naming.cls
@@ -408,7 +408,7 @@ public class UTIL_ACCT_Naming {
         Pattern MyPattern = Pattern.compile(regexPattern);
         Matcher MyMatcher = MyPattern.matcher(acctNameFormat);
 
-        //Checks {!{FirstName}}
+        //Checks {!{!FirstName}}
         String regexPatternDoubleToken = '\\{![^\\}]*\\{!.*\\}[^\\{!]*\\}';
         Pattern MyPatternDoubleTokenCheck = Pattern.compile(regexPatternDoubleToken);
         Matcher MyMatcherDoubleTokenCheck = MyPatternDoubleTokenCheck.matcher(acctNameFormat);
@@ -550,7 +550,6 @@ public class UTIL_ACCT_Naming {
         }
 
         //Handles processing fullNameSpec
-        System.debug('processNames checkFullNameSpec -->' + ns.fullNameSpec);
         String acctName = stringConNames(cons[i], ns.fullNameSpec);
         acctName = acctName.replace('FirstNameSpec', firstName);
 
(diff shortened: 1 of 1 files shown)


src/classes/UTIL_ACCT_Naming.cls
```diff
@@ -135,7 +135,9 @@ public class UTIL_ACCT_Naming {
         }
 
         if (defaultRecTypeId == hhAccountRecordTypeId) {
-            dynamicSoql += 'Phone, Fax FROM Contacts WHERE Exclude_from_Household_Name__c != true)';
+            dynamicSoql += 'Phone, Fax FROM Contacts WHERE Exclude_from_Household_Name__c != true ORDER BY LastName, FirstName)';
```
robertbwatson: @kdy618 So, unfortunately, the Contact Name is an encryptable field and, thus, can't be used in the ORDER BY clause :-( (See: https://help.salesforce.com/articleView?id=security_pe_considerations_general.htm&type=5)

The fields can be queried though, so we might need to do the ordering ourselves...

If you run the "setup_encryption" flow in your dev scratch org, you should then be able to go to the Platform Encryption | Encryption Policy settings and encrypt the Contact Name field. I'm assuming that after you do so, you will get errors here since the code is ordering by LastName & FirstName. (The errors will either be instant or may not appear until trying to re-deploy the code or run the unit tests.)

Does this all sound right, @lifewithryan?

robertbwatson: I haven't read through the main refactoring PR yet, but it would be a good idea to encrypt all of the fields on the Contact to try to make sure there aren't other areas that we may need to fix in the refactoring as well.

src/classes/UTIL_ACCT_Naming.cls
```diff
@@ -400,26 +401,30 @@ public class UTIL_ACCT_Naming {
...
 
-        //Set setFNSpec if it finds {!{FirstName}}
         if (MyMatcherDoubleTokenCheck.find()) {
+            //If it finds {!{!FirstName}}
             setFNSpec = true;
-        }
-
-        if (MyMatcher.find()) {
-            return acctNameFormat.subString(MyMatcher.start()-1, MyMatcher.end());
+            firstNameFormat = acctNameFormat.subString(MyMatcherDoubleTokenCheck.start(), MyMatcherDoubleTokenCheck.end());
```
robertbwatson: Are there unit tests that we can update here to verify the alphabetical order is now fixed?

src/classes/UTIL_ACCT_Naming.cls
robertbwatson: ^ Definitely test this first! I might be wrong and I would hate to make you go through the work of fixing this for no reason.

kdy618: @robertbwatson , good point and catch. I'll work on testing out the encryption to check if it breaks.

PR #970, opened 2019-10-16
Ensure that Account Naming for HouseHold is in Alphabetical Order

### LINKED ISSUE #966: Ensure that Account Naming for HouseHold is in Alphabetical Order

kdy618:
# Critical Changes

# Changes
We have fixed the issue with the inconsistent Account naming in alphabetical order.

# Issues Closed
Closes #966 

# New Metadata

# Deleted Metadata

# Testing Notes


src/classes/UITL_SortContacts.cls
```diff
@@ -0,0 +1,106 @@
+/*
```
robertbwatson: General comment - this class & the test class need to be added to the package.xml. It looks like there is a build failure on the branch that I think is related to needing to do this.

src/classes/UITL_SortContacts.cls
```diff
@@ -0,0 +1,106 @@
...
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Utilities
+* @description Utility for Sorting Contacts
+*/
+public class UITL_SortContacts implements Comparable{
```
robertbwatson: Nit: space after Comparable

src/classes/UITL_SortContacts.cls
```diff
@@ -0,0 +1,106 @@
...
+    public String FirstName;
+    public String LastName;
+    public Contact contactRecord;
+
+    /*******************************************************************************************************
+    * @description Constructor that sets the Contact's First and Last Name and the Contact record itself.
+    * @param c is the Contact record
+    * @return
```
robertbwatson: Nit: You can get rid of the return since there isn't a return

src/classes/UITL_SortContacts.cls
```diff
@@ -0,0 +1,106 @@
...
+    */
+    public UITL_SortContacts(Contact c) {
+        FirstName = c.FirstName;
+        LastName = c.LastName;
+        contactRecord = c;
+    }
+
+    /*******************************************************************************************************
+    * @description method that
```
robertbwatson: Might want to finish this sentence :)

src/classes/UITL_SortContacts.cls
```diff
@@ -0,0 +1,106 @@
...
+    public Integer compareTo(Object compareTo) {
+        return sortByLastNameFirstName(compareTo);
+    }
+
+    /*******************************************************************************************************
+    * @description checks whether it needs to sort by Last Name or First Name
+    * @param compareTo is the object that is being passed for comparison.
+    * @return Integer
+    */
+    public Integer sortByLastNameFirstName(Object compareTo) {
```
robertbwatson: Can this method (sortByLastNameFirstName) as well as sortByFirstName and sortByLastName below be private methods? That way the only method that is public is the compareTo method and the class is responsible for calling the internal, private methods.

src/classes/UITL_SortContacts.cls
```diff
@@ -0,0 +1,106 @@
...
+    * @description method that
+    * @param compareTo is the object that is being passed for comparison.
+    * @return Integer
+    */
+    public Integer compareTo(Object compareTo) {
+        return sortByLastNameFirstName(compareTo);
+    }
+
+    /*******************************************************************************************************
+    * @description checks whether it needs to sort by Last Name or First Name
```
robertbwatson: Doesn't it also do the sorting? (since it calls out to the other methods...)

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -1117,4 +1116,150 @@ private class ACCT_IndividualAccounts_TEST {
         List<Contact> returnedContactDelete = [SELECT Id, IsDeleted FROM Contact WHERE Id = :con.Id];
         System.assertEquals(0, returnedContactDelete.size());
     }
+
+    /*********************************************************************************************************
+    * @description Test Method to test sorting Contacts alphabetically by LastName FirstName
+    * with {!{!FirstName}} {!LastName} Household format.
+    */
+    @isTest
+    private static void sortContactsFirstLastNames() {
```
robertbwatson: These test methods are great. What happens if one of the First Names is null/blank? (I think only Last Name is required...)

src/classes/UTIL_ACCT_Naming.cls
```diff
@@ -158,11 +162,16 @@ public class UTIL_ACCT_Naming {
         accountNamingFormat = checkAccountFormat();
 
         NameSpec ns;
+
+        List<Contact> returnSortedContacts = new List<Contact>();
+		returnSortedContacts = sortContacts(cons);
```
robertbwatson: It looks like these two lines can just be truncated down to: `List<Contact> returnSortedContacts = sortContacts(cons);`

src/classes/UTIL_ACCT_Naming.cls
```diff
@@ -195,6 +204,31 @@ public class UTIL_ACCT_Naming {
         return accountNamingFormat;
     }
 
+    /*******************************************************************************************************
+    * @description This calls the UTIL_SortContacts to sort Contacts in alphabetical order.
+    * @param cons a list of Contacts that needs to be alphabetized.
+    * @return List<Contact>.
+    */
+
+    public static List<Contact> sortContacts(Contact[] cons) {
```
robertbwatson: Can this just be a private method?

robertbwatson: Also, nitpick: I prefer `List<Contact>` notation over `Contact[]`

src/classes/UTIL_ACCT_Naming.cls
```diff
@@ -195,6 +204,31 @@ public class UTIL_ACCT_Naming {
...
+        List<UITL_SortContacts> sortedContacts = new List<UITL_SortContacts>();
+        List<Contact> finalSortedContacts = new List<Contact>();
+
+        UITL_SortContacts sortCons;
+
+        for (Contact c : cons) {
+            sortCons = new UITL_SortContacts(c);
+            sortedContacts.add(sortCons);
```
robertbwatson: You could just get rid of line 217 above. This could be consolidated to `sortedContacts.add(new UTIL_SortContacts(c));`

src/classes/UTIL_SortContacts_TEST
```diff
@@ -0,0 +1,161 @@
...
+        contacts[5].FirstName = 'Charles';
+        contacts[5].LastName = 'Brett';
+        contacts[6].FirstName = 'Cat';
+        contacts[6].LastName = 'Pierre';
+        contacts[7].FirstName = 'Lyla';
+        contacts[7].LastName = 'Pierre';
+        contacts[8].FirstName = 'Annie';
+        contacts[8].LastName = 'Pierre';
+
+       	Test.startTest();
```
robertbwatson: Nit: misaligned here

src/classes/UTIL_SortContacts_TEST
```diff
@@ -0,0 +1,161 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Utilities
+* @description Test for UTIL_SortContacts
+*/
+@isTest
+private class UTIL_SortContacts_TEST {
```
robertbwatson: Same comment as in the other test class - these are great, but I'd maybe like to see an example where the FirstName was null/blank.

src/classes/UITL_SortContacts.cls
robertbwatson: Ohh... this is misspelled. Can we easily rename this to `UTIL_SortContacts`? (not UITL)

src/classes/UITL_SortContacts.cls
kdy618: I can't believe I completely missed that typo the whole entire time. I've updated the prefix and the places it is referenced. I've also updated the package.xml and added this new class & its test class.

src/classes/UTIL_SortContacts_TEST
kdy618: Added a new test method sortContactsWithoutFirstNames() to test contact insertion with null firstnames.

src/classes/ACCT_IndividualAccounts_TEST.cls
kdy618: Added new test method sortContactsWithoutFirstNames() to test blank firstnames.

PR #991, opened 2019-11-06
Add Einstein Analytics Template

jstvz:
This PR adds an Einstein Analytics template, a new task to deploy it, and a standalone installer plan. Additionally a custom task to check for the presence of `EinsteinAnalytics*` permission sets has been added as an installer preflight check.

----

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
(When installer links are available)
Installer:
- `install`: Verify that the step is skipped when `DeveloperWave` is not included in the scratch org definition.
- `ra_einstein_template`: Verify that the preflight validation fails when `DeveloperWave` is not included in the scratch org definition.

Post-install:
- From the app waffle, select **Einstein Analytics Studio -> Create -> Start from Template -> Education Cloud**


tasks/check_einstein.py
```diff
@@ -0,0 +1,23 @@
...
+    task_options = {
+        "permission_sets": {
+            "description": "List of permission set names to check for, (ex: EinsteinAnalyticsUser)",
+            "required": True,
+        }
+    }
+
+    def _run_task(self):
+        query = self._get_query()
+        print(query)
```
davisagli: Use self.logger or remove this; print goes into the ether on MetaCI

cumulusci.yml
```diff
@@ -37,6 +48,17 @@ tasks:
...
+    deploy_einstein_templates:
+        group: 'EDA: Einstein Templates'
+        description: Deploys Einstein Analytics Templates
+        class_path: cumulusci.tasks.salesforce.Deploy
+        ui_options:
+            name: Deploy Einstein Analytics Templates
+        options:
+            path: unpackaged/config/analytics
+            namespace_inject: $project_config.project__package__namespace
+            unmanaged: False
```
davisagli: I saw at least one use of the hed__ prefix in the wave metadata. Should that be tokenized, or are we only trying to support use with managed EDA?

cumulusci.yml
```diff
@@ -262,4 +286,26 @@ plans:
...
-                    unmanaged: false
+                    unmanaged: False
+            4:
+                task: deploy_einstein_templates
+                ui_options:
+                    is_required: False
+                checks:
+                    - when: "tasks.check_einstein_licenses()"
+                      action: skip
+                      message: Skipped because Einstein Analytics permission sets are not available.
```
davisagli: It looks like you have the logic backward here? The check_einstein_licenses task returns true if 1 or more permset was found, and this check says to hide the step in that case.

tasks/check_einstein.py
jstvz: Missed that one, thanks.
```suggestion
```

cumulusci.yml
jstvz: Managed-only at this point.

cumulusci.yml
```diff
@@ -262,4 +286,26 @@ plans:
...
+        title: Install Education Cloud Recruiting Analytics Template
+        is_listed: False
+        tier: additional
+        steps:
+            1:
+                task: deploy_einstein_templates
+        checks:
+            - when: "not tasks.has_einstein_licenses()"
+              action: error
+              message: Einstein Analytics permission sets are not available.
```
davisagli: @jstvz if you want you can put the `checks` in the `deploy_einstein_templates` task config so you don't have to duplicate it.

PR #994, opened 2019-11-12
W-037259: [EDA] Create custom settings to prevent deletion

kdy618:
# Critical Changes

# Changes
We have updated the Hierarchy Custom Settings to include 9 new checkbox fields. These checkboxes, designed for internal-use only, provide the scaffolding for a forthcoming deletion-prevention feature; they currently do not provide any functionality. We strongly advise that EDA Admins do not manage Hierarchy Custom Settings directly and instead use the EDA Settings UI.

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
Please see description column on this spreadsheet. https://salesforce.quip.com/oU2qAs876VyW#CQFACAEqoyM


robertbwatson:
@kdy618 @juliesnow  Just wanted to add a comment here that I don't think this should go under Critical Changes. We might also want to consider language that makes it clear that these new settings do absolutely nothing at this time and are placeholders for future work.

xinnanlisf:
@dashattuck it looks like the robot test is still failing after several attempts. Could you take a look at it please? Thank you!

kdy618:
@robertbwatson  @xinnanlisf  ty so much! Moved the description to below critical change.

kdy618:
@juliesnow , could you kindly take a look if these new checkboxes should be under changes?

juliesnow:
> @kdy618 @juliesnow are these descriptions still a work in progress? I ask because I don't think the plan is to prevent deletion unless there are certain EDA child records present. If we just leave it as preventing deletion, people may think that we will block deletion outright. Maybe I'm not on the same page, but I didn't that was the case.

@ryanguest - Not sure if we've changed them since you made this comment, so please take another look. The descriptions indicate which child records are involved, and I've matched the descriptions for existing classes to the descriptions for those classes in the Apex docs, so I believe them to be accurate.

src/objects/Hierarchy_Settings__c.object
```diff
@@ -355,6 +355,96 @@
         <type>Text</type>
         <unique>false</unique>
     </fields>
+    <fields>
+        <fullName>Prevent_Account_Deletion__c</fullName>
+        <defaultValue>false</defaultValue>
+        <description>Prevent an Account from being deleted if it has any related records (for example, Affiliation, Program Enrollment, or Course Connection records).</description>
```
robertbwatson: @kdy618 @lifewithryan @xinnanlisf  Is this ready for QA or do we need to make changes here?

(Did we decide earlier on "Reserved for future use."? I'm not sure what we ultimately decided!)

dgamboni:
@robertbwatson Do you know if this closes any issues? It looks like the settings don't actually do anything so I'm guessing it doesn't close any of the open issues related to deletion at this time.

xinnanlisf:
@dgamboni no it doesn't have a github issue I believe since it is an internal item. @kdy618 correct me if I am wrong.

lifewithryan:
> @robertbwatson Do you know if this closes any issues? It looks like the settings don't actually do anything so I'm guessing it doesn't close any of the open issues related to deletion at this time.

@dgamboni -- correct this doesn't really close any issues. It's opening the door for us to being handling the data integrity items.

kdy618:
@xinnanlisf, yes you are correct. This is an internal item. @dgamboni

src/objects/Hierarchy_Settings__c.object
kdy618: @robertbwatson , yep we will use "Reserved for future use." This is now back to dev review. :)

dgamboni:
> @dgamboni no it doesn't have a github issue I believe since it is an internal item. @kdy618 correct me if I am wrong.

@xinnanlisf @lifewithryan @jacebryan I don't think this is a truly internal task, though, because there potential for customers to see these changes if they go to Setup > Custom Settings (as opposed to, say, a cci task for our qa_org flows, which it seems customers would never be exposed to.) So, I'm thinking maybe it would technically need an issue? (Though I'm not advocating for more 'paperwork', as it were, and am 100% fine if we don't have a related issue -- I'm just trying to grok the rationale.)

dgamboni:
Tested / Passes QE. @spinkelman Will be merging this PR.

spinkelman:
Merging per @ dgamboni request.

PR #1012, opened 2019-11-25
Add upgraded org regression flow

jstvz:
Roughly, the existing `regression_org` flow includes these steps:

      1. installs beta dependencies
      2. installs managed release
      3. upgrades to latest beta
      4. deploys config_regression in the current directory

While this confirms that the last managed release can be upgraded, it doesn't reflect what customers will see in their org after a push upgrade (as it includes updates to unmanaged metadata that existing customers wouldn't receive). This PR adds a new flow, `upgraded_org`, that includes unmanaged metadata from the last managed release, then upgrades to the latest beta release to cover this use case.

In addition, the automation docs have been rewritten in the new style and moved to a new top-level directory, `documentation/`.

----

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
```
cci flow run upgraded_org --org release
```


robertbwatson:
@jstvz I see you have Kim & Dina as reviewers -- did you want @kdy618 to review this before Dina QA's it? (Just wondering if it should be assigned to somebody and what next steps are)

robertbwatson:
> @jstvz , this looks great! Super excited about it as well as the notes on the document for testing. Will this be mimic for K12's regression testing for existing customers as well?
> Lastly, I wanted to ask what latest_release does? I tried to look for where it is but had no luck. I'm wondering what's the difference between latest_release : config_regression vs. config_regression.

@kdy618 I'll let James chime in if I'm wrong, but I think after reading through this, `latest_release` grabs the latest version from the repo (https://github.com/SalesforceFoundation/EDA/pull/1012/files#diff-b0e88920e25a1ba7fb4868790bbe6be8R18) and configures the org based on the latest release. The `upgraded_org` flow then installs the latest beta on top of that. This simulates an existing customer being upgraded to the latest release.

The `config_regression` flow performs similar setup, but it includes all of our unmanaged metadata, even changes we've made in the upcoming release. Since existing customers will *not* receive unmanaged metadata changes when they are upgraded, this does *not* correctly simulate an existing customer upgrade which is why we're introducing a new `upgraded_org` flow.

I'm not sure about your K-12 question. We may need to make additional changes in that repo to cover all upgrade scenarios.

jstvz:
Spot on, Bobby.

dgamboni:
@jstvz Does this mean we will have 2 "top-level" flows to run regression?
**regression_org** for new customers 
**upgraded_org** for existing customers

PR #1021, opened 2019-12-04
[EDA] Prevent Deletion of Contact if There are Associated Child Records

### LINKED ISSUE #904: Block Deletion of Contact, Course Offering, or the Course Connection when Term Grades & Attendance Event records exist
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Course Connections, Contacts, and Course Offerings can all be deleted even if they are associated with Attendance Events and Term Grades. This leads to data integrity issues.

**What is the expected behavior?**
Course Connections, Contacts, and Course Offerings should not be able to be deleted if there are associated Attendance Events and/or Term Grades.

**What is the motivation / use case for changing the behavior?**
Data integrity

kdy618:
# Critical Changes

# Changes
- We've added functionality to prevent the deletion of any Contact record that has related Address, Affiliation, Application, Attendance Event, Attribute, Behavior Involvement, Contact Languages, Course Connection, Course Offering (Faculty), Program Enrollment, Relationship, Test records. This change is for internal-use only and is disabled at this time. You won't see any difference in functionality.

# Issues Closed
Closes #904
# New Metadata

# Deleted Metadata

# Testing Notes


src/classes/CON_CannotDelete_TDTM.cls
```diff
@@ -13,60 +13,79 @@
...
     POSSIBILITY OF SUCH DAMAGE.
 */
 /**
 * @author Salesforce.org
 * @date 2016
 * @group Accounts
 * @group-content ../../ApexDocContent/Contacts.htm
-* @description Stops a Contact from being deleted if it has any Affiliation, 
-* Program Enrollment or Course Enrollment children.
+* @description Stops a Contact from being deleted if it has any Address, Affiliation,
```
lifewithryan: @kdy618 just checking if this is up to date. I know you were working on the list of objects we were checking for. If it's up to date all good.

src/classes/CON_CannotDelete_TDTM.cls
```diff
@@ -13,60 +13,79 @@
...
     /*******************************************************************************************************
-    * @description Stops a Contact from being deleted if it has any Affiliation, 
-    * Program Enrollment or Course Enrollment children.
-    * @param listNew the list of Contacts from trigger new. 
-    * @param listOld the list of Contacts from trigger old. 
-    * @param triggerAction which trigger event (BeforeInsert, AfterInsert, etc.). 
-    * @param objResult the describe for Accounts 
-    * @return dmlWrapper.  
+    * @description Stops a Contact from being deleted if it has any Affiliation,
+    * Program Enrollment, Term Grade, Attendance Event,or Course Enrollment children.
```
lifewithryan: @kdy618 - not sure if this list needs to be updated?

src/classes/CON_CannotDelete_TDTM.cls
```diff
@@ -13,60 +13,79 @@
...
+                        (SELECT ID FROM Contact.Courses_Taught__r LIMIT 1),
+                        (SELECT ID FROM Contact.Program_Enrollments__r LIMIT 1),
+                        (SELECT ID FROM Contact.Relationships__r LIMIT 1),
+                        (SELECT ID FROM Contact.Student_Course_Enrollments__r LIMIT 1),
+                        (SELECT ID FROM Contact.Term_Grades__r LIMIT 1),
+                        (SELECT ID FROM Contact.Tests__r LIMIT 1)
+                        FROM Contact
+                        WHERE ID in :oldlist])
+        {
+        	if (c.Addresses__r.size() > 0 || c.Affl_Accounts__r.size() > 0 ||
```
lifewithryan: I know it seems like it might be an odd ask, but can we move these checks to a method that does the checks and returns either true or false? That just seems like a ton of parameters. Something like:

if(hasChildObjects(con)) { ... }

src/classes/CON_CannotDelete_TEST.cls
```diff
@@ -35,39 +35,1368 @@
...
+
+        List<Contact> returnNewContacts = consAccsAddrs.contacts;
+
+
+        List<Contact> queryContacts = [SELECT Id
+                                       FROM Contact
+                                       WHERE Id IN :returnNewContacts];
 
+        Test.startTest();
+        Database.DeleteResult[] results = Database.delete(queryContacts, false);
```
lifewithryan: this comment goes for all the tests. Do we know if one failure is stopping the whole batch of deletes? If so, probably not a big deal, but if not, we may want to have some tests that test situations where when we delete some contacts, a few don't have any blockers and others do.

src/classes/CON_CannotDelete_TDTM.cls
kdy618: Thank you for the reminder. I removed "relationship" from it.

src/classes/CON_CannotDelete_TEST.cls
kdy618: @lifewithryan  good point. Added a new test that test if there are 2 records, 1 Contact has child related records, and 1 without and if I delete both that the Contact with child related records will not be deleted and the one without child related records will be deleted.

src/classes/CON_CannotDelete_TDTM.cls
lifewithryan: I see that we've talked about this and you have removed Relationship from it. However, it looks like we are still querying for relationship and also checking it in the Boolean hasChildRecord method. Is this still relevant? Wasn't the "relationships" the one that was blocking things via the platform that we couldn't do anything about?

kdy618: @lifewithryan  good catch 👍 I completely forgot that part.

src/classes/CON_CannotDelete_TDTM.cls
```diff
@@ -13,60 +13,89 @@
...
+   * @description Evaluates whether the Contact has any child related records.
+   * @param c is the current Contact record.
+   * @return Boolean.
+   ********************************************************************************************************/
+   private static Boolean hasChildRecords(Contact c) {
+       return (c.Addresses__r.size() > 0 || c.Affl_Accounts__r.size() > 0 ||
+            	c.Applications_Applicant__r.size() > 0  || c.Applications_Preparer__r.size() > 0 ||
+                c.Attendance_Events__r.size() > 0 || c.Attributes__r.size () > 0 ||
+                c.Behavior_Involvements__r.size() > 0 || c.Contact_Languages__r.size() > 0 ||
+				c.Courses_Taught__r.size() > 0 || c.Program_Enrollments__r.size() > 0 ||
```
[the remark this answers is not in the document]
kdy618: @lifewithryan , nope, but for some reason on VS code it doesn't show the extra indentation. :(

spinkelman:
Passes QA - Merging
Will Close WI

PR #1025, opened 2019-12-06
Block Deletion of Terms when Term Grades Exist

xinnanlisf:
# Critical Changes

# Changes
Updates to Ensure Data Integrity : We have added logic to one of the Custom Hierarchy Settings, to prevent deletion of Term records when there are related Term Grade records. This feature is designed for internal-use only at this time, and is being released disabled; you will not see any difference in functionality. We strongly advise that EDA Admins do not manage Hierarchy Custom Settings directly and instead use the EDA Settings UI.

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


src/classes/TERM_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,116 @@
...
+        Test.startTest();
+            delete term;
+        Test.stopTest();
+        
+        List<Term__c> termsList = [SELECT Id FROM Term__c];      
+        System.assertEquals(0, termsList.size());       
+    }
+
+    /*********************************************************************************************************
+    * @description Test deletion of Terms with Term Grades and preventing deletion setting is turned off
```
kdy618: @xinnanlisf , I might be confusing things but if the Prevent_Term_Deletion__c = True, shouldn't it throw an error and not allow the deletion of the term record?

src/classes/TERM_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,116 @@
...
+        // Insert Term
+        Term__c term = UTIL_UnitTestData_TEST.getTerm(academicAccount[0].Id, 'Spring 20');
+        insert term;
+    }
+    
+    /*********************************************************************************************************
+    * @description Test deletion of Terms with Term Grades
+    */
+    @isTest 
+    static void testDeleteTermsWithTermGrades() {
```
kdy618: @xinnanlisf , I'm sorry but I think I'm missing something here. This test method pass which is great but that's confusing me since we're not setting `UTIL_CustomSettingsFacade.getSettingsForTests(new` Hierarchy_Settings__c(Prevent_Term_Deletion__c = TRUE) `?

src/labels/CustomLabels.labels
```diff
@@ -1525,7 +1525,7 @@
         <language>en_US</language>
         <protected>true</protected>
         <shortDescription>Select a desired picklist value for the batch job to selct a value from</shortDescription>
-        <value>Select a field for Preferred Phone to copy the value from when the standard Phone field has no value, or more than one &apos;Phone&apos; field is defined.</value>
+        <value>Select a field for Preferred Phone to copy the value from when the standard Phone field has no value, or more than one 'Phone' field is defined.</value>
```
kdy618: @xinnanlisf , is this change part of the original branch?

xinnanlisf: @kdy618 should be good now!

src/classes/TERM_CannotDelete_TDTM.cls
```diff
@@ -0,0 +1,75 @@
...
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER 
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT 
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN 
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE 
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Terms
```
robertbwatson: Not that it probably matters much until we fix the ApexDoc generator (if we ever do), but the existing TERM_ classes are under the group "Term"

src/classes/TERM_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,131 @@
...
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2019
+* @group Terms
```
robertbwatson: Not that it probably matters much until we fix the ApexDoc generator (if we ever do), but the existing TERM_ classes are under the group "Term"

dgamboni:
Tested / Passes QE. Will merge this ticket. @juliesnow please review release notes after PTO at your convenience. Thanks!

PR #1031, opened 2019-12-17
W-037247 - When new Contact is created, map Contact currency to Account Currency

xinnanlisf:
# Critical Changes

# Changes
## Update to Default Currency Settings for Accounts
In orgs that have enabled multiple currencies, the currency setting for a newly created Contact that is not related to an existing Household or Administrative Account will now be copied to the currency setting on the Account that's created for the new Contact.

For important information about enabling multiple currencies in your org, see [Manage Multiple Currencies](https://help.salesforce.com/articleView?id=admin_currency.htm&type=5) and [Considerations for Enabling Multiple Currencies](https://help.salesforce.com/articleView?id=admin_enable_multicurrency_implications.htm&type=5).

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
- Go to Setup -> Company Information, click edit and check 'Activate Multiple Currencies'
- Go to Setup -> Manage currency, add EUR as a new currency type
- Go to EDA Setting page, change the default account model to 'Household'
- Add 'Contact currency' to HEDA contact page layout and 'Account currency' to HEDA Household page layout
- Create a new contact with currency = 'EUR'
- Make sure the currency of related account, aka test contact 'Household', is 'EUR' as well


src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -1528,4 +1528,80 @@ private class ACCT_IndividualAccounts_TEST {
...
+    /*********************************************************************************************************
+    * @description In a multi-currency enabled org, create a contact with different currency than the default one.
+    * Make sure that the account automatically created and the contact have the same currency.
+    */
+    @isTest
+    private static void insertContactMultiCurrencyEnabled() {
+        if (UserInfo.isMultiCurrencyOrganization()) {
+            //Test data setup
+            String nonDefaultCurrencyCode = '';
+            String qryCurrencyTypeString = 'SELECT IsoCode FROM CurrencyType WHERE IsCorporate = false LIMIT 1';
```
lifewithryan: More out of curiosity here, is there a reason we are using the database.query and not a [SELECT Isocode From.... type of query here? Nothing I would require changing, just curious.

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -1528,4 +1528,80 @@ private class ACCT_IndividualAccounts_TEST {
...
+
+            //Run the logic
+            Test.startTest();
+            insert testContact;
+            Test.stopTest();
+
+            //Assert the result
+            String qryString = 'SELECT Id, Name, CurrencyIsoCode FROM Account';
+            List<Account> resultAccounts = Database.query(qryString);
+            system.assertEQUALS(1, resultAccounts.size(), 'More than one Account were created.');
```
lifewithryan: nit: System.AssertEquals() but no big deal :)

I like that you are using the assert messages here, I think we should all get in that habit. Thank you :)

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -1528,4 +1528,80 @@ private class ACCT_IndividualAccounts_TEST {
         System.assertEquals('Ms. Betty, Lily and Tom White Family', requeryFinalAccount.Name);
 
     }
+
+    /*********************************************************************************************************
+    * @description In a multi-currency enabled org, create a contact with different currency than the default one.
+    * Make sure that the account automatically created and the contact have the same currency.
+    */
+    @isTest
+    private static void insertContactMultiCurrencyEnabled() {
```
lifewithryan: I may have missed it but do we want unit tests for Admin Accounts as well?

src/classes/ACCT_IndividualAccounts_TEST.cls
xinnanlisf: Great question! The reason here is because those currency fields are not available unless an org enables multi-currency. The build will fail if we use [SELECT...FROM... ].

src/classes/ACCT_IndividualAccounts_TEST.cls
xinnanlisf: Good catch. Maybe we can remove that setting setup line so that it is using default model?

src/classes/ACCT_IndividualAccounts_TEST.cls
lifewithryan: aaah yes thank you :)

src/classes/ACCT_IndividualAccounts_TEST.cls
lifewithryan: do you mean line 1549?

dgamboni:
Tested / Passes EDA QE. As per Ryan, OK to merge.

PR #1032, opened 2019-12-17
Update trialforce template ID 

jofsky:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
`cci org browser trial`

### DIFF
--- a/orgs/trial.json
+++ b/orgs/trial.json
@@ -1,6 +1,6 @@
 {
   "orgName": "EDA - Trial Org",
-  "template": "0TT1Q000004XHAF",
+  "template": "0TT1Q000004XHAK",
   "settings": {
     "lightningExperienceSettings": {
       "enableS1DesktopEnabled": true


orgs/trial.json
```diff
@@ -1,6 +1,6 @@
 {
   "orgName": "EDA - Trial Org",
-  "template": "0TT1Q000004XHAF",
+  "template": "0TT1Q000004XHAK",
```
lifewithryan: @jstvz @jofsky Can I get some clarity around what's happening behind the scenes here? Does EDA have a new Template?

jstvz: The EDA Trialforce template is updated with every production release. The basic workflow for EDA is:

1. A new production version is cut and the TSO is upgraded
2. CCE makes updates to the TSO
3. A new trialforce template ID is created on the day of the production push and WLMA/repository are updated.

lifewithryan: okay cool. I just don't recall ever seeing this in a PR before. Thanks!

PR #1033, opened 2019-12-17
Feature/robot/affiliation mappings

dashattuck:
This test verifies that all standard fields have the proper default values.  It verifies that all text fields can retain null value on save, that all text fields can return text value on save, that all checkboxes can retain checked and unchecked values on save.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


robot/EDA/resources/AffiliationsPageObject.py
```diff
@@ -76,13 +75,7 @@ def open_item(self, locator, error_message, capture_screen):
...
-#        self.selenium.click_element(locator)
-
-        self.selenium.driver.execute_script(
-            "arguments[0].click()", 
-            self.selenium.driver.find_element_by_xpath(locator)
-        )
-
+        self.selenium.click_element(locator)
 
         if capture_screen:
```
skristem: Are you trying to capture a screenshot when there is an error? If yes u can use the property @capture_screenshot_on_error on the keyword and it will take care of the screenshot. If you are capturing a screenshot for specific scenario then it is better to provide those details in the keyword documentation

robot/EDA/resources/AffiliationsPageObject.py
```diff
@@ -140,20 +133,6 @@ def Go_to_affiliations_edit_mode(self, loc):
         self.selenium.click_element(affiliations_locators["button_save_affiliation"])
         self.eda.close_toast_message()
 
-    def _expandShadowRoot(self, element):
```
skristem: please add documentation

robot/EDA/resources/AffiliationsPageObject.py
```diff
@@ -140,20 +133,6 @@ def Go_to_affiliations_edit_mode(self, loc):
...
-        shadow_root = self.selenium.driver.execute_script('return arguments[0].shadowRoot', element)
-        return shadow_root
-
-
-    def expand_shadow_root(self):
```
skristem: please add documentation

robot/EDA/resources/AffiliationsPageObject.py
dashattuck: It's an arg value - so I'm only capturing a screenshot as passed to the function, which in production is always passed in as 'False'.  During debugging, all I have to do is toggle the arg to be 'True' and then I can a nifty screenshot :-)

skristem: Then maybe provide some info in comments to say pass capture_screen as False to turn off screen capture and  True to turn on screen capture. Also, if u mention 
"@capture_screenshot_on_error    
    def open_item(self, locator, error_message):" it will capture screenshot only if test case fails on this keyword (which is useful for debugging)and you wouldn't have to worry about it as additional step to ur keyword. So please see if that meets ur purpose

robot/EDA/resources/AffiliationsPageObject.py
skristem: please ignore, I was looking at wrong version, since this is deleted no documentation needed

robot/EDA/resources/AffiliationsPageObject.py
dashattuck: This was deleted prior to code review - no documentation needed since function was removed previously.

robot/EDA/resources/AffiliationsPageObject.py
dashattuck: Sounds good - I'll explain the purpose of the parameter.

PR #1034, opened 2019-12-18
[EDA] New Object: Application

xinnanlisf:
# Critical Changes

# Changes
## New Application Object
We've added a new object to help you track information about a student's application to an educational institution or program. For sample use cases, configuration information, and the steps required to create Application records, see (insert link).
**Note:** After you create Applications, create Test records to track the tests an applicant has completed and Test Score records to track the results of a test.

# Issues Closed

# New Metadata
* Application__c Object
* Application Tab
* Application Trigger

# Deleted Metadata

# Testing Notes


src/layouts/Application__c-EDA Application Layout.layout
```diff
@@ -0,0 +1,113 @@
...
+        <detailHeading>false</detailHeading>
+        <editHeading>true</editHeading>
+        <label>Information</label>
+        <layoutColumns>
+            <layoutItems>
+                <behavior>Readonly</behavior>
+                <field>Name</field>
+            </layoutItems>
+            <layoutItems>
+                <behavior>Required</behavior>
```
robertbwatson: It's not clear from reading the spec if this field should be required at the database level or just on the page layout. We should confirm this.

src/layouts/Application__c-EDA Application Layout.layout
```diff
@@ -0,0 +1,113 @@
...
+                <field>Applicant__c</field>
+            </layoutItems>
+            <layoutItems>
+                <behavior>Required</behavior>
```
robertbwatson: It's not clear from reading the spec if this field should be required at the database level or just on the page layout. We should confirm this.

src/objects/Application__c.object
```diff
@@ -0,0 +1,268 @@
...
+        <type>Default</type>
+    </actionOverrides>
+    <actionOverrides>
+        <actionName>View</actionName>
+        <type>Default</type>
+    </actionOverrides>
+    <allowInChatterGroups>false</allowInChatterGroups>
+    <compactLayoutAssignment>SYSTEM</compactLayoutAssignment>
+    <deploymentStatus>Deployed</deploymentStatus>
+    <enableActivities>true</enableActivities>
```
robertbwatson: This wasn't checked on the object in the spec, so we should confirm whether or not this should be enabled.

src/objects/Application__c.object
```diff
@@ -0,0 +1,268 @@
...
+        <inlineHelpText>The status of the application review process.</inlineHelpText>
+        <label>Application Status</label>
+        <required>false</required>
+        <trackHistory>false</trackHistory>
+        <trackTrending>false</trackTrending>
+        <type>Picklist</type>
+        <valueSet>
+            <restricted>true</restricted>
+            <valueSetDefinition>
+                <sorted>false</sorted>
```
robertbwatson: Comments in the spec suggest this should be sorted alphabetically.

src/objects/Application__c.object
```diff
@@ -0,0 +1,268 @@
...
+    <fields>
+        <fullName>Preparer__c</fullName>
+        <deleteConstraint>SetNull</deleteConstraint>
+        <description>The Contact preparing the application, often a guardian.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The Contact preparing the application. For example, a guardian or a person assisting someone with special needs.</inlineHelpText>
+        <label>Preparer</label>
+        <referenceTo>Contact</referenceTo>
+        <relationshipLabel>Applications (Preparer)</relationshipLabel>
+        <relationshipName>Applications1</relationshipName>
```
robertbwatson: May want to consider this API relationship name being something like Applications_Preparer? (Thoughts @lifewithryan ?)

src/objects/Application__c.object
```diff
@@ -0,0 +1,268 @@
...
+    </fields>
+    <fields>
+        <fullName>Applicant__c</fullName>
+        <deleteConstraint>SetNull</deleteConstraint>
+        <description>The Contact applying to the institution or program.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The Contact applying to the institution or program.</inlineHelpText>
+        <label>Applicant</label>
+        <referenceTo>Contact</referenceTo>
+        <relationshipLabel>Applications</relationshipLabel>
```
robertbwatson: The acceptance criteria in the specs says `Applications (Applicant)` - so we may want to reach out and verify that's still true.

src/objects/Application__c.object
```diff
@@ -0,0 +1,268 @@
...
+        <type>Lookup</type>
+    </fields>
+    <label>Application</label>
+    <listViews>
+        <fullName>All</fullName>
+        <filterScope>Everything</filterScope>
+        <label>All</label>
+    </listViews>
+    <nameField>
+        <displayFormat>App-{0000}</displayFormat>
```
robertbwatson: App or APP? (Spec shows `APP`)

src/objects/Application__c.object
```diff
@@ -0,0 +1,268 @@
...
+                    <default>false</default>
+                    <label>Transfer</label>
+                </value>
+                <value>
+                    <fullName>Graduate</fullName>
+                    <default>false</default>
+                    <label>Graduate</label>
+                </value>
+                <value>
+                    <fullName>Post-Doc</fullName>
```
robertbwatson: I have questions around whether or not this is the correct term (https://en.wikipedia.org/wiki/Postdoctoral_researcher) and if we need to add any additional picklist values here - I'll add my comment to the spec on Monday and followup next week.

src/layouts/Application__c-EDA Application Layout.layout
xinnanlisf: I thought the reason of requiring fields on layout was because users can decide if they want the fields to be required or not but I am ok with double checking in spec.

src/objects/Application__c.object
xinnanlisf: Looks like it is not enabled. I will update this.

src/objects/Application__c.object
xinnanlisf: Oh I missed that comment, thanks for catching it.

src/objects/Application__c.object
xinnanlisf: I have updated this. :)

src/layouts/Application__c-EDA Application Layout.layout
lifewithryan: Correct, I've been avoiding requiring at the database level and only requiring via page layout.

src/objects/Application__c.object
lifewithryan: Yes, let's move with Bobby's suggestion. I thought at one point the spec called that out, but I'm betting in the numerous changes, copy/pastes to different versions it got dropped somewhere along the way.

src/objects/Application__c.object
lifewithryan: When using prefixes like that I alway go Upper Case, so APP

src/objects/Application__c.object
lifewithryan: I wanna say I saw that actually come up in the comments. I personally *hate* calling it Post-Doc and think we should be explicit here, but we should tag Samara in that doc, ultimately it would be her call I think.

src/objects/Application__c.object
xinnanlisf: I think this is only used in code when referring a relationship in query, but I guess it does not hurt to give it a better api name.

xinnanlisf: Although, I assume we need to change the applicant relationship name to Applications_Applicant too?

src/objects/Application__c.object
lifewithryan: @xinnanlisf @robertbwatson I think there may be some confusion here as the comments in the spec were a bit mixed up.
The Application Status, should be in that chronological order like it is in the spec doc: Submitted, Incomplete, Received, In Review, Waitlist, Admit, Deny, Deferred Offer, Accepted Offer, Declined Offer 

The Application Type picklist should be in alphabetical order.

src/objects/Application__c.object
xinnanlisf: Also...should we better call it Application_Preparers? @lifewithryan @robertbwatson

lifewithryan: I think Applications Preparer is actually more correct. When viewing a contact record, it will be the related list of Applications that this contact was the Preparer for. So multiple applications but one preparer (which is the contact we are viewing). Same for the Applicant version.

src/objects/Application__c.object
```diff
@@ -0,0 +1,263 @@
...
+                    <default>false</default>
+                    <label>Graduate</label>
+                </value>
+                <value>
+                    <fullName>Transfer</fullName>
+                    <default>false</default>
+                    <label>Transfer</label>
+                </value>
+                <value>
+                    <fullName>Undergrad</fullName>
```
robertbwatson: According to the spec (& help text), this should be Undergraduate.

unpackaged/config/dev/profiles/Admin.profile
```diff
@@ -36,6 +36,9 @@
         <layout>Account-%%%NAMESPACE%%%HEDA Organization Layout</layout>
         <recordType>Account.%%%NAMESPACED_ORG%%%University_Department</recordType>
     </layoutAssignments>
+    <layoutAssignments>
+        <layout>%%%NAMESPACE%%%Application__c-%%%NAMESPACE%%%EDA Application Layout</layout>
```
robertbwatson: @lifewithryan @xinnanlisf It looks like we only maintain layout assignments in the admin.profile here for objects where we have more than one layout (for example, all objects where there are record types). Since we only have one layout for the Application object, I'm going to go ahead and remove these three lines.

src/objects/Application__c.object
```diff
@@ -0,0 +1,263 @@
...
+        <description>The type of program or track the applicant is applying to. For example, Undergraduate, Graduate, or Transfer.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>The type of program or track the applicant is applying to. For example, Undergraduate, Graduate, or Transfer.</inlineHelpText>
+        <label>Application Type</label>
+        <required>false</required>
+        <trackHistory>false</trackHistory>
+        <trackTrending>false</trackTrending>
+        <type>Picklist</type>
+        <valueSet>
+            <restricted>false</restricted>
```
robertbwatson: @lifewithryan @xinnanlisf Everywhere else that we have picklist fields marked as unrestricted, the `restricted` tag simply doesn't exist in the XML. I'm going to remove it from these picklist fields.

dgamboni:
This has passed QA. @juliesnow please review the Release Notes. Thanks!

dgamboni:
Tested / Passes QE. Merging PR.

PR #1037, opened 2019-12-27
[EDA] New Objects: Test & Test Score

robertbwatson:
# Critical Changes

# Changes
## New Test and Test Score Objects
- We've added two new objects, Test and Test Score, to help you track information about the tests a student has completed and the results of those tests. These objects are intended to track high stakes tests, like the SAT, GMAT, or other tests that might influence a student's acceptance into an educational institution or program. For sample use cases, configuration information, and the steps required to create Test and Test Score records, see (insert link).

- For all EDA installations, the Test Score related list is automatically added to the Test page layout. 

# Issues Closed

# New Metadata
* Test__c Object
* Test_Score__c Object
* Test Tab
* Test Score Tab
* Test Trigger
* Test Score Trigger

# Deleted Metadata

# Testing Notes


src/objects/Test_Score__c.object
```diff
@@ -0,0 +1,168 @@
...
+    </actionOverrides>
+    <actionOverrides>
+        <actionName>View</actionName>
+        <type>Default</type>
+    </actionOverrides>
+    <allowInChatterGroups>false</allowInChatterGroups>
+    <compactLayoutAssignment>SYSTEM</compactLayoutAssignment>
+    <deploymentStatus>Deployed</deploymentStatus>
+    <description>The score on a section of a specific test or a composite score.</description>
+    <enableActivities>true</enableActivities>
```
lifewithryan: something Kevin Swiggum brought up during discussion was whether or not we wanted to enable things like activities by default if there's not an obvious use case, and instead letting the customer enable this since it is something they can do. I think I agree with that. I know we've been wishy washy and inconsistent about these types of settings though but without having actual use cases for how activities would be used on something like a test score object, maybe we should start "not" enabling this type of thing.

src/objects/Test_Score__c.object
```diff
@@ -0,0 +1,168 @@
...
+                    <fullName>Language</fullName>
+                    <default>false</default>
+                    <label>Language</label>
+                </value>
+                <value>
+                    <fullName>Mathematics</fullName>
+                    <default>false</default>
+                    <label>Mathematics</label>
+                </value>
+                <value>
```
lifewithryan: I thought we were going with Overall at the end? (I honestly don't have a dog in that fight, but I know it was brought up)

src/objects/Test_Score__c.object
robertbwatson: I'm just going by what the spec says. I did confirm that this setting, as well as various other object settings, can be updated by the customer once the object is packaged.

src/objects/Test_Score__c.object
robertbwatson: Yep -- we are. It showed up at the end when I looked at it in Salesforce but for some reason is still alphabetizied here. I'll fix it.

robertbwatson:
> I wonder if we should consider tweaking the names of the class files a bit to make it obvious that "Test" refers to the object and not a Unit test related file. I do notice however that we are properly leveraging the UPPER case "TEST" for our actual unit testing stuff, so maybe it's a non-issue.

@lifewithryan I think this is back to you for review - I moved the "Overall" picklist value to the bottom. I'm not sure about renaming the new trigger files because it's following the convention we've used everywhere else. I think because we use "_TEST" to designate test classes that we should be okay with TDTM_Test and TDTM_Test_Score but let me know if you can think of a better name.

robertbwatson:
@juliesnow "For new EDA installations, the Test Score related list is automatically added to the Test page layout. If your organization has an existing EDA installation, you'll want to manually add the Test Score related list to your Test page layout to best take advantage of these new objects."

This is incorrect... all EDA installations will have the Test Score related list on the Test page layout, since these are new objects we're introducing at the same time in this release.

dgamboni:
Tested / Passes QE. There's an open question as to whether "Percentile" should be labeled with a % sign or not, so I will leave this open until that is resolved.

dgamboni:
> @juliesnow "For new EDA installations, the Test Score related list is automatically added to the Test page layout. If your organization has an existing EDA installation, you'll want to manually add the Test Score related list to your Test page layout to best take advantage of these new objects."
> 
> This is incorrect... all EDA installations will have the Test Score related list on the Test page layout, since these are new objects we're introducing at the same time in this release.

@juliesnow @robertbwatson I updated the RN's to reflect this.

PR #1056, opened 2020-01-14
[EDA] - HH & Admin Account Naming Format is Incorrect If Prefix is Not Inside Braces

### LINKED ISSUE #1053: [EDA] - HH & Admin Account Naming Format is Incorrect If Prefix is Not Inside Braces
* **Do you want to request a *feature* or report a *bug*?** : BUG

* **What is the current behavior?** : 
When the running User's language is in Spanish or in French, the "Administrative Account Name Format" and "Household Account Name Format" is duplicating some words.

* **What is the expected behavior?** : 
The Name Formats follow the convention (and do not duplicate words.)

* **What is the motivation / use case for changing the behavior?** : 
User experience

Steps to Reproduce : 
1. Change running user language to French (Setup > Users > Edit > Change Language > Save.)
2. Go to Parameters EDA > Systeme > Modifier > [This should change the drop-downs at the bottom to **Compte administratif {!LastName}**] > Enregistrer
3. Create a Contact "Fox Mulder"
4. Look at the Account Name

  Expected Result : Compte administratif Mulder 
  Actual Result : Compte administratif Compte administratif Mulder

5. Go to Parameters EDA > Systeme > Modifier > Change to HH Account Model & Change Naming Format to **Foyer {!{!FirstName}} {!LastName}**

  Expected Result : Foyer Dana Scully
  Actual Result : Foyer Foyer Dana Scully

Note that this also happens in Spanish. It does not happen in English.

kdy618:
# Critical Changes

# Changes
We've fixed an issue with Account naming formats. Now, if an Account format contains a prefix that's not surrounded by braces (for example, **Familia {!LastName}**), it will no longer duplicate the prefix when the Account is created or renamed. 

# Issues Closed
Closes #1053

# New Metadata

# Deleted Metadata

# Testing Notes
See https://salesforce.quip.com/9YVNA2Ey1dNV#TBEACAPXhMY for testing details


src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -41,8 +41,8 @@ private class ACCT_IndividualAccounts_TEST {
     */
     @isTest
     public static void insertContactTriggerOff() {
-        UTIL_CustomSettingsFacade.getSettingsForTests(
-            new Hierarchy_Settings__c(Account_Processor__c = UTIL_Describe_API.getAdminAccRecTypeID()));
+        UTIL_CustomSettingsFacade.getSettingsForTests(new Hierarchy_Settings__c
+                                                        (Account_Processor__c = UTIL_Describe_API.getAdminAccRecTypeID()));
```
xinnanlisf: I am interested in the reason behind moving 'Account_Processor__c' to a different line.

kdy618: @xinnanlisf , I think it was more of a consistency thing across all test methods. I can move it back.

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -634,16 +633,13 @@ private class ACCT_IndividualAccounts_TEST {
 
         assertAccount = [SELECT Id, RecordType.Name, Name
                          FROM Account];
-
```
xinnanlisf: @kdy618, looks like there is some inconsistency on empty lines between query and assertion code.

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -1860,6 +1890,59 @@ private class ACCT_IndividualAccounts_TEST {
 
     }
 
+    /*********************************************************************************************************
+    * @description Test Method to test whether Account Name is renamed appropriately if current users's
+    * language is not set to English. This also test if Prefixes are appropriately stripped out of the
+    * Account format so that it doesn't duplicate the prefix in the Account rename.
+    */
+    @isTest
+    private static void testPrefixAndNonEnglishLanguage(){
```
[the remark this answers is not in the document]
lifewithryan: I assume that my above comment is irrelevant. I believe after we chatted, it really had nothing to do with the language and more to do that any word outside the braces at the beginning was causing an issue. (I had made the above comment last week but never submitted the review as we were chatting about it.)

src/classes/ACCT_IndividualAccounts_TEST.cls
```diff
@@ -1853,10 +1881,66 @@ private class ACCT_IndividualAccounts_TEST {
...
+        UTIL_CustomSettingsFacade.getSettingsForTests(new Hierarchy_Settings__c(Account_Processor__c = UTIL_Describe_API.getHhAccRecTypeID(),
+                                                     	Household_Account_Naming_Format__c = 'Foyer {!LastName} ({!{!FirstName}})',
+                                                     	Automatic_Household_Naming__c = true));
+
+        String andConnector = Label.defaultNamingConnector;
+
+        Id sysAdminProfileId = UTIL_Profile.getInstance().getProfileIds(UTIL_Profile.SYSTEM_ADMINISTRATOR)[0];
```
lifewithryan: If this really isn't testing another language, but rather testing the prefix, lets perhaps change the name of the test class. However, if you wanna try, we could write a unit test much like this one but perhaps set the user language to Spanish or something and see if it works. Might be overkill if the issue *really* wasn't the language though.

src/classes/ACCT_IndividualAccounts_TEST.cls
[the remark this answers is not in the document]
kdy618: @lifewithryan , yes the bug is really with prefix and not other language. I used "Foyer" as a prefix instead of using the standard salutation "Mr./Ms." etc.

src/classes/ACCT_IndividualAccounts_TEST.cls
kdy618: @lifewithryan , yep I can certainly update the method name since the bug is really prefix and nothing to do with translation. Also, I was able to test the language translation thing by using custom labels for the "and" because it was failing the test class before when I hard coded "And" instead of using the custom label that contains "and" value.

lifewithryan: Nice. Thank you!

dgamboni:
Tested / Passes QE. @juliesnow please review the RN's when you get a chance and ping Kim or me with any questions. Thanks!

PR #1066, opened 2020-01-23
Set minimum_cumulusci_version to 3.5.0

jstvz:
Required for the `update_admin_profile` task to run.

----

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -1,3 +1,4 @@
+minimum_cumulusci_version: 3.5.0
 project:
     name: EDA
     package:


cumulusci.yml
```diff
@@ -1,3 +1,4 @@
+minimum_cumulusci_version: 3.5.0
```
kdy618: @jstvz , we're setting minimum_cumulusci_version to version 3.5.0 because 3.5.0 has the fix for the errors we were seeing for the update_admin_profile task?

jstvz: Yes. We made a change to the default `update_admin_profile` task to only deploy profiles, which changed the directory we store those in. After EDA's and K12's custom tasks were updated to reference the new variables they no longer worked with versions  before v3.5.0.

kdy618: @jstvz , gotcha! Thank you so much for explaining this. 👍

PR #1080, opened 2020-02-05
Fix failing EDA Unit Tests that fail only when installed and run from a K12 org.

### LINKED ISSUE #1088: Failing EDA Unit Tests that fail only when installed and run from a K12 org.

Bug

there are 4 failing EDA Unit Tests that fail only when installed and run from a K12 org.
The unit tests are all in the AFFL_AccRecordType_TEST.cls (addRecordTypeToAcc, checkNPE, fillInPrimaryFieldForContacts and updateAdminAccRecordTypeBulk).

All the unit tests should pass when run from both EDA and K-12

Ensure functionality works

* **Acceptance Criteria**

  - **Given**: I have k12 installed
  - **When**: I run all unit tests
  - **Then**: All unit tests should pass

skosgi1:
# Critical Changes

# Changes
We've updated the way Affiliation Mappings works when EDA orgs are created for new customers. Now, Affiliations Mappings will use API names for Account record types instead of using labels. We've also updated several Affiliations TDTM classes to accommodate this difference. With these changes, EDA unit tests will pass when run on orgs that also have the K-12 Architecture Kit installed. 

# Issues Closed
#1088 

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/AFFL_AccChange_TDTM.cls
+++ b/src/classes/AFFL_AccChange_TDTM.cls
@@ -72,15 +72,23 @@ public with sharing class AFFL_AccChange_TDTM extends TDTM_Runnable {
                 Integer i = 0;
                 for (SObject so : newlist) {
                		Affiliation__c affl = (Affiliation__c)so;
-               		String lookupFieldLabel = AFFL_MultiRecordType_TDTM.afflMapper.accRecTypeToContactLabel.get(affl.Affiliation_Type__c);
-           			String lookupFieldName = AFFL_MultiRecordType_TDTM.afflMapper.contactLabelNames.get(lookupFieldLabel);
+               		String lookupLabelOrAPIName = AFFL_MultiRecordType_TDTM.afflMapper.accRecTypeToContactLabel.get(affl.Affiliation_Type__c);
+                  if (lookupLabelOrAPIName == NULL) {
+                      String affiliationTypeAPI = afflMapper.accRecTypeLabelToAPI.get(affl.Affiliation_Type__c);
+                      lookupLabelOrAPIName = afflMapper.accRecTypeToContactLabel.get(affiliationTypeAPI);
+                  }
+           			  String lookupFieldAPIName = AFFL_MultiRecordType_TDTM.afflMapper.contactLabelNames.get(lookupLabelOrAPIName);
 
                		Affiliation__c afflOld;
-               		String lookupFieldLabelOld, lookupFieldNameOld;
+               		String lookupFieldLabelOrAPIOld, lookupFieldAPINameOld;
                		if(oldlist[i] != null) {
                    		afflOld = (Affiliation__c)oldlist[i];
-                   		lookupFieldLabelOld = AFFL_MultiRecordType_TDTM.afflMapper.accRecTypeToContactLabel.get(afflOld.Affiliation_Type__c);
-                   		lookupFieldNameOld = AFFL_MultiRecordType_TDTM.afflMapper.contactLabelNames.get(lookupFieldLabelOld);
+                   		lookupFieldLabelOrAPIOld = AFFL_MultiRecordType_TDTM.afflMapper.accRecTypeToContactLabel.get(afflOld.Affiliation_Type__c);
+                        if (lookupFieldLabelOrAPIOld == NULL) {
+                            String affiliationTypeAPI = afflMapper.accRecTypeLabelToAPI.get(afflOld.Affiliation_Type__c);
+                            lookupFieldLabelOrAPIOld = afflMapper.accRecTypeToContactLabel.get(affiliationTypeAPI);
+                        }
+                   		lookupFieldAPINameOld = AFFL_MultiRecordType_TDTM.afflMapper.contactLabelNames.get(lookupFieldLabelOrAPIOld);
                		}
 
                		if (isAfflTypeEnforced == true) {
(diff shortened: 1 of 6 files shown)


src/classes/AFFL_ContactChange_TDTM.cls
```diff
@@ -102,23 +105,23 @@ public with sharing class AFFL_ContactChange_TDTM extends TDTM_Runnable {
 
 	                   	//CONTACT FIELD CHANGED
 	                   	if(affl.Primary__c && affl.Contact__c != null && afflOld.Contact__c != null
-	                   		&& affl.Contact__c != afflOld.Contact__c && newRelatedContact != null && oldRelatedContact != null && lookupFieldName != null) {
+	                   		&& affl.Contact__c != afflOld.Contact__c && newRelatedContact != null && oldRelatedContact != null && lookupFieldAPIName != null) {
```
lifewithryan: some questions here: same as before with the checking for null and then checking if they're aren't equal, if the old contact field is null, and we are changing to not be null, do we need to allow for that?  Secondly, if affOld.Contact is essentially oldReleatedContact, do we need both of those comparisons? I see that one is the contact itself and one is the contact Id, so I think that case could we just use newRelated and oldRelated and drop the other two?

src/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -159,16 +176,22 @@ public with sharing class AFFL_MultiRecordTypeMapper {
     * @return String The name of the key affiliation Contact field.
     ********************************************************************************************************/
 	public String getKeyAfflFieldByAccRecordType(ID recordTypeId) {
-		for(String recTypeName : accRecTypeToContactLabel.keySet()) {
-            if (Schema.Sobjecttype.Account.getRecordTypeInfosByName().get(recTypeName) != null)
-            {
+		for (String recTypeName : accRecTypeToContactLabel.keySet()) {
```
lifewithryan: the spacing is off here, in vs code looks like these are tabs and not spaces?

src/classes/STG_InstallScript.cls
```diff
@@ -142,12 +142,12 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
-            mappings.add(new Affl_Mappings__c(Name = 'Household Account', Account_Record_Type__c = 'Household Account', Primary_Affl_Field__c = 'Primary Household'));
-            mappings.add(new Affl_Mappings__c(Name = 'Educational Institution', Account_Record_Type__c = 'Educational Institution', Primary_Affl_Field__c = 'Primary Educational Institution'));
-            mappings.add(new Affl_Mappings__c(Name = 'University Department', Account_Record_Type__c = 'University Department', Primary_Affl_Field__c = 'Primary Department'));
-            mappings.add(new Affl_Mappings__c(Name = 'Sports Organization', Account_Record_Type__c = 'Sports Organization', Primary_Affl_Field__c = 'Primary Sports Organization'));
+            mappings.add(new Affl_Mappings__c(Name = 'Academic Program', Account_Record_Type__c = 'Academic_Program', Primary_Affl_Field__c = 'Primary Academic Program', Auto_Program_Enrollment__c = true, Auto_Program_Enrollment_Status__c = 'Current', Auto_Program_Enrollment_Role__c = 'Student'));
```
lifewithryan: with this fix in place, do we still need/want to do the record types as API Names as well? (I think with these modifications they should work either way correct? Can users also use the field api names in Primary_Affl_Field__c (such as Primary_Academic_Program__c)? If so, do we want to change that now?

skosgi1: I think we have W-037866 for that.

src/classes/AFFL_AccChange_TDTM.cls
```diff
@@ -95,29 +103,29 @@ public with sharing class AFFL_AccChange_TDTM extends TDTM_Runnable {
 
                    		//ACCOUNT FIELD CHANGED
                    		if(affl.Primary__c && affl.Account__c != null && afflOld.Account__c != null
-                       		&& affl.Account__c != afflOld.Account__c && lookupFieldName != null) {
+                       		&& affl.Account__c != afflOld.Account__c && lookupFieldAPIName != null) {
```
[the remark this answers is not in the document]
skosgi1: I have actually just changed the label name here, I am not entirely sure how the logic works. Please let me know if you want me to dig deeper.

src/classes/AFFL_ContactChange_TDTM.cls
skosgi1: I have actually just changed the label name here, I am not entirely sure how the logic works. Please let me know if you want me to dig deeper.

src/classes/AFFL_MultiRecordTypeMapper.cls
skosgi1: Fixed it.

src/classes/AFFL_AccChange_TDTM.cls
```diff
@@ -95,29 +103,29 @@ public with sharing class AFFL_AccChange_TDTM extends TDTM_Runnable {
...
-                               		relatedContact.put(lookupFieldName, affl.Account__c);
+                           		if (lookupFieldAPINameOld != null) {
+                               		relatedContact.put(lookupFieldAPINameOld, null);
+                               		relatedContact.put(lookupFieldAPIName, affl.Account__c);
                            		}
                        		}
                        		dmlWrapper.objectsToUpdate.add(relatedContact);
 
-                       		//afflMulti.uncheckOtherPrimariesSameType(); --> we don't need to call this because
+                       		  //afflMulti.uncheckOtherPrimariesSameType(); --> we don't need to call this because
```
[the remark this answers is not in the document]
skosgi1: Fixed it.

src/classes/AFFL_AccChange_TDTM.cls
```diff
@@ -72,15 +72,23 @@ public with sharing class AFFL_AccChange_TDTM extends TDTM_Runnable {
...
+               		String lookupLabelOrAPIName = AFFL_MultiRecordType_TDTM.afflMapper.accRecTypeToContactLabel.get(affl.Affiliation_Type__c);
+                  if (lookupLabelOrAPIName == NULL) {
+                      String affiliationTypeAPI = afflMapper.accRecTypeLabelToAPI.get(affl.Affiliation_Type__c);
+                      lookupLabelOrAPIName = afflMapper.accRecTypeToContactLabel.get(affiliationTypeAPI);
+                  }
+           			  String lookupFieldAPIName = AFFL_MultiRecordType_TDTM.afflMapper.contactLabelNames.get(lookupLabelOrAPIName);
 
                		Affiliation__c afflOld;
-               		String lookupFieldLabelOld, lookupFieldNameOld;
+               		String lookupFieldLabelOrAPIOld, lookupFieldAPINameOld;
```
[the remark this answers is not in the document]
skosgi1: Fixed it.

src/classes/STG_InstallScript.cls
lifewithryan: oh I thought that work item was strictly for the UI to reflect the label when viewing and the API Name when editing, but if that bit is included, no worries.

src/classes/AFFL_AccChange_TDTM.cls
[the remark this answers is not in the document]
lifewithryan: I think that is fine, it's not hurting anything other than some readability.

src/classes/AFFL_ContactChange_TDTM.cls
lifewithryan: well it certainly wouldn't hurt to gain some understanding at a bit lower level, but it's not something I would hold up a merge for.

src/classes/STG_InstallScript.cls
skosgi1: Yep, we decided to make that change in that WI. Resolving.

dgamboni:
Tested / Passes QE. @juliesnow please give love to our RN's! Thanks!

PR #1090, opened 2020-02-10
[EDA] Prevent Deletion of a Case when there is a Related Behavior Involvement Record

### LINKED ISSUE #1091: Prevent Deletion of Case Record If There Are Child Records
Note: FOR INTERNAL USE ONLY. For support questions, please use the EDA Chatter group in The Power of Us Hub.
For feature requests, please use the Ideas tab in The Power of Us Hub.

Do you want to request a feature or report a bug?
Feature

What is the current behavior?
When users delete a CASE with Behavior Involvement as child records, the system will allow users to delete the CASE.

What is the expected behavior?
We want to provide customers the ability to prevent deletion of CASEs if it has Behavior Involvement child records by allowing them to toggle a checkbox in the Hierarchy Settings.

What is the motivation / use case for changing the behavior?
By giving customers the ability to control whether a CASE with child records should be deleted or not, it will help them with maintaining a clean system.

Acceptance Criteria

Given: If Prevent_CASE_Deletion__c checkbox is enabled in Hierarchy Settings
When: I try to delete a CASE record with child records
Then: I should not be able to delete the CASE record.

skosgi1:
# Critical Changes

# Changes
We've added functionality to prevent the deletion of Case records that have related Behavior Involvement records. This change is for internal-use only and is disabled at this time. You won't see any difference in functionality.

# Issues Closed
Closes #1091 

# New Metadata

# Deleted Metadata

# Testing Notes

Scenario 1: “Prevent Case Deletion” checkbox is enabled and Case record has related Behavior Involvement records.
Check “Prevent Case Deletion” checkbox in Hierarchy Settings.
Create a Case record
Associate a Behavior Involvement record.
Delete the Case record.
You should receive a pop up error that you cannot delete this Course Offering. See https://salesforce.quip.com/xZuIAOMoQoJD#PPOACAn2eHJ for the error message.

Scenario 2: “Prevent Case Deletion” checkbox is enabled and Case record has no related child records.
Check “Prevent Case Deletion” checkbox in Hierarchy Settings.
Create a Case record
Do not associate any child records
Delete the Case record.
You should be able to delete the record

Scenario 3: “Prevent Case Deletion” checkbox is disabled and Case record has Behavior Involvement record child records.
UnCheck “Prevent Case Deletion” checkbox in Hierarchy Settings.
Create a Case record
Associate a Behavior Involvement record.
Delete the Case record.
You should be able to delete the record

Scenario 4: “Prevent Case Deletion” checkbox is disabled and Case record has no related child records.
UnCheck “Prevent Case Deletion” checkbox in Hierarchy Settings.
Create a Case record
Delete the Case record.
You should be able to delete the record

### DIFF
--- a/src/classes/CASE_CannotDelete_TEST.cls
+++ b/src/classes/CASE_CannotDelete_TEST.cls
@@ -82,7 +82,7 @@ private class CASE_CannotDelete_TEST {
 
         Test.startTest();
             List<Case> casesToDelete = [SELECT Id FROM Case];
-            Database.DeleteResult[] results = Database.delete(casesToDelete, false);
+            Database.DeleteResult[] results = Database.delete(casesToDelete, FALSE);
         Test.stopTest();
 
         // Ensure no Cases are deleted
@@ -95,11 +95,11 @@ private class CASE_CannotDelete_TEST {
     */
     @isTest
     static void testCaseDeleteSettingOff() {
-        UTIL_CustomSettingsFacade.getSettingsForTests(new Hierarchy_Settings__c(Prevent_Case_Deletion__c = False));
+        UTIL_CustomSettingsFacade.getSettingsForTests(new Hierarchy_Settings__c(Prevent_Case_Deletion__c = FALSE));
 
         Test.startTest();
             List<Case> casesToDelete = [SELECT Id FROM Case];
-            Database.DeleteResult[] results = Database.delete(casesToDelete, false);
+            Database.DeleteResult[] results = Database.delete(casesToDelete, FALSE);
         Test.stopTest();
 
         // Ensure all Cases are deleted
--- a/src/classes/CASE_CannotDelete_TEST.cls
+++ b/src/classes/CASE_CannotDelete_TEST.cls
@@ -1,0 +1,109 @@
+/*
+    Copyright (c) 2020, Salesforce.org
+    All rights reserved.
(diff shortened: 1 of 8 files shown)


src/classes/CASE_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,109 @@
...
+        List<Case> casesAfterDelete = [SELECT Id FROM Case];
+        System.assertEquals(2, casesAfterDelete.size());
+    }
+
+    /*************************************************************************************************************
+    * @description Test deletion of Cases with Behavior Involvement and preventing deletion setting is turned Off
+    */
+    @isTest
+    static void testCaseDeleteSettingOff() {
+        UTIL_CustomSettingsFacade.getSettingsForTests(new Hierarchy_Settings__c(Prevent_Case_Deletion__c = False));
```
lifewithryan: NIT: very minor, we should probably as a team figure out if we want to do true|false, True|False, or TRUE|FALSE, but we probably should be consistent.

skosgi1: @lifewithryan Thank you. I fixed it, can you give it another shot?

spinkelman:
Passes QA - Will merge in the morning when more people are around incase merging causes issues. (which it shouldn't)

PR #1103, opened 2020-02-24
[EDA] Prevent Deletion of a Behavior Involvement Record when there is a Related Behavior Response Record

### LINKED ISSUE #1102: [EDA] Prevent Deletion of a Behavior Involvement Record when there is a Related Behavior Response Record

Feature
When users delete a Behavior Involvement with child records, the system will allow users to delete the Behavior Involvement and it will not delete child records.
We want to provide customers the ability to prevent deletion of Behavior Involvement if it has related child records by allowing them to toggle a checkbox in the Hierarchy Setttings.
By giving customers the ability to control whether a  Behavior Involvement with child records should be deleted or not, it will help them with maintaining a clean system.
* **Acceptance Criteria**

  - **Given**: If Prevent_Course_Offering_Deletion__c checkbox is enabled in Hierarchy Settings
  - **When**: I try to delete a Course Connection record with child records
  - **Then**: I should not be able to delete the Course Connection record.

kdy618:
# Critical Changes

# Changes
We've added functionality to prevent the deletion of Behavior Involvement records that have related Behavior Response records. This change is for internal-use only and is disabled at this time. You won't see any difference in functionality.

# Issues Closed
Closes #1102

# New Metadata

# Deleted Metadata
BEH_CannotDelete_TDTM
BEH_CannotDelete_TEST

# Testing Notes
Scenario 1:  “Prevent Behavior Involvement Deletion” checkbox is enabled and Behavior Involvement has related Behavior Response child record.
- Check “Prevent Behavior Involvement Deletion” checkbox in Hierarchy Settings.
- Create multiple Contact records  
- Create multiple Case records and associate the Contacts to the new Case records 
- Create multiple Behavior Involvement records and associate the Contacts and Cases created previously
- Create multiple Behavior Responses and associate to the Behavior Involvements 
- Delete Behavior Involvement records
- Result: You should receive a pop up error that you cannot delete this Behavior Involvement records

Scenario 2: “Prevent Behavior Involvement Deletion” checkbox is enabled and Behavior Involvement has NO related Behavior Response child record.
- Check “Prevent Behavior Involvement Deletion” checkbox in Hierarchy Settings.
- Create multiple Contact records  
- Create multiple Case records and associate the Contacts to the new Case records 
- Create multiple Behavior Involvement records and associate the Contacts and Cases created previously
- Delete Behavior Involvement records
- Result: You not receive a pop up error and should be able to delete the Behavior Involvement records. 

Scenario 3:  “Prevent Behavior Involvement Deletion” checkbox is disabled and Behavior Involvement has related Behavior Response child record.
- Check “Prevent Behavior Involvement Deletion” checkbox in Hierarchy Settings.
- Create multiple Contact records  
- Create multiple Case records and associate the Contacts to the new Case records 
- Create multiple Behavior Involvement records and associate the Contacts and Cases created previously
- Create multiple Behavior Responses and associate to the Behavior Involvements 
- Delete Behavior Involvement records
- Result: You not receive a pop up error and should be able to delete the Behavior Involvement records. 

Scenario 4: “Prevent Behavior Involvement Deletion” checkbox is disabled and Behavior Involvement has NO related Behavior Response child record.
- Check “Prevent Behavior Involvement Deletion” checkbox in Hierarchy Settings.
- Create multiple Contact records  
- Create multiple Case records and associate the Contacts to the new Case records 
- Delete Behavior Involvement records
- Result: You not receive a pop up error and should be able to delete the Behavior Involvement records.


src/classes/BEH_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,156 @@
...
+        
+        List<Behavior_Involvement__c> returnBEH = [SELECT Id
+                                                   FROM Behavior_Involvement__c
+                                                   WHERE Id IN :bEHInvolvments]; 
+        
+        //Verifies that only 2 Behavior Involvements were deleted 
+        System.assertEquals(2, returnBEH.size()); 
+        
+        //Verifies that the Behavior Involvement records with Behavior Response were not deleted
+        System.assertEquals(Label.CannotDelete, results[0].errors[0].message);
```
lifewithryan: Are we guaranteed here that it will always be results[0]? May wanna try something like our map trick here as well.

src/classes/BEH_CannotDelete_TDTM.cls
```diff
@@ -0,0 +1,83 @@
...
+public with sharing class BEH_CannotDelete_TDTM extends TDTM_Runnable{
+    /*******************************************************************************************************
+    * @description Get the setting of preventing behavior involvement deletion
+    */
+    private static Boolean enabledPreventBehaviorInvolvementDeletion = UTIL_CustomSettingsFacade.getSettings().Prevent_Behavior_Involvement_Deletion__c;
+    
+    /*******************************************************************************************************
+     * @description Stops a Behavior Involvement from being deleted
+     * if it has any associated Behavior Response. 
+     * @param listNew the list of Course Offering from trigger new.
```
lifewithryan: Looks like some bad copy/pasta here, I don't think we're touching Course Offering stuff here.

src/classes/BEH_CannotDelete_TEST.cls
kdy618: @lifewithryan , for this particular scenario, I've created 4 BEH and 2 of them have BR records associated and 2 does not. Since I've asserted on 108 that the return query of BEH after the delete DML is 2, the order on line 111 and 112 assert should not necessarily matter since we're doing the same assumption that both records exist. Please let me know if you think it is still worth being precise on this scenario if the assert we're testing is that the 2 returned BEH records exist and we're not doing any data comparison/check.

lifewithryan: @kdy618 -- I think essentially the additional check is probably superfluous and we could move along without it. Asserting that we only have 2 left instead of four essentially verifies our expected outcome. The other was just a nice to have and appreciate that you added it.
I will leave it up to you. I think it is fine without but if the developer within you would still like to polish it off, feel free :)

spinkelman:
Passes QA - Merging
Closing WI

PR #955, opened 2019-10-01
W-035944 - Add new custom flow and custom tasks to enable Shield Platform Encryption

robertbwatson:
A new custom flow and three new custom tasks were developed for EDA in order to setup Shield Platform Encryption more easily in a developer scratch org.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
1. Create a new developer scratch org.
2. After deploying EDA to the scratch org (using the `dev_org` or `qa_org` flow), run the `setup_encryption` flow.
3. Open the org and navigate to *Setup | Platform Encryption | Encryption Policy | Encrypt Fields*
4. Enable all standard fields for encryption.
    a. Optionally, enable EDA custom fields for encryption.
5. Run the EDA unit tests. Confirm they all pass.

(See Quip doc in WI for more details)

PR #1109, opened 2020-02-28
[EDA] Pushing code for ryan to look at unsure on validation however i…

jofsky:
…t did successfully run

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


unpackaged/config/platformencryption_fields/package.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lifewithryan: @jofsky Is the thought here that these field should be marked as encrypted when we run this flow? If so, if we wanted all of them, we can just add to this?

unpackaged/config/platformencryption_fields/objects/Contact.object
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lifewithryan: @jofsky and we'll have to add the fields/objects we want to encrypt to this directory? If this works, that is going to be friggin' awesome.

cumulusci.yml
```diff
@@ -97,6 +97,40 @@ tasks:
...
+
+    platformencryption_deploy_permset:
+        class_path: cumulusci.tasks.salesforce.Deploy
+        options:
+            path: unpackaged/config/platformencryption_permset
+            namespace_inject: $project_config.project__package__namespace
+
+    platformencryption_assign_permset:
+        class_path: cumulusci.tasks.apex.anon.AnonymousApexTask
+        options:
```
jstvz: Can we use `SFDXOrgTask` instead? See: `sfdx force:user:permset:assign --help`

cumulusci.yml
```diff
@@ -254,7 +300,20 @@ flows:
                     include_beta: True
             5:
                 task: install_managed_beta
-
+    new_org:
```
jstvz: Looks like you based this on `feature/upgraded-org-docs`, so you've included some unrelated changes.

cumulusci.yml
```diff
@@ -97,6 +97,40 @@ tasks:
...
+                    AssigneeId = UserInfo.getUserId()
+                );
+
+    platformencryption_create_tenant_secret:
+        description: 'Create a placeholder Tenant Secret to use Platform Encryption'
+        class_path: cumulusci.tasks.sfdx.SFDXOrgTask
+        options:
+            command: 'force:data:record:create -s TenantSecret -v "Description=test"'
+
+    platformencryption_deploy_field_configuration:
```
```suggestion
    platformencryption_deploy_encrypted_fields:
```

cumulusci.yml
```diff
@@ -198,6 +232,18 @@ flows:
                     namespaced_org: True
                     skip_record_types: True
 
+    config_platformencryption:
```
robertbwatson: @jofsky @lifewithryan This caught my eye - just wanted to see if any of this was duplicate work of what I already implemented in #955? Some of it looks the same, some of it potentially new. There's a `setup_encryption` flow and `deploy_encryption_permset`, `assign_encryption_permset`, & `create_tenant_secret` tasks.

lifewithryan: Yea that's probably duplicate, Josh's items also then configures the encrypted fields for us as well so we don't have to touch anything manually. So we will want to either leverage your tasks or remove yours and just leverage Josh's. I would suggest just leverage Josh's now since he's got pretty much a soup to nuts implementation.
I apologize for not being clearer that we already had automation to setup the key, etc.

jstvz: I'd suggest we use the existing tasks/flow, and append `platformencryption_deploy_field_configuration` to `setup_encryption`.

robertbwatson: I see it also already documented in https://github.com/SalesforceFoundation/EDA/blob/master/documentation/automation.md -- we could consider expanding on that documentation a bit, perhaps.

There's also a Quip doc I spun up a few months ago: https://salesforce.quip.com/hd1UAWXDUqJB

lifewithryan:
@jstvz -- this is saying we can't use standard fields? This does us no good then until we can get around that. Is there a reason the PortalsEnabled feature is turned on? Perhaps for now we as devs will just need to run this as dev scratch orgs to see if things blow up. Ideally we would need all the other "stuff" enabled, but we want to ensure that the code we write doesn't include encrypted fields in places it shouldn't. This includes standard fields.

jstvz:
@lifewithryan:

> @jstvz -- this is saying we can't use standard fields? 

In Enterprise editions, yes. The problem is that the template org EE scratch orgs are based on has legacy portals enabled. We can use DE scratch orgs as a workaround until Spring '20.

xinnanlisf:
@jofsky or @jstvz, just a quick follow up on this one, do you guys have any update on this PR?

PR #1115, opened 2020-03-09
add for loop to robot file to make it less verbose

jessingrass:
# Critical Changes

# Changes
Pushing this up for general review - not ready to be merged, should not be merged right now.
# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


robot/EDA/resources/locators.py
```diff
@@ -159,13 +159,13 @@
...
     "account_types": {
         "administrative": "//span[contains(text(),'Administrative')]/parent::*",
-        "academic": "//div[contains(@class,'slds-form-element__control')]//span[contains(text(),'Academic Program')]",
-        "administrativecheckbox": "//div[contains(@class,'slds-form-element__control')]//span[contains(text(),'Administrative')]",
-        "business": "//div[contains(@class,'slds-form-element__control')]//span[contains(text(),'Business Organization')]",
-        "educational": "//div[contains(@class,'slds-form-element__control')]//span[contains(text(),'Educational Institution')]",
-        "household": "//div[contains(@class,'slds-form-element__control')]//span[contains(text(),'Household Account')]",
-        "sports": "//div[contains(@class,'slds-form-element__control')]//span[contains(text(),'Sports Organization')]",
-        "university": "//div[contains(@class,'slds-form-element__control')]//span[contains(text(),'University Department')]",
+        # "academic": "//div[contains(@class,'slds-form-element__control')]//span[contains(text(),'Academic Program')]",
```
boakley: If we're not going to use these old locators, you might was well delete them rather than just comment them out. If we decide we need them again, we can find them in the git history of the file.

robot/EDA/tests/browser/settings/eda_settings_accounts_and_contacts_edit_mode.robot
```diff
@@ -21,68 +27,37 @@ Validate Edit Mode For Accounts and Contacts Settings
...
-    Click on Element                account_types.sports
-
-    Wait for Locator                account_types.university 
-    Click on Element                account_types.university 
-'''
- 
-    # Click Button                  Save (in page object for this test)
+Click Checkboxes
+    Test Checkboxes                 Academic Program
+    Test Checkboxes                 Administrative
```
boakley: A minor quibble: since you're using this keyword to check a single checkbox now that you aren't using a template, it might make more sense to call this "Test checkbox" (singular)

Otherwise, this test is definitely a big improvement in readability over the original version. 👍

robot/EDA/resources/locators.py
jessingrass: Yes! I just commented them out for testing, in case my code didn't work. Now that it does, I'll remove.

PR #1119, opened 2020-03-11
[EDA] Prevent Deletion of Program Plan Record if There are Associated Child Records

### LINKED ISSUE #1120: Updates to PPlan_CannotDelete_TDTM.cls to utilize Prevent_Program_Plan_Deletion__c hierarchy settings
Feature

When a user tries to delete a Program Plan that has child records, the system prevents it from deleting. The feature is working fine, however, it doesn't take into consideration the 'Prevent_Program_Plan_Deletion__c' flag when preventing the delete.

Take into consideration the value of 'Prevent_Program_Plan_Deletion__c' when trying to delete the Program Plan record with associated child records. This checkbox should be turned on for new and existing Customers by default. 

Consistency with other CannotDeleted TDTMs

* **Acceptance Criteria**

  - **Given**: There are Program Plans with child record association and 'Prevent_Program_Plan_Deletion__c' hierarchy custom setting is enabled.
  - **When**: When the delete button is clicked 
  - **Then**: The Program Plan record should not be deleted.

  - **Given**: There are Program Plans with child record association and 'Prevent_Program_Plan_Deletion__c' hierarchy custom setting is not enabled.
  - **When**: When the delete button is clicked 
  - **Then**: The Program Plan record should be deleted.

skosgi1:
# Critical Changes

# Changes
We've expanded the functionality that prevents the deletion of certain records when they have related child records. Now, the following record types cannot be deleted if they have the following related records.

- Program Plan records can't be deleted if they have related Program Enrollment or Plan Requirement records.

This functionality is for internal use only and is enabled in Hierarchy Custom Settings for these objects by default in new and existing orgs. If you've disabled any of the object _CannotDelete_TDTM classes for these objects in an existing org, they will remain disabled even if they are enabled in Hierarchy Custom Settings.
**Note:** We strongly recommend that EDA Admins don't manage Hierarchy Custom Settings directly and, instead, use the EDA Settings UI, which will be provided in an upcoming release.

# Issues Closed
Closes #1120 

# New Metadata

# Deleted Metadata

# Testing Notes
Scenario 1: (Before merged into Master)

1) Ensure  'Prevent Program Plan Deletion' is defaulted to TRUE on scratch org creation i.e., Set up -> Custom Settings -> Hierarchy Settings -> Prevent Program Plan Deletion -> Should've been set to TRUE
2) Create a Program Plan
3) Create Plan Requirement and/or Program Enrollment as child records to the Program Plan created in step 2
4) Delete the Program Plan created in step 2
5) You should receive the error message https://salesforce.quip.com/KmXZALIhSsgW#ZYaACAcCF8m
6) Now, uncheck the 'Prevent Program Plan Deletion' check box from step 1
7) Delete the Program Plan created in step 2
8) The Program plan should be deleted

Scenario 2: (After Merged into Master)
We're utilizing install script to set 'Prevent Program Plan' Hierarchy Settings to TRUE for both existing users and new users. It would be nice to test if that this is reflected in a Push upgrade. May be using upgraded_org flow. 

Note: The install script was updated assuming this would go into release 1.96. If it doesn't make into 1.96 the install script should be updated accordingly.


cumulusci.yml
```diff
@@ -78,6 +78,7 @@ tasks:
                 STG_InstallScript.insertRelationshipLookups();
                 TDTM_Global_API.setTdtmConfig(TDTM_Global_API.getDefaultTdtmConfig(), 'hed');
                 STG_InstallScript.enablePreferredPhoneSync();
+                STG_InstallScript.enablePreventDeletionHierarchyCustomSettings(NULL);
```
xinnanlisf, line 81: Not a big fan of passing "NULL" as parameters...

src/classes/STG_InstallScript.cls
```diff
@@ -332,4 +337,19 @@ global without sharing class STG_InstallScript implements InstallHandler {
             UTIL_OrgTelemetry.dispatchTelemetryBatchJob();
         }
     }
+    
+    /**************************************************************************************************************************
+    * @description Enables prevent deletion by default on Hierarchy Custom Settings for new installs and when upgrading to 1.95
+    * @return void
+    */
+    global static void enablePreventDeletionHierarchyCustomSettings(Version previousVersion) {
+        if ((previousVersion == NULL) || (previousVersion != null && previousVersion.compareTo(new Version(1,95)) < 0)) {
```
xinnanlisf, line 346: Let's make sure we are consistent with "NULL" or "null"

xinnanlisf: Also, quick question, is there a reason behind not using custom setting facade. configSetting()?

cumulusci.yml
skosgi1: @xinnanlisf For this we are utilizing same method for both new installs and upgrades so this way it is reusable, wanted to check if you have any other suggestions?

src/classes/STG_InstallScript.cls
skosgi1: Hmm looks like that's how we have been using for other methods too like populateAdminRecordTypeSetting, setDefaultContactLanguageFluency @xinnanlisf

src/classes/STG_InstallScript.cls
```diff
@@ -74,8 +74,8 @@ global without sharing class STG_InstallScript implements InstallHandler {
             populateAdminRecordTypeSetting(previousVersion);
 
             setDefaultContactLanguageFluency(previousVersion);
-
-            enablePreventDeletionHierarchyCustomSettings(previousVersion); 
+
```
xinnanlisf, line 77: shall we remove these indentations?

xinnanlisf: Also, there is some other funky indentations in this class that i noticed. Could we clean them up using this chance?

src/classes/STG_InstallScript.cls
```diff
@@ -338,20 +338,21 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
     * @return void
     */
     global static void enablePreventDeletionHierarchyCustomSettings(Version previousVersion) {
-        if ((previousVersion == NULL) || (previousVersion != null && previousVersion.compareTo(new Version(1,95)) < 0)) {
+        if ((previousVersion == NULL) || (previousVersion != NULL && previousVersion.compareTo(new Version(1,95)) < 0)) {
             UTIL_Debug.debug('***Upgrading to version 1.95 or new installs');
             Hierarchy_Settings__c orgSettings = UTIL_CustomSettingsFacade.getSettings();
-            if (orgSettings.Prevent_Plan_Requirement_Deletion__c == null || orgSettings.Prevent_Plan_Requirement_Deletion__c != true) {
-                orgSettings.Prevent_Plan_Requirement_Deletion__c = true; 
+            if (orgSettings.Prevent_Plan_Requirement_Deletion__c == NULL || orgSettings.Prevent_Plan_Requirement_Deletion__c != TRUE) {
```
xinnanlisf, line 349: I don't think a boolean custom setting can be null thou...

cumulusci.yml
xinnanlisf: right, by using custom setting facade. configSetting(), we can set up the default value for custom settings. This will only happen for new installs since configSetting() will only run when context.previousVersion() == null.

xinnanlisf:
TODO: @skosgi1 to update the version number for custom setting populating.

src/classes/STG_InstallScript.cls
skosgi1: Resolving, discussed offline.

src/classes/STG_InstallScript.cls
skosgi1: did some minor changes in the class, let me know if this looks good.

skosgi1:
> TODO: @skosgi1 to update the version number for custom setting populating.

@xinnanlisf I have not updated the version number yet, let me know if you don't see this making to 1.95. will then update accoirdingly TIA.

xinnanlisf:
> > TODO: @skosgi1 to update the version number for custom setting populating.
> 
> @xinnanlisf I have not updated the version number yet, let me know if you don't see this making to 1.95. will then update accoirdingly TIA.

yep just a reminder for us if it has to slip to our next release.

src/classes/STG_InstallScript.cls
skosgi1: updated based on feedback.

cumulusci.yml
skosgi1: Got it! thanks for the clarification. Updated based on feedback.

src/classes/STG_InstallScript.cls
```diff
@@ -338,24 +343,37 @@ global without sharing class STG_InstallScript implements InstallHandler {
         }
     }
     
-    /**********************************************************************************************************
-    * @description Enables prevent deletion by default on Hierarchy Custom Settings for new installs and 
-    * when upgrading to 1.95
+    /**************************************************************************************************************************
+    * @description Enables prevent deletion by default on Hierarchy Custom Settings for new installs and when upgrading to 1.95
```
xinnanlisf, line 347: One small thing, I know this has been merged already but could we make the description a little bit specific. After all, they are all enabling deletion prevention?

skosgi1: good catch, updated @xinnanlisf

src/classes/STG_InstallScript.cls
```diff
@@ -338,24 +345,37 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+            upsert orgSettings;
+        }
+    }
+
+    /*****************************************************************************************************************************************
+    * @description Enables Prevent Program Plan Deletion by default on Hierarchy Custom Settings for new installs and when upgrading to 1.96
+    * @return void
+    */
+    global static void enablePreventPPlanDeletionHierarchyCustomSettings(Version previousVersion) {
+        if ((previousVersion == null) || (previousVersion != null && previousVersion.compareTo(new Version(1,96)) < 0)) {
```
kdy618, line 372: @skosgi1 , I was wondering and maybe this is not a great way of thinking because we want to break things down modularly that 1 method does 1 thing, that instead of creating a 2nd enablePreventPPlanDeletionHierarchyCustomSettings method, that we add the if statement that checks if it is version 1.96 to enablePreventDeletionHierarchyCustomSettings method so that we only have 1? But again, I can see the argument that maybe we want 1 method to do 1 thing. @MatthewBlanski

kdy618: [enablePreventDeletionMethods.txt](https://github.com/SalesforceFoundation/EDA/files/4504932/enablePreventDeletionMethods.txt)

src/classes/PPlan_CannotDelete_TDTM.cls
```diff
@@ -46,18 +50,21 @@ public with sharing class PPlan_CannotDelete_TDTM extends TDTM_Runnable {
...
         if (triggerAction == TDTM_Runnable.Action.BeforeDelete) {
-
             for (Program_Plan__c programPlan : [SELECT ID,
                                                         (SELECT ID FROM Plan_Requirements__r LIMIT 1),
                                                         (SELECT ID FROM Program_Enrollments__r LIMIT 1)
                                                 FROM Program_Plan__c
                                                 WHERE ID IN :oldlist]) {
-
-                if(programPlan.Plan_Requirements__r.size() > 0
+                if (programPlan.Plan_Requirements__r.size() > 0
```
MatthewBlanski, line 67: This is more of a stylistic thought, but have we considered using List.isEmpty() and ! List.isEmpty() instead of size > 0? It does deviate from our prior code, but it feels more human-readable in how it is being explicit about what we are trying to have the code do.

lifewithryan: I usually prefer isEmpty() calls, however I haven't in the past really ever planted a flag anywhere. I like that we are thinking about it. The only thing I've noticed in the past is when calling !Somelist.isEmpty() that it is sometimes easy for eyes to miss the "!" -- all that said though, the context when reviewing said code block helps ensure that you're looking for it more closely when you see the isEmpty()

src/classes/STG_InstallScript.cls
```diff
@@ -338,24 +345,37 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
-    /**********************************************************************************************************
-    * @description Enables prevent deletion by default on Hierarchy Custom Settings for new installs and 
-    * when upgrading to 1.95
+    /************************************************************************************************************************************************************************
+    * @description Enables Prevent Account Deletion and Plan Requirement Deletion by default on Hierarchy Custom Settings for new installs and when upgrading to 1.95
     * @return void
     */
     global static void enablePreventDeletionHierarchyCustomSettings(Version previousVersion) {
-        if ((previousVersion == NULL) || (previousVersion != null && previousVersion.compareTo(new Version(1,95)) < 0)) {
+        if ((previousVersion == null) || (previousVersion != null && previousVersion.compareTo(new Version(1,95)) < 0)) {
```
MatthewBlanski, line 353: I don't feel that we need the previousVersion != null check in the logic as in the logic flow previousVersion equals null is evaluated first, which would evaluate to true and proceed to the code in the block. Any condition after that is implicitly previousVersion not equals null. I realize this is a style fix, but thought it worth the comment. Not a required change.

MatthewBlanski: After a further review on this, I feel like we might be well-served by having a private static helper method that performs this kind of check consistently and makes the code more readable, taking the version as an argument.

src/classes/STG_InstallScript.cls
MatthewBlanski: After looking over the code and getting some context, I feel like since we have enablePreventDeletionHierarchyCustomSettings and it is global, we should leverage the method for how it is named. This aligns more closely with option 2 from what Kim provided. What I would prefer to see would be blocks for the 1.95 and 1.96 version that then contain individual methods for each discrete setting change and return the modified settings. Having the upsert be in the global method seems preferable in this case and should improve testability.

src/classes/PPlan_CannotDelete_TDTM.cls
MatthewBlanski: Another perfectly valid way to make it more readable would be List.isEmpty() == false, which may be even more readable!

skosgi1:
@MatthewBlanski  Updated the code based on your feedback, can you please give it another shot? Thx.

src/classes/STG_InstallScript.cls
```diff
@@ -337,25 +338,62 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+            orgSettings = enablePreventPPlanDeletionHierarchyCustomSettings();
+        }
+        upsert orgSettings;
+    }
+
+    /*****************************************************************************************************************************************************************
+    * @description Enables Prevent Account Deletion and Plan Requirement Deletion by default on Hierarchy Custom Settings for new installs and when upgrading to 1.95
+    * @return void
+    */
+    global static Hierarchy_Settings__c enablePreventForAccountsAndPlanRequirement() {
```
MatthewBlanski, line 365: Please split this out into two methods, one for account deletion and one for plan requirement deletion? I'd prefer that we have one method per task so we can pinpoint any errors that might occur.

Please also add unit tests for those methods.

src/classes/STG_InstallScript.cls
```diff
@@ -337,25 +338,62 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
-    * when upgrading to 1.95
+
+    /*********************************************************************************************************
+    * @description Enables Prevent Hierarchy Custom Settings for new installs and push upgrades
     * @return void
     */
     global static void enablePreventDeletionHierarchyCustomSettings(Version previousVersion) {
-        if ((previousVersion == NULL) || (previousVersion != null && previousVersion.compareTo(new Version(1,95)) < 0)) {
+        Hierarchy_Settings__c orgSettings = UTIL_CustomSettingsFacade.getSettings();
+        Boolean isNewInstallOrPushUpgradedTo195 = newInstallOrPushUpgrade(previousVersion, new Version(1,95));
```
MatthewBlanski, line 348: This variable naming is delightfully explicit as to its purpose.

src/classes/STG_InstallScript.cls
```diff
@@ -337,25 +338,62 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+        }
+        return orgSettings;
+    }
+
+    /*****************************************************************************************************************************************
+    * @description Method to check if the package is being newly installed or push upgraded to a particular version
+    * @return true if it is either a new install or push upgraded to certain version else return false
+    */
+    @testVisible
+    private static boolean newInstallOrPushUpgrade(Version previousVersion, Version newVersion) {
```
MatthewBlanski, line 393: Consider renaming this to isNewInstallOrPushUpgrade or something similar to make it clear that this is more of a check than a function that performs changes.

src/classes/STG_InstallScript.cls
```diff
@@ -337,25 +338,62 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+            orgSettings.Prevent_Plan_Requirement_Deletion__c = true;
+        }
+        return orgSettings;
+    }    
+
+    /*****************************************************************************************************************************************
+    * @description Enables Prevent Program Plan Deletion by default on Hierarchy Custom Settings for new installs and when upgrading to 1.96
+    * @return void
+    */
+    global static Hierarchy_Settings__c enablePreventPPlanDeletionHierarchyCustomSettings() {
```
MatthewBlanski, line 380: Please add a unit test for this method.

skosgi1:
> This is a step in a good direction. The variable naming is well fit to their purpose, making the code easier to read. I've requested a few changes, mostly in splitting one method a step further and in adding unit tests for the new methods.

Thanks @MatthewBlanski can you please give it another go?

src/classes/STG_InstallScript.cls
```diff
@@ -337,25 +338,72 @@ global without sharing class STG_InstallScript implements InstallHandler {
...
+
+    /**************************************************************************************************************
+    * @description Enables Prevent Hierarchy Custom Settings for new installs and push upgrades
     * @return void
     */
     global static void enablePreventDeletionHierarchyCustomSettings(Version previousVersion) {
-        if ((previousVersion == NULL) || (previousVersion != null && previousVersion.compareTo(new Version(1,95)) < 0)) {
+        Hierarchy_Settings__c orgSettings = UTIL_CustomSettingsFacade.getSettings();
+        Boolean isNewInstallOrPushUpgradedTo195 = isNewInstallOrPushUpgrade(previousVersion, new Version(1,95));
+        Boolean isNewInstallOrPushUpgradedTo196 = isNewInstallOrPushUpgrade(previousVersion, new Version(1,96));
```
skosgi1, line 349: @MatthewBlanski if there are any Customers still on 1.94 and when they upgrade to 1.95 'isNewInstallOrPushUpgrade(previousVersion, new Version(1,96)' would return true because 1.94 < 1.96 as a result 'Prevent_Program_Plan_Deletion__c' is marked as true along with 'Prevent_Plan_Requirement_Deletion__c' and 'Prevent_Account_Deletion__c'. Not sure if it is a valid scenario and if so it is ok to make 'Prevent_Plan_Requirement_Deletion__c' to true in that case so wanted to check with you. TIA.

MatthewBlanski: @skosgi1 Since the intent behind this is to handle upgrade scenarios without being disruptive to a production org, this would be desired behavior as we are effectively defaulting these prevent deletion settings to true. Theoretically, if someone were capable of upgrading from version 1.0 all the way to the current version, we would want the entire run to process to bring the org up to speed.

spinkelman:
Passes QA - Merging
Will Close WI

PR #1152, opened 2020-04-08
[EDA] Update the API version number EDA wide

m-kinkade:
# Critical Changes

# Changes
We've updated EDA to use API Version v48.0 to align with the most current version of the API. Additionally, we have removed all references to the deprecated mobileReady field from all customTab object files. You'll see no change to your org as a result of this update.

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/ADDR_Seasonal_BATCH.cls-meta.xml
+++ b/src/classes/ADDR_Seasonal_BATCH.cls-meta.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
-    <apiVersion>45.0</apiVersion>
+    <apiVersion>48.0</apiVersion>
     <status>Active</status>
 </ApexClass>
--- a/src/tabs/Course__c.tab
+++ b/src/tabs/Course__c.tab
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <CustomTab xmlns="http://soap.sforce.com/2006/04/metadata">
     <customObject>true</customObject>
-    <mobileReady>false</mobileReady>
     <motif>Custom83: Pencil</motif>
 </CustomTab>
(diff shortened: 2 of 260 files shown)


m-kinkade:
> Hey @m-kinkade looks like we have versions in package.xml too, there are a few package.xml files in src and unpackaged.

@skosgi1 I've updates those files as requested and removed references to <mobileReady> as well.

src/tabs/Course__c.tab
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <CustomTab xmlns="http://soap.sforce.com/2006/04/metadata">
     <customObject>true</customObject>
-    <mobileReady>false</mobileReady>
```
jofsky, line 4: @m-kinkade @skosgi1 whats the reason for this to reenable, disabled  mobile functionality?

src/classes/ADDR_Seasonal_BATCH.cls-meta.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
-    <apiVersion>45.0</apiVersion>
+    <apiVersion>48.0</apiVersion>
```
jofsky, line 3: @m-kinkade  @skosgi1 @jstvz The maintenance of this seems very tedious, have we considered writing a task to inject the api version? I'm happy to work with you @skosgi1 and @m-kinkade  to demonstrate what I mean. (@jstvz  just tagging for your two cents.)

src/tabs/Course__c.tab
skosgi1: @jofsky looks like that is not supported in api 48.0 https://success.salesforce.com/answers?id=9063A000000a5FKQAY

src/classes/ADDR_Seasonal_BATCH.cls-meta.xml
dgamboni: @jofsky Sounds like a great idea!

dgamboni:
Passes smoke test.

PR #1157, opened 2020-04-14
Feature/api versioning

jofsky:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -4,7 +4,7 @@ project:
     package:
         name: EDA
         namespace: hed
-        api_version: 45.0
+        api_version: 48.0
         install_class: STG_InstallScript
     git:
         prefix_release: rel/
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -170,6 +170,19 @@ flows:
             2:
                 task: uninstall_packaged
 
+    beta_dependencies:
+        steps:
+            2:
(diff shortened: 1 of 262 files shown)


jofsky:
@davisagli this was my concern, just wanted a way for the team to not hard code 256 files. But i understand.

lifewithryan:
@jofsky @davisagli I think all we really do is global search/replace in our IDEs anyway. So this is appreciated but perhaps overhead and if there are concerns in other areas, we should probably just move on. One of our devs did notice however that once we run that flow, it seemed to persist the API numbers to disk anyway, so its probably not quite read for prime time as if this is something we were going to do, it shouldn't persist, it should modify in transit.
I think we can close this perhaps?

jofsky:
Thanks for the discussion @davisagli @lifewithryan sorry for wasting your time. @davisagli can I work on moving the Find replace task to CCI I know we had discussed the possibility of you creating one, wanted to check to see where you were at on that/if I could help.

davisagli:
@jofsky please look at the existing FindReplace task https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/tasks/util.py#L125 and create a github issue or work item describing what changes you'd like to make to it.

PR #1165, opened 2020-04-27
[EDA] Prevent Deletion of Course Record  if There are Associated Course Records

kdy618:
# Critical Changes

# Changes
## Updates to Ensure Data Integrity
We've added functionality to prevent the deletion of Course records with related Course Offering or Plan Requirement records. 

These changes are for internal-use only and are disabled at this time. You won't see any difference in functionality.

**Note:** We strongly recommend that EDA Admins don't manage Hierarchy Custom Settings directly and, instead, use the EDA Settings UI, which will be provided in an upcoming release. However, if you do enable any of these settings, note the following:

- If you have disabled the related COUR_CannotDelete_TDTM classes for the object, the settings will remain disabled even though you've enabled them in Hierarchy Custom Settings.
- If you enable any of these settings, be sure to disable them before merging records or deleting duplicate records.
# Issues Closed
Closes #1164
# New Metadata
Checkbox Field on Hierarchy Settings:
- Prevent_Course_Deletion__c

Apex Classes: 
- COUR_CannotDelete_TDTM
- COUR_CannotDelete_TEST
# Deleted Metadata

# Testing Notes
https://salesforce.quip.com/tpgtAkYA9bi9#KLfACAC3QG7

PR #1169, opened 2020-04-29
[EDA] Prevent Deletion of Address Record if There are Associated Child Records

### LINKED ISSUE #1163: [EDA] Prevent Deletion of Address Record if There are Associated Child Records

Feature

When a user attempts to delete an Address record that is associated with child records, the system deletes the address record, leaving the child Contact and/or Account records intact with the Current Address field cleared.

When a user attempts to delete an Address record that is associated with child records, the system will utilize a new custom setting field (Prevent Address Deletion) to determine whether to allow the deletion.  If selected, attempts to delete an Address record with associated child records will be blocked and an error message displayed to the user.  If unchecked, the functionality will remain the same as outlined in the current behavior section.

From previous github requests and investigations of deletion of parent records with child records, it was determined that we need to implement the functionality to allow users the ability to control if an Address record with related child record can be deleted. 

* **Acceptance Criteria**

  - **Given**: The “Prevent Address Deletion” checkbox in Hierarchy Settings (Custom Setting) is unchecked.
  - **When**: A user clicks “Delete” button on an Address record that has child related records.
  - **Then**: The Address record is successfully deleted, no error message is displayed to the user and the child records are not deleted.

  - **Given**: The “Prevent Address Deletion” checkbox in Hierarchy Settings (Custom Setting) is checked.
  - **When**: A user clicks “Delete” button on an Address record that has child related records.
  - **Then**: An error message is displayed to the user and the user is unable to delete the Address record.

m-kinkade:
# Critical Changes

# Changes
We've added functionality to prevent the deletion of Address records with related Account or Contact records.

- If you have disabled the related ADDR_CannotDelete_TDTM classes for the object, the settings will remain disabled even though you've enabled them in Hierarchy Custom Settings.

# Issues Closed
Closes #1163  

# New Metadata

- Prevent_Address_Deletion__c

Apex Classes:

- ADDR_CannotDelete_TDTM

- ADDR_CannotDelete_TEST

# Deleted Metadata

# Testing Notes
https://salesforce.quip.com/TPlxAVsuZhl9#NOMACAHbV0K


src/classes/ADDR_CannotDelete_TDTM.cls
```diff
@@ -0,0 +1,89 @@
...
+        return new DmlWrapper();
+    }
+    
+    /*******************************************************************************************************
+     * @description Evaluates whether the Address has any child (Account or Contact) related records.
+     * @param a is the current Address record.
+     * @return Boolean.
+     ********************************************************************************************************/
+    private static Boolean hasChildRecords(Address__c a) {
+
```
skosgi1, line 85: Can you please remove static here ref: https://github.com/SalesforceFoundation/EDA/pull/1165#discussion_r416075994, add a unit test method to test this method by adding @testVisible and also change the parameter name from 'a'?

src/classes/ADDR_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,452 @@
...
+        ];
+
+        System.assert(returnAddresses.isEmpty());
+
+        for (Database.DeleteResult result : results) {
+            System.assertEquals(TRUE, result.isSuccess());
+        }
+    }
+
+    /*********************************************************************************************************
```
m-kinkade, line 336: @skosgi1 I added the following three test cases (starting here) to test the hasChildRecords method in ADDR_CannotDelete_TDTM

src/classes/ADDR_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,452 @@
...
+
+        
+        System.assert(returnAddresses.isEmpty());
+
+        for (Database.DeleteResult result : results) {
+            System.assertEquals(TRUE, result.isSuccess());
+        }
+    }
+
+    /*********************************************************************************************************
```
m-kinkade, line 207: @skosgi1 this is a new test case

src/classes/ADDR_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,452 @@
...
+        ];
+
+        System.assertEquals(addrsWithChildContact.size(), returnAddresses.size());
+
+        for (Database.DeleteResult result : results) {
+            System.assertEquals(Label.CannotDelete, result.errors[0].message);
+        }
+    }
+
+    /*********************************************************************************************************
```
m-kinkade, line 109: @skosgi1 this is a new test case

src/classes/ADDR_CannotDelete_TDTM.cls
```diff
@@ -0,0 +1,90 @@
...
+        if (triggerAction == TDTM_Runnable.Action.BeforeDelete) {
+            for (Address__c addr : [
+                                    SELECT Id, 
+                                        (SELECT Id FROM Address__c.Accounts__r LIMIT 1),
+                                        (SELECT Id FROM Address__c.Contacts1__r LIMIT 1)
+                                    FROM Address__c WHERE Id IN :oldlist
+                                    
+                                ]) {
+                
+                if (this.hasChildRecords(addr)) {
```
skosgi1, line 69: this. 👍😊

src/classes/ADDR_CannotDelete_TDTM.cls
```diff
@@ -0,0 +1,90 @@
...
+        return new DmlWrapper();
+    }
+    
+    /*******************************************************************************************************
+     * @description Evaluates whether the Address has any child (Account or Contact) related records.
+     * @param addr is the current Address record.
+     * @return Boolean.
+     ********************************************************************************************************/
+    @testVisible
+    private Boolean hasChildRecords(Address__c addr) {
```
skosgi1, line 85: @MatthewBlanski something out of the scope of this PR, since we are/will be using hasChildRecords method in prevent delete PRs, should we consider moving this to its own class with switch or is it just an over kill?

MatthewBlanski: @skosgi1 For now, let's keep this inside this class. This would be a good candidate to move to an address service later downstream.

src/classes/ADDR_CannotDelete_TDTM.cls
```diff
@@ -0,0 +1,90 @@
...
+    * @param newlist the list of Addresses from trigger new. 
+    * @param oldlist the list of Addresses from trigger old. 
+    * @param triggerAction which trigger event (BeforeInsert, AfterInsert, etc.). 
+    * @param objResult the describe for Addresses 
+    * @return DmlWrapper.  
+    ********************************************************************************************************/
+    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist, 
+    TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+        
+        if (!enabledPreventAddressDeletion) {
```
MatthewBlanski, line 54:
```suggestion
        if (!ADDR_CannotDelete_TDTM.enabledPreventAddressDeletion) {
```
Please add the classname before referencing a static variable so we may have that reference explicitly understood.

src/classes/ADDR_CannotDelete_TDTM.cls
```diff
@@ -0,0 +1,90 @@
...
+    
+    /*******************************************************************************************************
+     * @description Evaluates whether the Address has any child (Account or Contact) related records.
+     * @param addr is the current Address record.
+     * @return Boolean.
+     ********************************************************************************************************/
+    @testVisible
+    private Boolean hasChildRecords(Address__c addr) {
+        
+        return ( addr.Accounts__r.size() > 0 || addr.Contacts1__r.size() > 0 );
```
MatthewBlanski, line 87:
```suggestion
        return ( addr.Accounts__r.isEmpty() == false || addr.Contacts1__r.isEmpty() == false ); 
```
@m-kinkade Please use .isEmpty() when checking whether or not a list has values.

src/classes/ADDR_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,452 @@
...
+                (SELECT Id FROM Address__c.Contacts1__r),
+                (SELECT Id FROM Address__c.Accounts__r)
+            FROM Address__c
+            WHERE Id IN :addrsWithChildRecords
+        ];
+
+        ADDR_CannotDelete_TDTM addrCannotDeleteClass = new ADDR_CannotDelete_TDTM();
+        
+        for (Address__c addr : returnAddresses) {
+            System.assert(addrCannotDeleteClass.hasChildRecords(addr) == TRUE);
```
MatthewBlanski, line 417: @m-kinkade No changes here. This communicated your intent clearly, and I wanted to call that out. If you believe that adding == TRUE or == FALSE may effectively communicate your intent in the assertion, you may want to consider the readability of System.assertEquals.

src/classes/ADDR_CannotDelete_TDTM.cls
m-kinkade: @MatthewBlanski would it make the code more readable to replace the ! with a check for FALSE instead?  
if (ADDR_CannotDelete_TDTM.enabledPreventAddressDeletion == FALSE) {

src/classes/ADDR_CannotDelete_TEST.cls
m-kinkade: Great suggestion.  I've updated my assertions to use System.assertEquals because I agree that it is more readable in most cases.

src/classes/ADDR_CannotDelete_TDTM.cls
MatthewBlanski: @m-kinkade I believe it would. When the check gets that long, it can be helpful. I wasn't going to list that as a required change.

src/classes/ADDR_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,451 @@
...
+    */
+    public static String hhAccRecordTypeID = UTIL_Describe.getHhAccRecTypeID(); 
+    public static String adminAccRecordTypeID = UTIL_Describe.getAdminAccRecTypeID();
+
+    /*********************************************************************************************************
+    * @description Test method to test that when Prevent_Address_Deletion__c is enabled in Hierarchy Settings, and
+    * Address has a child Account record associated to it, the Address record cannot be deleted.
+    */
+    @isTest 
+    private static void addrCannotDeleteWithAccount() {
```
MatthewBlanski, line 51: @m-kinkade My apologies for missing this the first time. Consider removing the 'addr' prefix for these methods as we have context from the class these are running in.

src/classes/ADDR_CannotDelete_TEST.cls
```diff
@@ -0,0 +1,451 @@
...
+            System.assertEquals(TRUE, addrCannotDeleteClass.hasChildRecords(addr));
+        }
+    }
+
+    /*********************************************************************************************************
+    * @description Test method for the hasChildRecords() method in ADDR_CannotDelete_TDTM to test that TRUE is
+    * returned when an Address record has both Contact and Account child records.
+    */
+    @isTest
+    private static void testAddrHasChildRecordsIsTrue() {
```
MatthewBlanski, line 393: @m-kinkade Another stylistic change that might be worth considering in the future is naming the test explicitly like hasChildRecordsIsTrue. With the context of the test class, @isTest, and the comments provided, we can probably ease up on some of the abbreviations and go with more human-language test methods.

spinkelman:
QA Passes - Merging
Will Close WI

Issue #1170, opened 2020-04-30
[EDA] Prevent Deletion of Test Record  if There are Associated Test Score Records


m-kinkade:
**Note: FOR INTERNAL USE ONLY. For support questions, please use the EDA Chatter group in [The Power of Us Hub](https://powerofus.force.com).
For feature requests, please use the Ideas tab in [The Power of Us Hub](https://powerofus.force.com).**

* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
When a user attempts to delete a Test record that is associated with child records (Test Score), the system deletes the Test record, leaving the child Test Score records intact with the Test field cleared.

* **What is the expected behavior?**
When a user attempts to delete a Test record that is associated with child records, the system will utilize a new custom setting field (Prevent Test Deletion) to determine whether to allow the deletion. If selected, attempts to delete a Test record with associated child records will be blocked and an error message displayed to the user. If unchecked, the functionality will remain the same as outlined in the current behavior section.

* **What is the motivation / use case for changing the behavior?**
From previous github requests and investigations of deletion of parent records with child records, it was determined that we need to implement the functionality to allow users the ability to control if a Test record with related child record can be deleted. 

* **Acceptance Criteria**

  - **Given**: The “Prevent Test Deletion” checkbox in Hierarchy Settings (Custom Setting) is unchecked.
  - **When**: A user clicks “Delete” button on a Test record that has child related records.
  - **Then**: The Test record is successfully deleted, no error message is displayed to the user and the child records are not deleted.

  - **Given**: The “Prevent Test Deletion” checkbox in Hierarchy Settings (Custom Setting) is checked.
  - **When**: A user clicks “Delete” button on a Test record that has child related records.
  - **Then**: An error message is displayed to the user and the user is unable to delete the Test record.

m-kinkade:
**lurch: attach W-040732

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001HuZsQAK" target="_blank">W-040732</a>

PR #1173, opened 2020-05-04
[EDA] Prevent Deletion of Test Record  if There are Associated Test Score Records

m-kinkade:
# Critical Changes

# Changes
We've added functionality to prevent the deletion of Test records with related Test Score records.

- If you have disabled the related TST_CannotDelete_TDTM classes for the object, the settings will remain disabled even though you've enabled them in Hierarchy Custom Settings.

# Issues Closed
Closes Issue #1170 

# New Metadata

- Prevent_Test_Deletion__c

Apex Classes:

- TST_CannotDelete_TDTM

- TST_CannotDelete_TEST

# Deleted Metadata

# Testing Notes
https://salesforce.quip.com/tu19APFVFuvF#OGIACAhkxdd

### DIFF
--- a/src/classes/TST_CannotDelete_TDTM.cls
+++ b/src/classes/TST_CannotDelete_TDTM.cls
@@ -1,0 +1,5 @@
+public with sharing class TST_CannotDelete_TDTM {
+    public TST_CannotDelete_TDTM() {
+
+    }
+}
--- a/src/classes/TST_CannotDelete_TDTM.cls
+++ b/src/classes/TST_CannotDelete_TDTM.cls
@@ -1,5 +1,91 @@
-public with sharing class TST_CannotDelete_TDTM {
-    public TST_CannotDelete_TDTM() {
+/*
+    Copyright (c) 2020, Salesforce.org
+    All rights reserved.
+    
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+    
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of Salesforce.org nor the names of
+      its contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+ 
+    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT 
+    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS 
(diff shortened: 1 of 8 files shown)


src/classes/TST_CannotDelete_TDTM.cls
```diff
@@ -0,0 +1,91 @@
...
+    TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
+        
+        if (TST_CannotDelete_TDTM.enabledPreventTestDeletion == FALSE) {
+            return new DmlWrapper(); 
+        }
+        
+        Map<Id, Test__c> oldmap = new Map<Id, Test__c>((List<Test__c>)oldList);
+        
+        if (triggerAction == TDTM_Runnable.Action.BeforeDelete) {
+            for (Test__c tst : [
```
xinnanlisf, line 62: The indentation looks a little weird here.

src/classes/TST_CannotDelete_TDTM.cls
```diff
@@ -0,0 +1,91 @@
...
+    }
+    
+    /*******************************************************************************************************
+     * @description Evaluates whether the Test has any child (Test Score) related records.
+     * @param tst is the current Test record.
+     * @return Boolean.
+     ********************************************************************************************************/
+    @testVisible
+    private Boolean hasChildRecords(Test__c tst) {
+
```
xinnanlisf, line 86: Same here. I don't think we need this spacing.

src/objects/Hierarchy_Settings__c.object
```diff
@@ -436,6 +436,15 @@
         <trackTrending>false</trackTrending>
         <type>Checkbox</type>
     </fields>
+    <fields>
+        <fullName>Prevent_Test_Deletion__c</fullName>
+        <defaultValue>false</defaultValue>
```
xinnanlisf, line 441: I thought we wanted it to be enabled by default?

m-kinkade:
> The PR looks good. Great work @m-kinkade! There are some minor indentation/spacing issues thou. Let's make sure we are consistent with spacing: https://github.com/SalesforceFoundation/EDA/pull/1165/files
> Let me know if any questions!
> 
> Also, I thought we wanted to enable this kind of setting by default now. Maybe that was me getting confused. I will also give a quick round of testing myself.

Thank you for reviewing @xinnanlisf...I'll make the changes now :)

src/objects/Hierarchy_Settings__c.object
m-kinkade: My understanding from the spec doc is it should be disabled for new and existing customers: 
"Ensure that Prevent_Test_Deletion__c is disabled for new and existing customers."

I'm happy to update if it was changed.  Is the spec doc the source of truth?

xinnanlisf: that makes sense.

spinkelman:
This passes QA as long as the branch conflict does not affect functional code of the feature. Once DEV confirms I will merge.

spinkelman:
Passes QA - Merging
Will Close WI

PR #1181, opened 2020-05-06
[EDA] Updating Name and Ownership In the Same Context Causes Duplicate Id Error in ACCT_IndividualAccounts_TDTM

### LINKED ISSUE #1129: [EDA] Updating Name and Ownership In the Same Context Causes Duplicate Id Error in ACCT_IndividualAccounts_TDTM
**Current Behavior:** 
When a Contact record's Name and Ownership is updated in the same time/context, a duplicate ID error pops up. 

**Expected Behavior:**
When a Contact record's Name and Ownership is updated in the same time/context, it should not throw an error. 

**Steps to Reproduce:** 
1. Created a Contact record 
2. Result: Account was created for the Contact record
3. Update the existing Contact's record Name & Ownership via data upload/execute anonymous/integration tools/etc. 
4. Result: The error below pops up below and the Contact record is not updated. 

"Line: 9, Column: 1
System.DmlException: Update failed. First exception on row 0 with id 0033i00000F1S6EAAV; first error: CANNOT_INSERT_UPDATE_ACTIVATE_ENTITY, hed.TDTM_Contact: execution of AfterUpdate caused by: System.ListException: Duplicate id in list: 0013i00000G7SQLAA3 Class.hed.TDTM_TriggerHandler.processDML: line 184, column 1 Class.hed.ACCT_IndividualAccounts_TDTM.handleUpdateProcessing: line 467, column 1 Class.hed.ACCT_IndividualAccounts_TDTM.handlesAfterInsertUpdate: line 277, column 1 Class.hed.ACCT_IndividualAccounts_TDTM.run: line 111, column 1 Class.hed.TDTM_TriggerHandler.runClass: line 145, column 1 Class.hed.TDTM_TriggerHandler.run: line 75, column 1 Class.hed.TDTM_Global_API.run: line 61, column 1 Trigger.hed.TDTM_Contact: line 33, column 1: []"

**Details and Steps to Reproduce**
**Current Behavior:** 
When a Contact record's Name and Ownership is updated in the same time/context, a duplicate ID error pops up. 

**Expected Behavior:**
When a Contact record's Name and Ownership is updated in the same time/context, it should not throw an error. 

**Steps to Reproduce:** 
1. Created a Contact record 
2. Result: Account was created for the Contact record
3. Update the existing Contact's record Name & Ownership via data upload/execute anonymous/integration tools/etc. 
4. Result: The error below pops up below and the Contact record is not updated. 

"Line: 9, Column: 1
System.DmlException: Update failed. First exception on row 0 with id 0033i00000F1S6EAAV; first error: CANNOT_INSERT_UPDATE_ACTIVATE_ENTITY, hed.TDTM_Contact: execution of AfterUpdate caused by: System.ListException: Duplicate id in list: 0013i00000G7SQLAA3 Class.hed.TDTM_TriggerHandler.processDML: line 184, column 1 Class.hed.ACCT_IndividualAccounts_TDTM.handleUpdateProcessing: line 467, column 1 Class.hed.ACCT_IndividualAccounts_TDTM.handlesAfterInsertUpdate: line 277, column 1 Class.hed.ACCT_IndividualAccounts_TDTM.run: line 111, column 1 Class.hed.TDTM_TriggerHandler.runClass: line 145, column 1 Class.hed.TDTM_TriggerHandler.run: line 75, column 1 Class.hed.TDTM_Global_API.run: line 61, column 1 Trigger.hed.TDTM_Contact: line 33, column 1: []"

skosgi1:
# Critical Changes

# Changes
We've fixed an error that occurs when a Contact's Last Name and Record Owner are updated in a single transaction.

# Issues Closed
Closes #1129 

# New Metadata
SRVC_UniqueRecord.cls
SRVC_UniqueRecord_TEST.cls

# Deleted Metadata

# Testing Notes
https://salesforce.quip.com/JjQmAbXNkSSQ#JMUACAQ3d5M


src/classes/SRVC_UniqueRecord_TEST.cls
```diff
@@ -0,0 +1,433 @@
...
+    @isTest
+    private static void combineRecordsTest() {
+
+        Contact testContact = new Contact();
+        testContact.Id = SRVC_UniqueRecord_TEST.generateFakeContactId();
+        testContact.firstName = 'Doe';
+
+        SObject sObjectConRecord = testContact.getSObjectType().newSobject(testContact.Id);
+
+        System.assert(((Contact)sObjectConRecord).firstName == NULL);
```
MatthewBlanski, line 377: @skosgi1 This is another potential stylistic change. Consider the readability of `System.assert(((Contact)sObjectConRecord).firstName == NULL);` against `System.assert( ((Contact)sObjectConRecord).firstName == NULL );`.

This may help sidestep a common short-circuit in the the way that we read that can allow us to gloss over repetitions of things like parenthesis or smaller words, such as 'the' earlier in this sentence.

src/classes/UTIL_Describe.cls
```diff
@@ -865,4 +866,33 @@ public class UTIL_Describe {
...
+    */
+    private static Map<SObjectType, List<String>> sObjectTypeToChildRelationshipNames = new Map<SObjectType, List<String>>();
+    
+
+    /*******************************************************************************************************
+    * @description Returns the list of all Child Relationship Name for a given Sobject Type
+    * @return sObjectTypeToChildRelationshipNames.get(sObjectTypeName) List of all the child Relationship Names
+    */
+    public static List<String> getAllChildRelationshipNames(SobjectType sObjectTypeName) {
+        if (sObjectTypeToChildRelationshipNames.get(sObjectTypeName) == NULL) {
```
MatthewBlanski, line 880: @skosgi1 Please use `UTIL_Describe.` before `sObjectTypeToChildRelationshipNames` to be explicit that we are interacting with a static variable.

src/classes/UTIL_Describe_API.cls
```diff
@@ -292,5 +292,13 @@ global class UTIL_Describe_API {
     global static String getStudentCharacteristicAttributeRecType() {
         return UTIL_Describe.getStudentCharacteristicAttributeRecType();
     }
+
+    /*******************************************************************************************************
+    * @description Returns the list of all Child Relationship Name for a given Sobject Type
+    * @return List of all the child Relationship Names
+    */
+    global static List<String> getAllChildRelationshipNames(SobjectType sObjectTypeName) {
```
MatthewBlanski, line 300: @skosgi1 Please remove this global method. Though it does follow our former pattern, that is a pattern we are slowly moving away from. Let's handle wait to handle interoperability concerns on this method until we need to rather than in anticipation of needing to. Our pattern may be more mature at that time.

src/classes/UTIL_Describe_TEST.cls
```diff
@@ -101,6 +101,35 @@ public with sharing class UTIL_Describe_TEST {
...
+        Account adminAccount = UTIL_UnitTestData_API.getMultipleTestAccounts(1, UTIL_Describe_API.getAdminAccRecTypeID())[0];
+        adminAccount.Name = 'Test Account 000';
+        insert adminAccount;
+
+        Test.startTest();
+            Schema.SObjectType sObjectTypeAcc = adminAccount.getSObjectType();
+            List<String> allChildRelationshipNames = UTIL_Describe_API.getAllChildRelationshipNames(sObjectTypeAcc);
+        Test.stopTest();
+
+        System.assert(allChildRelationshipNames != NULL);
```
MatthewBlanski, line 120: @skosgi1 Consider the readability of `System.assertNotEquals(NULL, allChildRelationshipNames)`. In many cases when we are testing for an expected or unexpected value, System.assertEquals and System.assertNotEquals may be good candidates for making the code readable.

src/classes/UTIL_Describe.cls
```diff
@@ -865,4 +866,33 @@ public class UTIL_Describe {
...
+                DescribeSObjectResult describe = sObjectTypeName.getDescribe();
+                for (Schema.ChildRelationship childRelationship : describe.getChildRelationships()) {
+                    if (childRelationship != NULL && childRelationship.getRelationshipName() != NULL) {
+                        childRelationshipNames.add(childRelationship.getRelationshipName());
+                    }
+                }
+                sObjectTypeToChildRelationshipNames.put(sObjectTypeName, childRelationshipNames);
+             } catch (Exception ex) {
+                 return NULL;
+               }
```
MatthewBlanski, line 893: @skosgi1 Please resolve the spacing issues on these lines.

Please have this return an empty string list. By forcing this to return null, we are creating behavior that we will need to account for further on in our code. If we can ensure that a method will never return null, then we can eliminate the handling of nulls for that method further on in our code.

If we want to have the behavior handled as though it is an error state, we should handle that as an exception so our code fails in a way that is informative to us and is meaningful to us when performing unit tests. We can catch the exception thrown and provide a custom exception with a custom label that is more informative to the end-user and debugger to the error that has occurred or we can preserve the underlying exception and test for that exception in our unit tests, failing the negative unit test if we encounter an unexpected exception.

I would prefer that we move away from any instance of try catch block that is generic enough to have the catch block be restated as catch{} as it can hide underlying unexpected behavior that can surface larger problems in the code.

skosgi1: Thanks @MatthewBlanski I updated the code based on our discussion. Can you please give it another shot? Thx.

src/classes/UTIL_Describe_TEST.cls
```diff
@@ -101,6 +101,37 @@ public with sharing class UTIL_Describe_TEST {
...
+    @isTest
+    private static void childRelationshipNamesServiceNegativeTest() {
+        try {
+            Test.startTest();
+                List<String> allChildRelationshipNames = UTIL_Describe.getAllChildRelationshipNames(NULL);
+            Test.stopTest();
+        } catch (Exception Ex) {
+            // Should throw a validation message, catching the message to the test the exception
+            System.assert(ex.getMessage().contains('Attempt to de-reference a null object'));
+        }
```
MatthewBlanski, line 135: @skosgi1 Please change the structure of this to handle the exception and test for not throwing an exception or encountering an unexpected exception. We can accomplish this by placing a System.assert(false,message) at the end of the try block and have an empty catch block for System.NullPointerException.

src/classes/SRVC_UniqueRecord.cls
```diff
@@ -0,0 +1,159 @@
...
+        for (SObject sObjectToCheck: allSObjectRecords) {
+            if (sObjectToCheck.Id == NULL) {
+                recordsWithoutIds.add(sObjectToCheck);
+            }
+            else {
+                recordsWithIds.add(sObjectToCheck);
+            }
+        }
+
+        returnMap.put('withoutIds', recordsWithoutIds);
```
xinnanlisf, line 155: On the picky side, I noticed this during our demo yesterday: shall we make these keys Constants? Also, maybe we can capitalize them?

skosgi1: Making constants will require additional variable declaration, weren't we trying to avoid such scenarios?

MatthewBlanski: If we are going that route, I am more comfortable making an inner Enum class and using that.

skosgi1: Will update the code accordingly.

skosgi1: @xinnanlisf @MatthewBlanski thanks for the suggestions. I have updated the code accordingly can you please give it another review? TIA.

spinkelman:
Passes QA per @dgamboni instruction - Merging
Will close WI

Issue #1174, opened 2020-05-04
[EDA] Prevent Deletion of Time Block Record  if There are Associated Course Records


m-kinkade:
**Note: FOR INTERNAL USE ONLY. For support questions, please use the EDA Chatter group in [The Power of Us Hub](https://powerofus.force.com).
For feature requests, please use the Ideas tab in [The Power of Us Hub](https://powerofus.force.com).**

* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
When a user attempts to delete a Time Block record that is associated with child records (Course Offering or Course Offering Schedule), the system deletes the Time Block record, leaving any child records intact with the Time Block field cleared.

* **What is the expected behavior?**
When a user attempts to delete a Time Block record that is associated with child records, the system will utilize a new custom setting field (Prevent Time Block Deletion) to determine whether to allow the deletion. If selected, attempts to delete a Time Block record with associated child records will be blocked and an error message displayed to the user.  If unchecked, the functionality will remain the same as outlined in the current behavior section.

* **What is the motivation / use case for changing the behavior?**
From previous github requests and investigations of deletion of parent records with child records, it was determined that we need to implement the functionality to allow users the ability to control if a Time Block record with related child record can be deleted. 

* **Acceptance Criteria**

  - **Given**: The “Prevent Time Block Deletion” checkbox in Hierarchy Settings (Custom Setting) is unchecked.
  - **When**: A user clicks “Delete” button on a Time Block record that has child related records.
  - **Then**: The Time Block record is successfully deleted, no error message is displayed to the user and the child records are not deleted.

  - **Given**: The “Prevent Time Block Deletion” checkbox in Hierarchy Settings (Custom Setting) is checked.
  - **When**: A user clicks “Delete” button on a Time Block record that has child related records.
  - **Then**: An error message is displayed to the user and the user is unable to delete the Time Block record.

m-kinkade:
**Lurch: attach W-040735

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x1E000001Hua7QAC" target="_blank">W-040735</a>

PR #1182, opened 2020-05-06
[EDA] Prevent Deletion of Time Block Record  if There are Associated Course Records

m-kinkade:
# Critical Changes

# Changes
We've added functionality to prevent the deletion of Time Block records with related Course Offering or Course Offering Schedule records.

- If you have disabled the related TB_CannotDelete_TDTM classes for the object, the settings will remain disabled even though you've enabled them in Hierarchy Custom Settings.

# Issues Closed
Closes Issue #1174 

# New Metadata

- Prevent_Time_Block_Deletion__c

Apex Classes:

- TB_CannotDelete_TDTM

- TB_CannotDelete_TEST

# Deleted Metadata

# Testing Notes
https://salesforce.quip.com/tocpA7KiYHWk#HNYACA6cKr0

### DIFF
--- a/src/classes/TB_CannotDelete_TDTM.cls
+++ b/src/classes/TB_CannotDelete_TDTM.cls
@@ -30,10 +30,10 @@
 /**
 * @author Salesforce.org
 * @date 2020
-* @group Time Block
+* @group Time Blocks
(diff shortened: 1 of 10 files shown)


spinkelman:
Passes QA - Merging
Will close WI

PR #1208, opened 2020-05-21
Add Standard Triggers

ZergyPoo:
# Critical Changes

# Changes
## New Trigger Handlers
We've added these Trigger Handlers, which you can access for more information:
- [TDTM_Event](https://github.com/SalesforceFoundation/EDA/blob/master/src/triggers/TDTM_Event.trigger)
- [TDTM_Lead](https://github.com/SalesforceFoundation/EDA/blob/master/src/triggers/TDTM_Lead.trigger)
- [TDTM_Task](https://github.com/SalesforceFoundation/EDA/blob/master/src/triggers/TDTM_Task.trigger)
- [TDTM_User](https://github.com/SalesforceFoundation/EDA/blob/master/src/triggers/TDTM_User.trigger)

They don't contain logic and don't appear in the Trigger Handlers tab. They trigger the product-specific TDTM handlers that are marked Active in the product where the logic is actually executed. EDA itself doesn't have any product-specific TDTM handlers for these objects. If you already have a trigger that calls the TDTM handler and TDTM classes for any of these objects and you're not preventing re-entrancy for the objects, remove your trigger to avoid calling the classes twice.

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


juliesnow:
@MatthewBlanski, @ZergyPoo - Please let me know if this Description text is okay. Ran initial draft w/ Bobby, who's approved.

spinkelman:
QA Passes - Merging
Closing WI

PR #1226, opened 2020-06-18
[Robot] Course connections settings page

Safiyab:
Checking warning messages when the 'Enable Course Connections' field is unchecked.
Verifying drop down values for two fields

Robot WI: [W-041783](https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000007hawEQAQ/view)

cci task run robot -o suites robot/EDA/tests/browser/settings/eda_settings_course_connections_edit_mode.robot

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


robot/EDA/resources/CourseConnectionsSettingsPageObject.py
```diff
@@ -20,3 +21,50 @@ def _is_current_page(self):
...
         )
+
+    def set_enable_course_connections(self):
+        """ Set the checkbox for 'Enable Course Connections' field """
+        locator = eda_lex_locators["eda_settings_cc"]["enable_cc_checkbox"]
+        self.selenium.wait_until_page_contains_element(locator)
+        self.selenium.click_element(locator)
+
+    def update_enable_cc_to_default(self):
+        """ Updating the `Enable Course Connections` checkbox to default value (false)"""
```
skristem, line 32: looks like first u r checking and updating only if not correct. Maybe u can adjust ur documentation to reflect that so that its clear if anyone else wants to use

robot/EDA/resources/EDA.py
```diff
@@ -331,3 +332,12 @@ def click_edit_on_eda_settings_page(self):
         locator = eda_lex_locators["eda_settings"]["edit"]
         self.selenium.wait_until_page_contains_element(locator, error="Edit button is not available on the page")
         self.selenium.click_element(locator)
+
+    def click_action_button_on_eda_settings_page(self, action):
+        """ Clicks on the action (eg: Save, Cancel) button on the EDA Settings page """
+        locator = eda_lex_locators["eda_settings"]["action"].format(lower(action))
```
rjanjanam, line 338: are u converting action to lowercase ?

robot/EDA/resources/EDA.py
```diff
@@ -331,3 +332,12 @@ def click_edit_on_eda_settings_page(self):
...
+        self.selenium.wait_until_page_contains_element(
+            locator, error=f"Action button with locator '{locator}' is not available on the EDA settings page")
+        self.selenium.click_element(locator)
+        if action == "Save":
```
rjanjanam, line 342: while having comparisons it always a good idea to compare In lowercase 
action.lower() == "save"

robot/EDA/tests/browser/settings/eda_settings_course_connections_edit_mode.robot
```diff
@@ -7,22 +7,30 @@ Library         cumulusci.robotframework.PageObjects
...
+    Set enable course connections
+    Verify enable course connections warning    false
+    Verify dropdown values                      Default Active Student Record Type
+    ...                                         Default
+    ...                                         Faculty
+    ...                                         Student
+    Verify dropdown values                      Default Faculty Record Type
+    ...                                         Default
+    ...                                         Faculty
+    ...                                         Student
```
rjanjanam, line 28: is it possible to modify or enhance the functionality of verify dropdown values to read  multiple lists ? Then you don't need to call the same keyword twice ?

robot/EDA/resources/locators.py
```diff
@@ -95,6 +96,14 @@
         "settings_tab": "(//li[@class='slds-tabs__item slds-text-heading--label slds-active' and @role='tab' and @title='Settings'])[1]/a[contains(text(),'Settings')]",
         "affl_mappings_tab": "//a[contains(text(),'Affiliation Mappings')]",
     },
+    "eda_settings_cc": {
+        "default_cc_checkbox": "//div[text()='Enable Course Connections']/following-sibling::div/descendant::img",
+        "dropdown_values": "//div[text()='{}']/following-sibling::div/select/option[text()='{}']",
+        "dropdown_values_count": "//div[text()='{}']/following-sibling::div/select/option",
+        "enable_cc_checkbox": "//div[text()='Enable Course Connections']/following-sibling::div[1]/descendant::span",
```
skristem, line 103: can u use something else instead of indexing? also this locator looks almost similar to line 100 can this be parameterized and reused something like //div[text()='{}']/following-sibling::div[some attribute]/descendant::{}

robot/EDA/resources/CourseConnectionsSettingsPageObject.py
```diff
@@ -20,3 +21,50 @@ def _is_current_page(self):
             locator_tab,
             error=f"Course Connections tab with locator '{locator_tab}' is not available on the page"
         )
+
+    def set_enable_course_connections(self):
```
skristem, line 25: what will happen if checkbox is already checked? may be first check if its checked already and if its not checked and check it

robot/EDA/tests/browser/settings/eda_settings_course_connections_edit_mode.robot
```diff
@@ -7,22 +7,30 @@ Library         cumulusci.robotframework.PageObjects
 Suite Setup     Open Test Browser
 Suite Teardown  Capture screenshot and delete records and close browser
 
+Test Setup      Run keywords
+...             Go to EDA settings tab          Course Connections      AND
+...             Update enable cc to default
+
 *** Test Cases ***
 Validate Edit Mode For Course Connections, Settings
```
skristem, line 15: please add documentation to the test case

robot/EDA/resources/EDA.py
Safiyab: Yes I had to. Coz the label on the UI says "Save" or "Cancel", whereas the locator is "save" or "cancel".
While we code, we generally use what is visible on the UI, so that's why I am converting to lower

robot/EDA/tests/browser/settings/eda_settings_course_connections_edit_mode.robot
Safiyab: What do you mean by that? Like having sub-dictionaries?

rjanjanam: yes, it is not a must but something to consider in the future

robot/EDA/resources/CourseConnectionsSettingsPageObject.py
Safiyab: Updated

robot/EDA/resources/locators.py
Safiyab: Index wasn't really required there anyway. Removed it

robot/EDA/resources/CourseConnectionsSettingsPageObject.py
Safiyab: That's why as part of the test setup, we are making sure the checkbox is defaulted to false. 

The reason why I am deferring this is we can't check the status of the checkbox when we are in edit mode. To catch this, I will have to go out of edit mode, check the status, click edit and toggle accordingly. Going out of edit mode is complicated, coz do I Save or Cancel ? If I Save, will it cause issues because I might not have completed all mandatory fields in my .robot file etc.

So a solution to this is: I have requested a change in the DOM and it is currently under SPEC review. I anticipate it to be available to us in the next week or so. And then I will update the code to check the status and THEN toggle. Sound okay to you ?

robot/EDA/tests/browser/settings/eda_settings_course_connections_edit_mode.robot
Safiyab: Done

PR #1228, opened 2020-06-19
Added escape single quotes to a dynamic soql on  tdtm_runnable

### LINKED ISSUE #1243: Add escape single quotes to a dynamic soql on tdtm_runnable

feature
Currently TDTM_Filter.cls and TDTM_Runnable.cls class execute dynamic SOQL query on a String that doesn't escape single quotes. 

Add escapeSingleQuotes() to the dynamic queries to protect against any potential SOQL injections.

This makes our classes more secure to any possible security threats.

skosgi1:
# Critical Changes

# Changes
We've added the escapeSingleQuotes method to the TDTM_Filter and TDTM_Runnable classes to protect against potential SOQL injections. You shouldn't notice any change in behavior as a result of this change.

# Issues Closed
Closes #1243 

# New Metadata

# Deleted Metadata

# Testing Notes
Testing may be accomplished by adding a filter to a Trigger Handler record. Any filter functionality still working would constitute a valid test.


xinnanlisf:
@skosgi1 I am wondering if we should do as @MatthewBlanski did by escaping all string variables.

MatthewBlanski:
> @skosgi1 I am wondering if we should do as @MatthewBlanski did by escaping all string variables.

We don't need to escape anything referenced with the ':variable' syntax as those are handled natively. Anything dynamic likely should be escaped.

xinnanlisf:
Hi @skosgi1, instead of escaping single quote for the whole query, lets escape only objectName now :)

skosgi1:
> Hi @skosgi1, instead of escaping single quote for the whole query, lets escape only objectName now :)

updated it @xinnanlisf Thx.

xinnanlisf:
although, @skosgi1 do we have a github issue linked to this? Also, do we need release note?

skosgi1:
> although, @skosgi1 do we have a github issue linked to this? Also, do we need release note?

Hmm I don't think we do, since this is internal.

skosgi1:
@juliesnow I have not added any release notes earlier thinking it is not needed. Looks like we want that. Can you please take a look when you get a chance, thank you.

PR #1242, opened 2020-07-06
[Robot] Validate the dropdown fields are disabled by default in edit mode 

rhakeshj:
Validate the drop down fields (Default Active Student Record Type, Default Faculty Record Type) are disabled when the course connections, settings tab is in edit mode.

Robot WI: W-041782

cci task run robot -o suites robot/EDA/tests/browser/settings/course_connections.robot

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/robot/EDA/resources/EDA.py
+++ b/robot/EDA/resources/EDA.py
@@ -402,6 +402,7 @@ class EDA(BaseEDAPage):
         """
         for field,expected_value in kwargs.items():
             locator = eda_lex_locators["eda_settings_cc"]["dropdown_field"].format(field)
+            self.selenium.page_should_contain_element(locator)
             self.selenium.wait_until_element_is_visible(locator,
                                                 error= "Element is not displayed for the user")
             actual_value = self.selenium.get_webelement(locator).get_attribute("disabled")
--- a/robot/EDA/resources/EDA.py
+++ b/robot/EDA/resources/EDA.py
@@ -408,4 +408,4 @@ class EDA(BaseEDAPage):
             actual_value = self.selenium.get_webelement(locator).get_attribute(expected_value)
             expected_value = bool(expected_value == "disabled")
             if not str(expected_value).lower() == str(actual_value).lower() :
-                raise Exception (f"Drop down field '{field}' status is '{actual_value}'")
+                raise Exception (f"Drop down field {field} status is {actual_value} instead of {expected_value}")
--- a/robot/EDA/resources/EDA.py
+++ b/robot/EDA/resources/EDA.py
@@ -397,14 +397,15 @@ class EDA(BaseEDAPage):
 
     def verify_dropdown_field_status(self, **kwargs):
         """ Verify the drop down field is disabled/enabled for the user
-            we have to pass the name of the field and the status of the field 
-            True = disabled and False = enabled
+            we have to pass the name of the field and the expected status
+            of the field as either enabled or disabled
         """
         for field,expected_value in kwargs.items():
             locator = eda_lex_locators["eda_settings"]["dropdown_field"].format(field)
             self.selenium.page_should_contain_element(locator)
             self.selenium.wait_until_element_is_visible(locator,
                                                 error= f"Element '{field}' is not displayed for the user")
-            actual_value = self.selenium.get_webelement(locator).get_attribute("disabled")
+            actual_value = self.selenium.get_webelement(locator).get_attribute(expected_value)
+            expected_value = bool(expected_value == "disabled")
             if not str(expected_value).lower() == str(actual_value).lower() :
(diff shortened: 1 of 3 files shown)


robot/EDA/resources/EDA.py
```diff
@@ -394,3 +394,17 @@ def verify_selected_dropdown_value(self,**kwargs):
             actual_value = self.selenium.get_webelement(locator).text
             if not str(value).lower() == str(actual_value).lower() :
                 raise Exception (f"Drop down value in '{field}' is not updated and the value is '{actual_value}'")
+
+    def verify_dropdown_field_status(self, **kwargs):
+        """ Verify the drop down field is disabled/enabled for the user
+            we have to pass the name of the field and the status of the field 
+            True = disabled and False = enabled
```
Safiyab, line 401: Per normal convention, should it be: true = enabled and false = disabled?

robot/EDA/resources/EDA.py
```diff
@@ -394,3 +394,17 @@ def verify_selected_dropdown_value(self,**kwargs):
...
+        """
+        for field,expected_value in kwargs.items():
+            locator = eda_lex_locators["eda_settings_cc"]["dropdown_field"].format(field)
+            self.selenium.page_should_contain_element(locator)
+            self.selenium.wait_until_element_is_visible(locator,
+                                                error= "Element is not displayed for the user")
```
Safiyab, line 407: Can you mention the field name in the error here ?
Since you are using dicts, if this keyword fails, having the {field} name in the error message will help with debugging

robot/EDA/resources/locators_49.py
```diff
@@ -102,6 +102,7 @@
     },
      "eda_settings_cc": {
         "default_cc_checkbox": "//div[text()='Enable Course Connections']/following-sibling::div/descendant::img",
+        "dropdown_field": "//div[text()='{}']/following-sibling::div/select",
```
Safiyab, line 105: Since your keyword is in EDA.py, it is better to move this locator to "eda_settings" dict rather than this one which deals specifically with Course Connections page.

robot/EDA/resources/EDA.py
rhakeshj: Yeah! But in this case we are retrieving the value of attribute "default" so the other way. By default when the drop down field is disabled , this attribute "disabled" returns the value as True and vice versa

robot/EDA/resources/EDA.py
rhakeshj: Updated error message to display field

robot/EDA/resources/locators_49.py
rhakeshj: Moved the xpath to eda_settings dictionary

robot/EDA/resources/EDA.py
Safiyab: Yeah but think of it from code perspective and not the functionality you are automating. What if you have to reuse this keyword for another page where default is enabled ?
I'm just trying to explain my thoughts here. You can leave it as-is, if this makes more sense to you.

skristem: why dont u just ask to pass the status as either enabled or disabled and u convert to either true or false? that way, there is no confusion

robot/EDA/resources/EDA.py
```diff
@@ -394,3 +394,17 @@ def verify_selected_dropdown_value(self,**kwargs):
...
+            True = disabled and False = enabled
+        """
+        for field,expected_value in kwargs.items():
+            locator = eda_lex_locators["eda_settings"]["dropdown_field"].format(field)
+            self.selenium.page_should_contain_element(locator)
+            self.selenium.wait_until_element_is_visible(locator,
+                                                error= f"Element '{field}' is not displayed for the user")
+            actual_value = self.selenium.get_webelement(locator).get_attribute("disabled")
+            if not str(expected_value).lower() == str(actual_value).lower() :
+                raise Exception (f"Drop down field '{field}' status is '{actual_value}'")
```
skristem, line 410: just a suggestion, may be little more detailed like Expected '{field}' Dropdown status to be {expected_value} bout found '{actual_value}

skristem: also u dont need quotes around {} again in f string

robot/EDA/resources/EDA.py
rhakeshj: @Safiyab @skristem : I haven't pushed the code yet. Please take a look at the below and let me know if this is ok to be a generic keyword

image

robot/EDA/resources/EDA.py
rhakeshj: Updated both

PR #1261, opened 2020-07-28
[EDA] Refactor: Contact Deceased/Do Not Contact

m-kinkade:
# Critical Changes

# Changes
## Performance Improvement to CON_DoNotContact_TDTM
We've refactored the CON_DoNotContact_TDTM class to improve readability, modularity, and performance.  The functionality of this class hasn't changed.

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
https://salesforce.quip.com/pkQxAEh3g8qK#QIFACA6cvDz


skosgi1:
This looks really good! Would it make sense to move the synchronization code to its own class? (it was not a part of original discussion/SPEC) @m-kinkade @MatthewBlanski @xinnanlisf

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -29,98 +29,185 @@
 */
 /**
 * @author Salesforce.org
-* @date 2015
+* @date 2020
```
skosgi1, line 32: Not sure if this should be changed as original created date is 2015?

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -29,98 +29,185 @@
...
+    * @description Synchronizes Household naming exclusions and communication preferences 
+    * with changes to the Deceased and Do Not Contact fields for the Contact object.
+    * @param newList the list of Contacts from trigger new
+    * @param oldList the list of Contacts from trigger old
+    * @param triggerAction which trigger event (BeforeInsert, AfterInsert, etc.)
+    * @param objResult the describe for Contacts
     * @return dmlWrapper.
     ********************************************************************************************************/
-    public override DmlWrapper run(List<SObject> newlist, List<SObject> oldlist,
+    @TestVisible
```
skosgi1, line 49: @TestVisible is not required as it is a public method.

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -29,98 +29,185 @@
...
+    public override DmlWrapper run(List<SObject> newList, List<SObject> oldList,
         TDTM_Runnable.Action triggerAction, Schema.DescribeSObjectResult objResult) {
 
-        for (integer i=0; i<newlist.size();i++) {
+        List<Contact> newContacts = (List<Contact>)newList;
```
skosgi1, line 53: as discussed in chat, something like newContactList.

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -29,98 +29,184 @@
...
-    * @param con The Contact
-    * @return void
-    ********************************************************************************************************/
-    public static void copyNamingExclusionsFromCheckboxes(Contact con) {
-        string mspValue = '';
+    /**
+     * @description Synchronizes the communication preference fields
+     * (Do not Call, Email Opt Out, Fax Opt Out) on a contact record
+     * to match the status specified by the "Do Not Contact" field.
+     * @param  contactsList List of contacts to synchronize
```
MatthewBlanski, line 157:
```suggestion
     * @param contactsList List of contacts to synchronize
```
Spacing

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -29,98 +29,184 @@
...
-            mspValue += 'Household__c.Name;';
-        if (con.Exclude_from_Household_Formal_Greeting__c)
-            mspValue += 'Household__c.Formal_Greeting__c;';
-        if (con.Exclude_from_Household_Informal_Greeting__c)
-            mspValue += 'Household__c.Informal_Greeting__c;';
+    /**
+     * @description Synchronizes the Household Naming Exclusion fields 
+     * (checkboxes and picklist) on a contact record to match the status 
+     * specified by the "Deceased" field.
+     * @param  contactsList List of contacts to synchronize
```
MatthewBlanski, line 172:
```suggestion
     * @param contactsList List of contacts to synchronize
```

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -29,98 +29,184 @@
...
-            mspValue = mspValue.substring(0,mspValue.length()-1);
+        this.syncHHNamingExclusionsPicklistFromCheckboxes(contactsList);
+    }
 
-        con.Naming_Exclusions__c = mspValue;
+    /**
+     * @description Synchronizes the naming exclusions specified by the selected 
+     * checkboxes with the selected values listed in the Naming Exclusion 
+     * multiselect picklist.
+     * @param  contactsList List of contacts to synchronize
```
MatthewBlanski, line 189:
```suggestion
     * @param contactsList List of contacts to synchronize
```

src/classes/UTIL_UnitTestData_TEST.cls
```diff
@@ -95,8 +95,14 @@ public class UTIL_UnitTestData_TEST {
     public static List<Contact> getMultipleTestContacts(Integer n) {
         List<contact> ContactsToAdd = new List<contact>();
         for(Integer i=0; i<n; i++) {
-            Contact newCon = new Contact(FirstName = i + 'Test' + i, LastName = i + 'Contact_forTests' + i, WorkEmail__c = 'fakeemail' + i + '@salesforce.org',
-                             Preferred_Email__c = 'Work Email', WorkPhone__c = i + '06-777-888' + i, PreferredPhone__c = 'Work Phone', OtherCity = 'Seattle');
+            Contact newCon = new Contact(FirstName = i + 'Test' + i,
```
MatthewBlanski, line 98: This is significantly more readable. Thank you.

src/classes/CON_DoNotContact_TEST.cls
```diff
@@ -29,255 +29,1346 @@
...
+        
+        for (Contact con : testContactsList){
+            con.Deceased__c = true;
+        }
+
+        Test.startTest();
+        insert testContactsList;
+        Test.stopTest();
+
+        List<Contact> contactsAfterInsertList = [SELECT Do_Not_Contact__c,
```
MatthewBlanski, line 62: In the future we may want to consider using a format that allows us to see the query in context with a single indentation similar to `[
    SELECT Do_Not_Contact__c
    ....
    ....
    FROM CONTACT
    WHERE Id IN :testContactsList];
]`

We're using longer variable names now, which means our SOQL query indentation is pretty deep in the format displayed in this PR. Just a thought for future discussion.

src/classes/CON_DoNotContact_TEST.cls
```diff
@@ -29,255 +29,1346 @@
...
+    }
 
-        Contact contact = UTIL_UnitTestData_TEST.getContact();
-        contact.do_not_contact__c = true;
+    /**************************************************************************************************************************
+    * @description Test method to verify that syncNamingExclusionsOnContact removes all household naming exclusions   
+    * when a null value is passed for Deceased.
+    */
+    @isTest 
+    private static void syncNamingExclusionsNull(){
```
MatthewBlanski, line 529: This is a good example of a test scenario for the odd behavior with Salesforce checkboxes. Good catch and coverage here. I admittedly would not have caught its absence in this code review.

skosgi1:
Nice work! Spacing looks good. Approving.

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -32,95 +32,177 @@
...
+            Map<Id, Contact> oldUpdatedContactsById = new Map<Id, Contact>((List<Contact>)oldlist);
+            this.processBeforeUpdate(newContactsList, oldUpdatedContactsById);
         }
 
         return new DmlWrapper();
-	}
+    }
+    
+    /**
+     * @description Determine if trigger fired is the result of a BeforeInsert action.
```
xinnanlisf, line 67: Sorry about being picky but it looks like other description has different spacing for * and /. Shall we make them consistent. Maybe we can align the / with the first stars.

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -32,95 +32,177 @@
...
+    private void syncNamingExclusionsOnContacts(List<Contact> contactsList){
+        for (Contact con : contactsList){
+            con.Exclude_from_Household_Name__c = con.Deceased__c;
+            con.Exclude_from_Household_Formal_Greeting__c = con.Deceased__c;
+            con.Exclude_from_Household_Informal_Greeting__c = con.Deceased__c;
+        }
 
-        if (mspValue.endsWith(';'))
-            mspValue = mspValue.substring(0,mspValue.length()-1);
+        this.syncHHNamingExclusionsPicklistFromCheckboxes(contactsList);
```
xinnanlisf, line 178: Looks like we have a chain of methods here. Shall we try to de-couple them since you can just call syncHHNamingExclusionsPicklistFromCheckboxes in 'process' methods?

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -32,95 +32,177 @@
...
 
-        con.Naming_Exclusions__c = mspValue;
+    /**
+     * @description Synchronizes the naming exclusions specified by the selected 
+     * checkboxes with the selected values listed in the Naming Exclusion 
+     * multiselect picklist.
+     * @param contactsList List of contacts to synchronize
+     */ 
+    @TestVisible
+    private void syncHHNamingExclusionsPicklistFromCheckboxes(List<Contact> contactsList) {
```
xinnanlisf, line 188: Naming is hard. I am wondering if we can avoid using generic phrases like 'FromCheckboxes'. People might ask what checkboxes and get confused.

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -32,95 +32,177 @@
...
-        con.Exclude_from_Household_Informal_Greeting__c = con.Deceased__c;
+            if (con.Do_Not_Contact__c == true){
+                doNotContactContactsList.add(con);
+            }
+        }
 
-        copyNamingExclusionsFromCheckboxes(con);
+        if (deceasedContactsList.isEmpty() == false){
+            this.syncNamingExclusionsOnContacts(deceasedContactsList);
+            this.syncDoNotContactFieldsOnContacts(deceasedContactsList);
```
kevinswiggum, line 108: @m-kinkade is this line necessary? Do Not Contact field is getting set on line 97, meaning this contact will also get added to the doNotContactContactsList. We're running the same contact through the syncDoNotContactFieldsOnContacts method twice.

src/classes/CON_DoNotContact_TDTM.cls
kevinswiggum, line 188: Was a decision made about getting rid of this? Any reason to keep it?

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -32,95 +32,177 @@
...
-        con.DoNotCall = con.Do_Not_Contact__c;
-        con.HasOptedOutOfFax  = con.Do_Not_Contact__c;
+            if (con.Do_Not_Contact__c != oldCon.Do_Not_Contact__c){
+                doNotContactContactsList.add(con);
+            }
+        }
 
+        if (deceasedContactsList.isEmpty() == false){
+            this.syncNamingExclusionsOnContacts(deceasedContactsList);
+            this.syncDoNotContactFieldsOnContacts(deceasedContactsList);
```
kevinswiggum, line 141: Same question as line 108

src/classes/CON_DoNotContact_TDTM.cls
MatthewBlanski: We are removing it in another PR post-refactor.

src/classes/CON_DoNotContact_TDTM.cls
```diff
@@ -32,95 +32,177 @@
...
+    private void processBeforeUpdate(List<Contact> newContactsList, Map<Id, Contact> oldContactsById) {
+        List<Contact> deceasedContactsList = new List<Contact>();
+        List<Contact> doNotContactContactsList = new List<Contact>();
+        
+        for (Contact con : newContactsList){
+            Contact oldCon = oldContactsById.get(con.Id);
+
+            if (con.Deceased__c != oldCon.Deceased__c){
+                con.Do_Not_Contact__c = con.Deceased__c;
+                deceasedContactsList.add(con);
```
kevinswiggum, line 131: @m-kinkade @MatthewBlanski  This variable name isn't accurate based on what it's doing. It's not necessarily that they've been marked deceased. It's that the deceased field has been changed. As it works now (and as it did before the refactor), unchecking the deceased field would also uncheck all the other fields. Same with the Do Not Contact field. I'm not suggesting we change the functionality as part of this PR but variable names could be more accurate (e.g changedDoNotConactList)

src/classes/CON_DoNotContact_TDTM.cls
m-kinkade: Great catch, I've updated and rerun the tests to ensure nothing broke in the process.

PR #1265, opened 2020-07-30
Updates to AFFL_MultiRecordType_TEST to prevent build failures

skosgi1:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
Updated AFFL_MultiRecordType_TEST to prevent build failures. Unit tests running and a successful build will likely be a sufficient test. No customer impact as this hasn't hit them yet.

### DIFF
--- a/src/classes/AFFL_MultiRecordType_TEST.cls
+++ b/src/classes/AFFL_MultiRecordType_TEST.cls
@@ -676,15 +676,12 @@ public with sharing class AFFL_MultiRecordType_TEST {
     private static void autoCreateProgramEnrollment() {		
         List<AffiliationMappingWrapper> affiliationMappingWrapperList =
             AFFL_MultiRecordType_TEST.setupMappingsByNameAndLabel();
-        String bizOrgRecordTypeName = affiliationMappingWrapperList[1].getAccountRecordTypeName();
 
-        //Turn on Academic Record auto-creation for Business Organization mapping - doesn't make a lot of sense for this type, but just for testing 
-        Affl_Mappings__c bizOrgMapping = [SELECT Auto_Program_Enrollment__c,
-            Auto_Program_Enrollment_Role__c,
-            Auto_Program_Enrollment_Status__c
-            FROM Affl_Mappings__c
-            WHERE Name = :bizOrgRecordTypeName];
+        AffiliationMappingWrapper bizOrgMappingWrapper = affiliationMappingWrapperList[1];
+        String bizOrgRecordTypeName = bizOrgMappingWrapper.getAccountRecordTypeName();
 
+        //Turn on Academic Record auto-creation for Business Organization mapping, which we are using as an arbitrary record type for testing.
+        Affl_Mappings__c bizOrgMapping = bizOrgMappingWrapper.getAffiliationMapping();
         bizOrgMapping.Auto_Program_Enrollment__c = true;
         bizOrgMapping.Auto_Program_Enrollment_Role__c = 'Student';
         bizOrgMapping.Auto_Program_Enrollment_Status__c = 'Current';
@@ -714,15 +711,12 @@ public with sharing class AFFL_MultiRecordType_TEST {
     private static void autoCreateProgramEnrollmentDifferentRole() {
         List<AffiliationMappingWrapper> affiliationMappingWrapperList =
             AFFL_MultiRecordType_TEST.setupMappingsByNameAndLabel();
-        String bizOrgRecordTypeName = affiliationMappingWrapperList[1].getAccountRecordTypeName();
 
-        //Turn on Academic Record auto-creation for Business Organization mapping - doesn't make a lot of sense for this type, but just for testing
-        Affl_Mappings__c bizOrgMapping = [SELECT Auto_Program_Enrollment__c,
-            Auto_Program_Enrollment_Role__c,
-            Auto_Program_Enrollment_Status__c
-            FROM Affl_Mappings__c
-            WHERE Name = :bizOrgRecordTypeName];
+        AffiliationMappingWrapper bizOrgMappingWrapper = affiliationMappingWrapperList[1];
+        String bizOrgRecordTypeName = bizOrgMappingWrapper.getAccountRecordTypeName();
 
+        //Turn on Academic Record auto-creation for Business Organization mapping, which we are using as an arbitrary record type for testing.
+        Affl_Mappings__c bizOrgMapping = bizOrgMappingWrapper.getAffiliationMapping();
         bizOrgMapping.Auto_Program_Enrollment__c = true;
         bizOrgMapping.Auto_Program_Enrollment_Role__c = 'Student';
         bizOrgMapping.Auto_Program_Enrollment_Status__c = 'Current';
--- a/src/classes/AFFL_MultiRecordType_TEST.cls
+++ b/src/classes/AFFL_MultiRecordType_TEST.cls
@@ -1000,12 +1000,8 @@ public with sharing class AFFL_MultiRecordType_TEST {
         String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
 
         List<Affl_Mappings__c> mappings = new List<Affl_Mappings__c>();
-            mappings.add(new Affl_Mappings__c(Name = 'Academic Program', Account_Record_Type__c = 'Academic_Program', Primary_Affl_Field__c = prefix + 'Primary_Academic_Program__c', Auto_Program_Enrollment__c = true, Auto_Program_Enrollment_Status__c = 'Current', Auto_Program_Enrollment_Role__c = 'Student'));
             mappings.add(new Affl_Mappings__c(Name = 'Business Organization', Account_Record_Type__c = 'Business_Organization', Primary_Affl_Field__c = prefix + 'Primary_Organization__c'));
             mappings.add(new Affl_Mappings__c(Name = 'Household Account', Account_Record_Type__c = 'HH_Account', Primary_Affl_Field__c = prefix + 'Primary_Household__c'));
-            mappings.add(new Affl_Mappings__c(Name = 'Educational Institution', Account_Record_Type__c = 'Educational_Institution', Primary_Affl_Field__c = prefix + 'Primary_Educational_Institution__c'));
-            mappings.add(new Affl_Mappings__c(Name = 'University Department', Account_Record_Type__c = 'University_Department', Primary_Affl_Field__c = prefix + 'Primary_Department__c'));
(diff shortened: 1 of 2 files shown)


src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -39,22 +39,384 @@ public with sharing class AFFL_MultiRecordType_TEST {
...
+    * tests using record type name and contact field api name
+    * @return The affiliation mapping wrappers set up for the tests.
+    **************************************************************************/
+    public static List<AffiliationMappingWrapper> setupAndGetMappingsByDevNameAndApiName() {
+        AFFL_MultiRecordType_TEST.initializeAccountRecordTypeIds();
+
+        List<AffiliationMappingWrapper> affiliationMappingWrapperList =
+            AFFL_MultiRecordType_TEST.getAfflMappingWrapperByDevNameAndApiName();
+
+        return AFFL_MultiRecordType_TEST.setupAndGetMappings(affiliationMappingWrapperList);
```
xinnanlisf, line 58: Also, I am not sure why we want to setupAndGetMappings in setupAndGetMapping...Maybe we want to give a better name to setupAndGetMappings()

src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -1109,188 +1443,396 @@ public with sharing class AFFL_MultiRecordType_TEST {
...
         System.assertEquals(null, primaryAfflLookupAPIName);
-       
+
+    }
+
+/**************************************************************************
+* INNER CLASS MODELS
+**************************************************************************/
+
+private abstract class AffiliationMappingWrapper {
```
xinnanlisf, line 1648: Curious what's the reason why we want to have this new inner class in unit test. We do have affiliationWrapper class already?

xinnanlisf: Per @MatthewBlanski, we will eventually move this to a separate class.

src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -39,22 +39,384 @@ public with sharing class AFFL_MultiRecordType_TEST {
...
+/**************************************************************************
+* SETUP
+**************************************************************************/
+
+    /**************************************************************************
+    * @description Sets up hierarchy settings and affiliation mappings for
+    * tests using record type name and contact field api name
+    * @return The affiliation mapping wrappers set up for the tests.
+    **************************************************************************/
+    public static List<AffiliationMappingWrapper> setupAndGetMappingsByDevNameAndApiName() {
```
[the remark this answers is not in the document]
MatthewBlanski: Removed and get here. In this case we have four scenarios that we support: Record Type Developer Name to Field Label, Record Type Name to Field Label, Record Type Developer Name to Field API name, and Record Type Name to Field API name. Those four entry points explicitly handle those scenarios, giving a one-line setup on the unit test methods that let us know what scenario we are working with up front and perform the multi-line setup there.

The alternative was to attempt to repeat the same setup lines to handle this on nearly every method, which was unappealing.

Ideally we'd have some other means of separation, such as potentially entirely different classes to run the tests through their use cases in the different supported configurations (or collapse this down to record type id and field API name) but that is not the hand we are dealt. I've tried to keep the changes local to this test class.

src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -39,22 +39,384 @@ public with sharing class AFFL_MultiRecordType_TEST {
...
+    }
+
+    /**************************************************************************
+    * @description Initializes the hierarchy settings for testing, setting the
+    * first
+    * @param accountProcessorRecordTypeId The record type id to use for the
+    * account processor
+    * @return The hierarchy settings created by this process.
+    **************************************************************************/
+    private static Hierarchy_Settings__c initializeHierarchySettings(Id accountProcessorRecordTypeId) {
```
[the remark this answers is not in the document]
MatthewBlanski: Removed. We can handle settings better in tests later down the line as per our discussion.

src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -39,22 +39,384 @@ public with sharing class AFFL_MultiRecordType_TEST {
...
+        fieldApiNameList.add(prefix + 'Primary_Household__c');
+        fieldApiNameList.add(prefix + 'Primary_Organization__c');
+        
+        return fieldApiNameList;
+    }
+
+    /**************************************************************************
+    * @description Initializes account record type ids
+    **************************************************************************/
+    private static void initializeAccountRecordTypeIds() {
```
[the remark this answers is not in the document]
MatthewBlanski: I was looking to limit the scope of changes originally, but was hoping to axe the static variables. Also done!

src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -39,22 +39,384 @@ public with sharing class AFFL_MultiRecordType_TEST {
...
+* UTILITIES
+**************************************************************************/
+
+    /**************************************************************************
+    * @description Gets the contact primary affiliation field api name for
+    * Business Organization
+    * @return the contact primary affiliation field api name for Business
+    * Organization
+    **************************************************************************/
+    private static String getBusinessOrganizationFieldApiName() {
```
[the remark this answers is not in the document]
MatthewBlanski: Inverting the order was an excellent callout and solve for this! Done and thank you!

src/classes/AFFL_MultiRecordType_TEST.cls
MatthewBlanski: Renamed this to commonSetup to be much more explicit that it is a set of shared setup steps. We can probably rename this further down the line to something even better.

src/classes/AFFL_MultiRecordType_TEST.cls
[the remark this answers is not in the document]
MatthewBlanski: Per our conversation, a necessary step...for now...to address the record type name and dev name matching the business org record type id pulled from UTIL_Describe_API.

src/classes/AFFL_MultiRecordType_TEST.cls
MatthewBlanski: As additional context, our current wrapper, the Affiliation Mapper, is a pseudo-service that has numerous static collections (such as lists of strings, etc), which does not provide the necessary individual behavior to handle the difference in functionality.

I'm hoping that during our contact refactor we look at replacing that class with something similar to these classes with more oomph so they appropriately represent what we 'know' about the mappings provided.

MatthewBlanski:
As a general comment, we had a few issues here that needed resolving:

1. We had instances where we were calling out to unmanaged fields and a numerical check for 6 fields, rather than the number of fields we had created mappings for. The fix for this was to compare the sizes of created mappings to the results of the tests.
2. In some methods, the mappings were not initialized, which had skated past before as the testing was using the static values cached to slip past. The fix for this was to initialize settings.
3. We had a record type name mismatch downstream in K-12. The fix for this was to determine the record type name and developer name based on the schema description of the record type id we were retrieving from `UTIL_Describe` so we are always comparing apples to apples.
4. We supported 4 scenarios and fixing the record type scenario in light of this required a reworking of how we performed our setups unless we wanted to vastly expand the setup for each unit test.
5. It is my hope that we move some of the code around this setup into individual test classes by scenario, the models into general use with some slight modifications, and generally reimagine how we go about testing this functionality at a later time when we have the bandwidth. Fixing the build issues took priority in this case.

It's worth noting that getting the field labels for the Contact Primary Affiliation fields could also use some attention in the future, but they were not causing build failures so tending to those is also work for later.

src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -36,25 +36,364 @@
...
+    }
+
+    /**************************************************************************
+    * @description Creates affiliation mappings from wrappers, setting the
+    * related mapping to the wrapper for later reference
+    * @param affiliationMappingWrapperList The list of mapping wrappers to
+    * create mappings for
+    * @return The affiliation mappings created 
+    **************************************************************************/
+    private static List<Affl_Mappings__c> createAffiliationMappings(
```
xinnanlisf, line 284: As a separate note, we can potentially put these into UTIL_UnitTestData_TEST. Well actually, it won't hurt to put them now since that class is only used for unit tests but sure let's keep them here for now.

src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -36,25 +36,364 @@
...
+    /**************************************************************************
+    * @description Gets the contact primary affiliation field labels with the 
+    * household field as the first index and the business organization field
+    * as the second index
+    * @return the list of contact primary affiliation field labels
+    **************************************************************************/
+    private static List<String> getContactPrimaryAffiliationFieldLabels() {
+        List<String> fieldLabelList = new List<String>();
+
+        fieldLabelList.add(
```
xinnanlisf, line 314: Aww this looks much nicer now, thank you for the change!

src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -346,27 +694,35 @@ public with sharing class AFFL_MultiRecordType_TEST {
...
+            AFFL_MultiRecordType_TEST.setupMappingsByNameAndLabel();
+        String bizOrgRecordTypeName = affiliationMappingWrapperList[1].getAccountRecordTypeName();
 
         //Turn on Academic Record auto-creation for Business Organization mapping - doesn't make a lot of sense for this type, but just for testing
-        Affl_Mappings__c bizOrgMapping = [select Auto_Program_Enrollment__c from Affl_Mappings__c where Name = 'Business Organization'];
+        Affl_Mappings__c bizOrgMapping = [SELECT Auto_Program_Enrollment__c,
+            Auto_Program_Enrollment_Role__c,
+            Auto_Program_Enrollment_Status__c
+            FROM Affl_Mappings__c
+            WHERE Name = :bizOrgRecordTypeName];
```
lifewithryan, line 724: Name is a potentially encryptable field, this will break shield.

MatthewBlanski: This is a custom setting, but we have a better way to get this anyway since we already have it stored in the mapping wrapper model.

src/classes/AFFL_MultiRecordType_TEST.cls
MatthewBlanski: It wouldn't hurt. I'd like us to look at this comprehensively as this gets pretty zany and centralizing this by the feature might be advantageous for us rather than adding to a list of unit test data in general.

skosgi1:
I think I still have to wrap my head around it completely, but I don't see anything that might break. Thanks for helping with this build failure.

PR #1305, opened 2020-08-25
updating the Run Backfill button on the Accounts and Contacts to leve…

lifewithryan:
# Critical Changes

# Changes
- Updated the Run Backfill button on the Accounts and Contacts tab in EDA Settings to use the existing custom label for that text so that it can be translated.
# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
- Verify the Run Backfill button on the Accounts and Contacts page in EDA Settings still reads "Run Backfill" (the translations themselves will be provided in another release).


dgamboni:
Run Backfill button still says "Run Backfill" in English and "Ejecutar Relleno" in Spanish.
Line 317 was changed as described in WI.
Passing QE.

PR #1309, opened 2020-08-25
[Robot] Validate relationships settings can retain value after save

rhakeshj:
Validate the relationships settings can retain values after making an update and saving it.

Robot WI: W-7989099

New/Existing Test - This is a new test

cci task run robot -o suites robot/EDA/tests/browser/settings/relationships.robot -o vars BROWSER:headlesschrome

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


robot/EDA/tests/browser/settings/relationships.robot
```diff
@@ -0,0 +1,26 @@
...
+Test Setup      Go to EDA settings tab          Relationships
+
+*** Test Cases ***
+Verify relationship settings can retain values on save
+    [Documentation]         Checks the values of 'Reciprocal Method' and 'Allow auto-created
+    ...                     duplicate relations' are retained after making an update to their values
+    ...                     and click on save button
+    [tags]                    unstable        W-7989099
+    Update reciprocal method value      Reciprocal Method=List Setting
+    Update checkbox value       Allow Auto-Created Duplicate Relations=false
```
rjanjanam, line 19: are these updates/edits in the context of the same UI modal ? Then maybe you can have one keyword covering them both and one keyword to validate them

rhakeshj: Yes it is in the same window. But I am planning use update checkbox keyword across pages. That's why I kept it as a separate keyword.

robot/EDA/resources/EDA.py
```diff
@@ -418,3 +436,33 @@ def verify_dropdown_field_status(self, **kwargs):
...
+        """ This method validates the checkbox value for the field passed in kwargs
+            Pass the field name and expected value to be verified from the tests using
+            keyword arguments. true - checked, false - unchecked
+        """
+        self.selenium.execute_javascript("window.scrollTo(0, document.body.scrollHeight)")
+        self.selenium.execute_javascript("window.scrollTo(document.body.scrollHeight, 0)")
+        for field,expected_value in kwargs.items():
+            locator = eda_lex_locators["eda_settings_system"]["default_checkbox"].format(field)
+            self.selenium.page_should_contain_element(locator)
+            self.selenium.wait_until_element_is_visible(locator,
```
skristem, line 450: cant u directly use wait_until_element_is_visible directly? if the locator is not there then it would not be visible

robot/EDA/resources/EDA.py
```diff
@@ -418,3 +436,33 @@ def verify_dropdown_field_status(self, **kwargs):
...
+
+    def verify_dropdown_value(self,**kwargs):
+        """ This method validates the dropdown value for the field passed in kwargs
+            Pass the field name and expected value to be verified from the tests using
+            keyword arguments
+        """
+        for field,expected_value in kwargs.items():
+            locator = eda_lex_locators["eda_settings_system"]["default_dropdown_value"].format(field,expected_value)
+            self.selenium.page_should_contain_element(locator)
+            self.selenium.wait_until_element_is_visible(locator,
```
skristem, line 464: same here

robot/EDA/resources/EDA.py
rhakeshj: That's how it was implemented at first but then while testing I saw inconsistent failures due to wait so I had to add this step to avoid inconsistent failures

robot/EDA/resources/EDA.py
rhakeshj: This step was added to eliminate inconsistent failures

PR #1315, opened 2020-09-01
CCE/EDA: ERROR: hed.TDTM_Affiliation: System.LimitException: hed:Too many SOQL queries: 101

### LINKED ISSUE #1316: EDA: ERROR: hed.TDTM_Affiliation: System.LimitException: hed:Too many SOQL queries: 101
[Issue first reported in the Hub on 8/24/2020
](https://powerofus.force.com/s/question/0D51E00005PABN1SAP/has-anybody-encountered-too-many-soql-queries101-when-importing-contact-records-into-eda)
Bug

When attempting to make updates to Affiliations in bulk AND Enable Record Type Validation is enabled the following error will be encountered: ERROR: hed.TDTM_Affiliation: System.LimitException: hed:Too many SOQL queries: 101

* **Steps to reproduce**

1. Verify Enable Record Type Validation is enabled 
2. Create 33 contacts with 1 primary affiliation (can be the same primary affiliation for all Contacts)
3. Update the Affiliations for all 33 contacts (i.e. change primary from 1 to 0)
4. You encounter the error: ERROR: hed.TDTM_Affiliation: System.LimitException: hed:Too many SOQL queries: 101

Process resolves without a "too many SOQL queries" exception.

Bulk affiliation updates don't resolve

MatthewBlanski:
# Critical Changes
* Fixed affiliation updates and some account updates causing too many soql queries.

# Changes

# Issues Closed
Closes #1316 
# New Metadata

# Deleted Metadata

# Testing Notes


src/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -70,9 +70,10 @@ public with sharing class AFFL_MultiRecordTypeMapper {
     public List<String> primaryAfflFieldNames;
 
     /*******************************************************************************************************
-    * @description Map of Account Record Type Label to API
+    * @description Static Map of Account Record Type Label to API
```
xinnanlisf, line 73: I am not convinced why we are caching this variable. Are we caching it only to provide access to Exception handler class? If we are passing the map as parameter, why do we still need to cache it? IMHO we should avoid using static variables if necessary since it is global state and can be difficult to debug.

src/classes/ERR_ExceptionHandler.cls
```diff
@@ -98,22 +98,30 @@ public with sharing class ERR_ExceptionHandler {
...
-            } else if (affl.Affiliation_Type__c == null) {
-                throw new ERR_ExceptionHandler.AfflAccountNoRecordTypeException(Label.afflAccountNoRecordType);
-            }
+    /*******************************************************************************************************
+    * @description Throws an exception if an affiliation's type is null or its parent account's record type is not in the list of valid record types.
+    * @param affl The affiliation to test.
+    * @param validRecordTypes The list of record types to validate against.
+    * @param accRecTypeLabelToAPI The map of account record type labels to account record type api names
+    */
+    public static void handleAfflNullRecordTypeException(Affiliation__c affl, List<String> validRecordTypes, Map<String,String> accRecTypeLabelToAPI) {
```
xinnanlisf, line 107: This should work but second look on this makes me feel this method is doing something that it shouldn't. Maybe we should let exception handler only handle "exceptions" while letting the tdtm handle the logic by moving this method to the tdtm class. That way we don't have to worry about caching since mapper is cached anyway. Eventually we will have affiliation service to isolate the service logic and trigger context logic.

src/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -1602,85 +1602,121 @@ public with sharing class AFFL_MultiRecordType_TEST {
 
     }
 
+    /**
+     * @description Tests whether the record type labels are cached and matches the size of record types for Account
```
xinnanlisf, line 1606: Like the test for limits. As a complete side note, this test class is over 2k lines and might be already out of control. We should  put it on top of our refactoring list. (maybe we already did)

MatthewBlanski: This class is unusually large and its permutations are complicated. I feel like when we move to a better architecture, we will be able to test these cases separately with more confidence.

src/classes/AFFL_MultiRecordTypeMapper.cls
MatthewBlanski: We are caching this because it is schema information populated by a query. The appropriate place to do this caching post-refactor would be in whatever replaces UTIL_Describe. For the time being, this 'quick' fix is meant to ameliorate needing to perform a SOQL query to get this information more than once a transaction.

src/classes/ERR_ExceptionHandler.cls
MatthewBlanski: This is not in the place I would like it to be. Moving it felt like it was outside the scope of this fix, especially given that we plan to replace this code entirely within the coming months.

PR #1317, opened 2020-09-01
CCE/EDA: ERROR: hed.TDTM_Affiliation: System.LimitException: hed:Too many SOQL queries: 101

### LINKED ISSUE #1316: EDA: ERROR: hed.TDTM_Affiliation: System.LimitException: hed:Too many SOQL queries: 101

MatthewBlanski:
# Critical Changes

# Changes
* Fixed affiliation updates and some account updates causing too many soql queries.
# Issues Closed
Closes #1316 

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/ERR_ExceptionHandler.cls
+++ b/src/classes/ERR_ExceptionHandler.cls
@@ -98,13 +98,24 @@ public with sharing class ERR_ExceptionHandler {
         return false;
     }
 
+    @testVisible
+    private static AFFL_MultiRecordTypeMapper affilMapper {
+        get {
+            if(affilMapper == null) {
+                affilMapper = new AFFL_MultiRecordTypeMapper();
+            }
+
+            return affilMapper;
+        }
+        set;
+    }
+
     public static void handleAfflNullRecordTypeException(Affiliation__c affl, List<String> validRecordTypes) {
         if (affl.Account__c != null) {
             // Get related Account's Record Type Name from 'Affiliation_Type__c' field
             String afflAccRecTypeName;
             if (affl.Affiliation_Type__c != null) {
-                AFFL_MultiRecordTypeMapper affilMapper = new AFFL_MultiRecordTypeMapper();
-                afflAccRecTypeName = affilMapper.accRecTypeLabelToAPI.get(affl.Affiliation_Type__c);
+                afflAccRecTypeName = ERR_ExceptionHandler.affilMapper.accRecTypeLabelToAPI.get(affl.Affiliation_Type__c);
             }
             //If parent account exists for affiliation and account has no record type
             //display the error message
--- a/src/classes/ERR_ExceptionHandler_TEST.cls
+++ b/src/classes/ERR_ExceptionHandler_TEST.cls
@@ -83,5 +83,15 @@ private with sharing class ERR_ExceptionHandler_TEST {
 		ERR_ExceptionHandler.beautifyExceptionMessage(ex);
 
 		system.assertEquals(oldMessage, ex.getMessage(), 'Unhandled exception type should be unmodified.');
-	}
+    }
+    
+    @isTest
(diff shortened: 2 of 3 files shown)


xinnanlisf:
The fix should work. 

Very picky notes alert that we can consider doing when it comes to the refactoring: I am wondering if we want to just put "public static AFFL_MultiRecordTypeMapper afflMapper = new AFFL_MultiRecordTypeMapper();" on the top as other classes just for consistency purpose.  Also, instead of "affilMapper", shall we name it afflMapper?

src/classes/ERR_ExceptionHandler.cls
```diff
@@ -98,13 +98,24 @@ public with sharing class ERR_ExceptionHandler {
...
+            if(affilMapper == null) {
+                affilMapper = new AFFL_MultiRecordTypeMapper();
+            }
+
+            return affilMapper;
+        }
+        set;
+    }
+
     public static void handleAfflNullRecordTypeException(Affiliation__c affl, List<String> validRecordTypes) {
```
kevinswiggum, line 113: Noting here that validRecordTypes is currently including ALL record types that are present in the mappings (not field specific). So I can still put a business organization in the Prim Education Inst field on contact. This PR doesn't fix or further break this issue and so we can address the bug later on.

lifewithryan:
@jofsky this PR fix has been approved, if we can move forward with merging here but we need a RelEng person to do so.

src/classes/ERR_ExceptionHandler.cls
MatthewBlanski: Agreed. This functionality is not appropriately enforced.

src/classes/ERR_ExceptionHandler_TEST.cls
```diff
@@ -83,5 +83,15 @@ private with sharing class ERR_ExceptionHandler_TEST {
 		ERR_ExceptionHandler.beautifyExceptionMessage(ex);
 
 		system.assertEquals(oldMessage, ex.getMessage(), 'Unhandled exception type should be unmodified.');
-	}
+    }
+    
+    @isTest
+    private static void testAffilMapper() {
+        AFFL_MultiRecordTypeMapper afflMapper1 = ERR_ExceptionHandler.affilMapper;
```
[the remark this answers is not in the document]
MatthewBlanski: @kevinswiggum This unit test validates that the caching is working appropriately. Our assumption for this fix is that the first call will create an object and return it and that subsequent calls in the same transaction will return that exact object in order to avoid the constructor running another SOQL query. It also performs an additional validation that the caching is avoiding any additional SOQL queries. If this unit test breaks, then it would serve as an appropriate warning that the caching is no longer serving its intended purpose and that the changes that caused it to break could constitute a breaking change that would reintroduce the bug.

We did miss the mark on this one slightly by not making the property final to further enforce that.

PR #1332, opened 2020-09-14
[EDA] Working EDA TSO

jofsky:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


MatthewBlanski (commented):
@jofsky Having considered another permutation, keeping the help text up to date is likely important as well. That doesn't get push-upgraded, so I believe we need the base definitions of every field if we want to ensure that that is synced. It's quite possible we missed some in the TSO.

PR #1348, opened 2020-10-02
[EDA] : Affiliations Mapping error in EDA

### LINKED ISSUE #1350: An error is thrown on Contact when Primary Affiliation Mappings have a record type that is not available to the running user

An error to be thrown on Contact when Primary Affiliation Mappings have a record type that is not available to the running user. 

No error should be thrown for running users with no access to record types when validation is turned on.

* **Acceptance Criteria**

  - **Given**: I am no access to record types that are available in Primary Affiliation Mappings of EDA settings
  - **When**: I create a Contact with Record Type Validation is turned on
  - **Then**: I should be able to create a Contact

skosgi1:
# Critical Changes

# Changes
We've fixed a customer-reported issue that affected orgs with Enable Record Type Validation enabled in EDA Settings. Before the fix, if the running user didn't have access to an Account record type involved in an Affiliation Mapping, we prevented them from creating or updating a Contact record. Now, we allow the Contact record to be saved.

# Issues Closed
Closes #1350 

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/UTIL_Describe.cls
+++ b/src/classes/UTIL_Describe.cls
@@ -787,7 +787,6 @@ public class UTIL_Describe {
 
         for (SObject rt : results) {
 
-            // Check RecordType IS available for the running user
             if (!Test.isRunningTest()) {
                 recordTypeByDeveloperName.put(String.valueOf(rt.get('DeveloperName')), recordTypeInfos.get(rt.Id));
                 recordTypeByName.put(String.valueOf(rt.get('Name')), recordTypeInfos.get(rt.Id));
--- a/src/classes/UTIL_Describe.cls
+++ b/src/classes/UTIL_Describe.cls
@@ -786,16 +786,8 @@ public class UTIL_Describe {
         Map<Id, Schema.RecordTypeInfo> recordTypeInfos = obj.getRecordTypeInfosByID();
 
         for (SObject rt : results) {
-
-            if (!Test.isRunningTest()) {
-                recordTypeByDeveloperName.put(String.valueOf(rt.get('DeveloperName')), recordTypeInfos.get(rt.Id));
-                recordTypeByName.put(String.valueOf(rt.get('Name')), recordTypeInfos.get(rt.Id));
-            // We don't want our tests to depend on the profile configuration in the org. Otherwise tests
-            // might fail just because the user running them does not have certain record types assigned.
-            } else {
-                recordTypeByDeveloperName.put(String.valueOf(rt.get('DeveloperName')), recordTypeInfos.get(rt.Id));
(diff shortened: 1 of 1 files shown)


src/classes/UTIL_Describe.cls
```diff
@@ -787,12 +787,9 @@ public class UTIL_Describe {
 
         for (SObject rt : results) {
 
-            // Check RecordType IS available for the running user
             if (!Test.isRunningTest()) {
-                if (recordTypeInfos.get(rt.Id).isAvailable()) {
-                    recordTypeByDeveloperName.put(String.valueOf(rt.get('DeveloperName')), recordTypeInfos.get(rt.Id));
-                    recordTypeByName.put(String.valueOf(rt.get('Name')), recordTypeInfos.get(rt.Id));
-                }
+                recordTypeByDeveloperName.put(String.valueOf(rt.get('DeveloperName')), recordTypeInfos.get(rt.Id));
```
lifewithryan, line 791: @MatthewBlanski @skosgi1 it looks like we know longer need the if(!Test.isRunningTest()) check now because (unless I am misreading it) lines 791 and 792 are essentially the same thing that is in the } else { block.  I want to be sure as well that this fix is the right fix. Matt, I heard you mention something about doing this at a different level...lets be sure on this one as all eyes will be upon us.

skosgi1: Looks like they are the same, updated it @lifewithryan

MatthewBlanski (approved):
These changes look good. Tests still pass. Confirmed that existing functionality is preserved for functional mapping, bad label/bad API name for record type, bad label/bad API name for contact field, and the new fix for user does not have a record type assigned.

kevinswiggum (approved):

src/classes/UTIL_Describe.cls
```diff
@@ -786,19 +786,8 @@ public class UTIL_Describe {
         Map<Id, Schema.RecordTypeInfo> recordTypeInfos = obj.getRecordTypeInfosByID();
 
         for (SObject rt : results) {
-
-            // Check RecordType IS available for the running user
-            if (!Test.isRunningTest()) {
-                if (recordTypeInfos.get(rt.Id).isAvailable()) {
```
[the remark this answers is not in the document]
MatthewBlanski: @kevinswiggum This was going to be hooked into other work items in the future. The reality of it was that the instructions in the spec were to replicate the way we pull record types in UTIL_Describe, but to return recordtypeinfos instead (which is what we needed in that case) in order to cache appropriately and get the information we require for the new functionality. When I went to review it, I did not flag that piece as an oddity.

juliesnow (approved):
Updated w/ text approved by Samsheeth.

spinkelman:
QA Passes - Approved to Merge

PR #1356, opened 2020-10-13
[W-8208286] Create ContentDocumentLink Trigger

cbunkers:
# Changes

## New Trigger Handlers
We've added a new Trigger Handler for the ContentDocumentLink object, which you can access for more information:
- [TDTM_ContentDocumentLink](https://github.com/SalesforceFoundation/EDA/blob/master/src/triggers/TDTM_ContentDocumentLink.trigger)

This Trigger Handler doesn't contain logic and doesn't appear in the Trigger Handlers tab. It triggers the product-specific TDTM handlers that are marked Active in the product where the logic is executed. EDA doesn't have product-specific TDTM handlers for this object. 

If you have an existing trigger that calls the TDTM handler and TDTM classes for the ContentDocumentLink and you're not preventing re-entrancy for this object, delete your trigger to avoid calling the classes twice.

# Issues Closed

# New Metadata

- TDTM_ContentDocumentLink (Trigger)

# Testing Notes


src/triggers/TDTM_ContentDocumentLink.trigger-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<ApexTrigger xmlns="http://soap.sforce.com/2006/04/metadata">
+  <apiVersion>49.0</apiVersion>
```
robertbwatson, line 3:
```suggestion
  <apiVersion>48.0</apiVersion>
```

robertbwatson (changes requested):
@cbunkers Looks great! Just one minor suggestion - it looks like EDA is still on v48.0, so I've suggested just using that API version for consistency.

I'm also adding @juliesnow to the review to look over the release notes. This is similar to PR #1208 that we merged in earlier this year.

annejackson:
Testing notes from when @spinkelman tested out PR 1208 that can also be applied for testing this:
Because these were empty triggers with no logic, Sarah verified the following: (can do for this PR as well)

1. Trigger appeared with proper name and version setting in Setup > Apex Triggers and contained NO logic.
2. Triggers did not appear in a search in the Trigger Handler tab. I tested this by entering the trigger name into the Trigger Handler "All Trigger Handlers" list view

robertbwatson (dismissed):

robertbwatson:
I think this is now ready for QA, so I'm assigning this over to @spinkelman for the quick testing listed out above from Anne.

cc @MatthewBlanski for awareness

src/classes/TDTM_TriggerScaffolds_TEST.cls
```diff
@@ -150,4 +150,20 @@ private class TDTM_TriggerScaffolds_TEST {
...
+        //just for test coverage, and to verify no exception gets thrown.
+        insert new ContentVersion(
+            Title = 'Test',
+            PathOnClient = 'Test.jpg',
+            VersionData = Blob.valueOf('Test Content Data')
+        );
+        Application__c a = new Application__c();
+        insert a;
+        Id cdId = [Select Id from ContentDocument Limit 1].Id;
+        insert new ContentDocumentLink(ContentDocumentId = cdId, LinkedEntityId = a.Id);
```
MatthewBlanski, line 164: Please wrap the actual test in start and stop test.
```suggestion
        Test.startTest();
        insert new ContentDocumentLink(ContentDocumentId = cdId, LinkedEntityId = a.Id);
        Test.stopTest();
```

MatthewBlanski (changes requested):
A simple change asking for test start and stop. Everything else checks out. I created an unmanaged TDTM trigger to test this and it ran as expected.

MatthewBlanski (approved):

spinkelman:
QA Passes - Merging
Will Close WI and notify teams

PR #1371, opened 2020-10-30
[Robot] Validate reciprocal relationship settings

rhakeshj:
Verify reciprocal relationship can be added in both read and edit mode
Verify reciprocal relationship can be deleted
Verify existing reciprocal setting can be updated and values retained on save

**Robot WI:**
W-8214822
W-8245483
W-8251460

**New/Existing Test** - This is a new test

cci task run robot -o suites robot/EDA/tests/browser/settings/relationships_reciprocal.robot

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


robot/EDA/resources/RelationshipsSettingsPageObject.py
```diff
@@ -21,13 +24,47 @@ def _is_current_page(self):
...
+    def get_total_settings_count(self):
+        """ This method will return the number of rows in the reciprocal settings"""
+        locator = eda_lex_locators["eda_settings_relationships"]["settings_count"]
+        self.selenium.wait_until_page_contains_element(locator, error="Reciprocal settings table is not available")
+        for i in range(3):
+                i += 1
+                actual_count = int(self.selenium.get_element_count(locator))
+                if actual_count > 0 :
+                    break
+        return actual_count
```
Safiyab, line 66: Why are we looping it ? Does it take a while for the count to update ?

robot/EDA/resources/RelationshipsSettingsPageObject.py
```diff
@@ -42,3 +79,159 @@ def update_reciprocal_method_value(self,**kwargs):
...
+            self.builtin.log(no_of_records)
+            if no_of_records > 0:
+                return int(actual_count)
+            else:
+                raise Exception("The reciprocal settings are not updated")
+        else:
+            if actual_count == settings_count:
+                no_of_records = actual_count - settings_count
+                self.builtin.log(no_of_records)
+                return int(actual_count)
```
Safiyab, line 201: If the counts are equal, why are you calculating the difference and returning it ?

robot/EDA/tests/browser/settings/relationships_reciprocal.robot
```diff
@@ -0,0 +1,127 @@
+*** Settings ***
+Documentation   Validate reciprocal relationships sub tab in EDA Settings
+Resource        robot/EDA/resources/EDA.robot
+Library         cumulusci.robotframework.PageObjects
+...             robot/EDA/resources/RelationshipsSettingsPageObject.py
+Suite Setup     Run keywords
+...             Open Test Browser       AND
+...             Go to EDA settings tab          Relationships      AND
+...             Go to relationships sub tab     Reciprocal Settings     AND
+...             Initialize test data
```
Safiyab, line 10: You should add this either in Test Setup or Suite setup, but not both

robot/EDA/tests/browser/settings/relationships_reciprocal.robot
```diff
@@ -0,0 +1,127 @@
...
+    Click action button on EDA settings page    Save
+    Scroll web page
+    Verify updated reciprocal setting      column=Name1
+    ...                                    Name=Name1
+    ...                                    Female=Female1
+    ...                                    Male=Male1
+    ...                                    Neutral=Neutral1
+
+Verify reciprocal relationship settings can be deleted to clean data
+    [Documentation]         Validates reciprocal settings can be deleted. Also checks if those
```
Safiyab, line 116: Do you want to call this in Teardown instead ? I am sure you thought about it and had some challenges. Just trying to understand the reasoning

Safiyab (approved):

robot/EDA/resources/RelationshipsSettingsPageObject.py
rhakeshj: Yes.Sometime the DOM loads in the background but the active checkbox takes some time to display it in UI

robot/EDA/resources/RelationshipsSettingsPageObject.py
rhakeshj: This is to maintain the uniformity in logging mainly to help readability. I am logging the difference in records but only returning the updated count in both cases which is needed for my test validation

robot/EDA/tests/browser/settings/relationships_reciprocal.robot
rhakeshj: I am accessing the suite variable settings count through UI for which I need the system to login and go to settings sub tab. So I had to keep this step in Suite Set Up. But I added the same in test set up too because my other test cases needs these steps as well. For now I can't think of anything else other than duplicating it once. But recently I met with dev's and identified a way to read the same count through API's. I will have to rework this Suite Setup later and change it to API and removing the UI steps is the plan

robot/EDA/tests/browser/settings/relationships_reciprocal.robot
rhakeshj: I thought keeping more than 5 steps in teardown is not a best way to do it and added as a separate test which looks good. If we had the ability to recall/reuse a test in robot then I would have included it in teardown.

Safiyab (approved):

smparedes (approved):
@rhakeshj Approved.

spinkelman (approved):
Approved. This test matches manual testing steps and verifies all items. Thanks!

PR #1372, opened 2020-10-30
[EDA] - Reciprocal Method Picklist Translation Fix

### LINKED ISSUE #1342: [EDA] - Reciprocal Method Picklist Translation Fix

Bug
The values in Reciprocal Method tab of EDA settings->Relationships->Settings for Reciprocal Method picklist values are hard coded and don't get translated.

The values should be translated when using different languages

localization

* **Acceptance Criteria**

  - **Given**: When I am on EDA settings->Relationships->Settings->
  - **When**: When I see the Reciprocal Method picklist values and when I am using a different locale
  - **Then**: I should have the ability to translate the value using translation files

skosgi1:
# Critical Changes

# Changes
We've created custom labels for List Setting and Value Inversion. These custom labels are referenced by the Reciprocal Method dropdown values under EDA Settings | Relationships | Settings so that those values can now be translated. The Reciprocal Method functionality hasn't changed.

# Issues Closed
Closes #1342 

# New Metadata
Custom Labels:

* stgReciprocalMethodListSetting
* stgReciprocalMethodValueInversion
# Deleted Metadata

# Testing Notes
Case A: Testing in English:

1) On load the default value should be 'List Setting' by default in EDA settings->Relationships->Settings
2) On edit, only 2 values should be visible 'List Setting'  and 'Value Inversion' . 'None' should never be an option.
3) The value in 'Reciprocal Method' field in the custom setting - 'Hierarchy Settings' (Setup -> Custom Settings -> Hierarchy Settings -> Manage) should be 'List Setting'
4) Changing the value in EDA Settings to 'Value Inversion' should change the value in Reciprocal Method' field in the custom setting to 'Value Inversion' as well.
5) Logging out/refreshing should preserve the values in both EDA settings UI and Custom Settings (Hierarchy Settings)

Case B: Testing in Spanish:
1) Add translations to the custom labels in Spanish, Setup -> Custom Label ->  stgReciprocalMethodListSetting -> Under translations, click New -> Language 'Spanish (Mexico)' say, 'Configuración de lista' and for the custom label stgReciprocalMethodValueInversion to  'Inversión de valor'.
2) Switch the language on your user to Spanish, Setup -> User -> User User -> Language -> 'Spanish'
3) Go to EDA settings->Relationships->Settings, the value for the 'Reciprocal Method' should now be 'Configuración de lista' and 'Inversión de valor' on edit screen. 'None' or its translated value should never be an option.
4) The values in 'Hierarchy Settings' (Setup -> Custom Settings -> Hierarchy Settings -> Manage) should still be in English. If the value in EDA settings is 'Configuración de lista' then Hierarchy setting should reflect 'List Settings'. If it is 'Inversión de valor' then the value in Hierarchy settings should be 'Value Inversion' 
5) Custom Settings should always have values in English corresponding to translated values in Spanish
6) Logging out/refreshing should preserve the values in both EDA settings UI and Custom Settings (Hierarchy Settings)

Other Notes:
This required an update to the CMP_Picklist_Dropdown Lighitng component, which is is utilized by the following picklists in EDA settings.
Accounts and Contacts tab: 
1) Select a field for Preferred Phone to copy....
2) Default Contact Language Fluency
On Affiliations -> Settings
1) Status for Affiliations Not Deleted
2) Role for Created Affiliations
3) Status for Created Affiliations
All the above 5 should have 'none' as an option.
--


lifewithryan (commented):
I don't see it here but maybe this is addressed elsewhere (in another branch) but we have some code in Apex that explicitly refers to these values. Do we know if they will work when translated? Specfically lines 303 and 388 in REL_Relationships_TDTM.cls. There are also some test classes, and telemetry. Maybe it's fine and apex always uses the non-translated value, not sure but would like to be confident in our answer.

skosgi1:
> here (in another branch) but we have some code in Apex that explicitly refers to these values. Do we know if they will work when translated? Specfically lines 303 and 388 in REL_Relationships_TDTM.cls. There are also some test classes, and telemetry. Maybe it's fine and apex always uses the non-translated value, not sure but would like to be confident in our answer.

We may have to change the TDTM, but I will have to double check on the telemetry @lifewithryan thanks for catching this.

skosgi1:
> > here (in another branch) but we have some code in Apex that explicitly refers to these values. Do we know if they will work when translated? Specfically lines 303 and 388 in REL_Relationships_TDTM.cls. There are also some test classes, and telemetry. Maybe it's fine and apex always uses the non-translated value, not sure but would like to be confident in our answer.
> 
> We may have to change the TDTM, but I will have to double check on the telemetry @lifewithryan thanks for catching this.

Updated the aura components to show labels on UI and store English values in the custom settings, no changes to the underlying apex has been made.

src/aura/STG_CMP_Rel/STG_CMP_Rel.cmp
```diff
@@ -6,7 +6,9 @@
     <aura:attribute name="Id" type="String" />
     <aura:attribute name="noRecSettings" type="String" />
     <aura:attribute name="noAutoCreateSettings" type="String" />
-    <aura:attribute name="reciprocalMethodOptions" type="String[]" default="List Setting,Value Inversion"/>
+    <aura:attribute name="reciprocalMethodOptions" type="Map" />
```
MatthewBlanski, line 9: This was a good solution to handle back end vs front end names.

MatthewBlanski (approved):
Happy path testing works out.
Switching between the two values displays and saves appropriately. Relationship functionality still works for reciprocal relationship creation.

juliesnow:
@skosgi1 - Updated PR text per approval via Slack.

dgamboni:
Tested / Passes QA 
FYI @spinkelman

PR #1380, opened 2020-11-10
Preferred Email does not respect Custom Email Fields

skosgi1:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes
Update to ensure logic works in Release orgs.


MatthewBlanski (approved):
This looks like a good way to invert the logic and handle the appropriate scenario

spinkelman:
Passes first QA check. Will be testing again after merge, since original bug was found in Release orgs after merge only.

PR #1386, opened 2020-11-10
[Robot] Update settings to use API call

rhakeshj:
This is a change to an existing test in which UI was used to get the no of records in reciprocal relationship settings. Now the same functionality is tested using API.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/robot/EDA/resources/EDA.robot
+++ b/robot/EDA/resources/EDA.robot
@@ -94,3 +94,12 @@ Create Organization Foundation
     ${account_id} =            Get Current Record Id
     Store Session Record       Account  ${account_id}
     [return]                   ${account_id}
+
+Get Records Count
+    [Documentation]         Returns the no of record identified by the given field_name and
+    ...                     field_value input for a specific object
+    [Arguments]             ${obj_name}             ${field_name}       ${field_value}
+    ${result} =             SOQL Query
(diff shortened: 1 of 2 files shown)


robot/EDA/resources/EDA.robot
```diff
@@ -94,3 +94,12 @@ Create Organization Foundation
     ${account_id} =            Get Current Record Id
     Store Session Record       Account  ${account_id}
     [return]                   ${account_id}
+
+Get Records Count
+    [Documentation]         Returns the no of record identified by the given field_name and
+    ...                     field_value input for a specific object
+    [Arguments]             ${obj_name}             ${field_name}       ${field_value}
```
Safiyab, line 101: Instead of ${field_name} and ${field_value}, it is better to have &{fields}. So this keyword can be used to filter multiple WHERE conditions when required.
Anyway, you don't have to modify it now if it is not required, but just a suggestion for the future.

Safiyab (approved):
Really happy with this change. It will make the setup stable and reliable.

robot/EDA/resources/EDA.robot
rhakeshj: Sure and noted. Thank you!

PR #1395, opened 2020-11-19
Fix beta test failure due to update_admin_profile managed option

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


xinnanlisf:
@jofsky, I believe this is the fix. I still can't run beta test form local but @skosgi1 has run this and seems to be working. Would you mind giving it a quick look? Thanks a lot!

jofsky:
@xinnanlisf I'm working on a better fix hold on.

xinnanlisf:
> @xinnanlisf I'm working on a better fix hold on.

awesome!

cumulusci.yml
```diff
@@ -232,6 +232,12 @@ flows:
     config_managed:
         # description: Configure an org for use as a dev org after package metadata is deployed.
         steps:
+            2:
+                task: update_admin_profile
+                options:
+                    managed: True
```
davidmreed, line 238: In theory this shouldn't be needed with CumulusCI 3.23.0 and higher - it knows how to infer the `managed` and `namespaced_org` options. Are you seeing behavior that indicates otherwise?

PR #1396, opened 2020-11-20
[Robot] Validate hierarchy settings and backfill settings

rhakeshj:
Verify backfill settings when course connections is enabled
Verify org wide custom settings are not over written by user custom settings

Robot WI:
W-8448330
W-8448341

New/Existing Test - This is a new test

cci task run robot -o suites robot/EDA/tests/browser/settings/hierarchy_settings.robot
cci task run robot -o suites robot/EDA/tests/browser/settings/course_connections_backfill.robot

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes


robot/EDA/resources/CourseConnectionsSettingsPageObject.py
```diff
@@ -42,6 +49,20 @@ def set_enable_course_connections(self):
                 return
         raise Exception("Clicking element 'Enable Course Connections' failed after multiple tries")
 
+    def select_backfill_checkbox(self):
```
skristem, line 52: might be a good idea to check first the status of the checkbox. If it is already checked then u do not want to change the status. I am thinking of scenarios where the test case with this keyword gets executed multiple times in the same org.

robot/EDA/resources/SetupPageObject.py
```diff
@@ -13,10 +15,34 @@ def _go_to_page(self):
...
     def perform_a_quick_find_on(self, item):
         """ Searches for the given item through Quick Find on Setup page """
         self.eda.populate_placeholder("Quick Find", item)
 
+    def select_frame(self, value):
+        """ Selects iframe using the value passed from the test. The value can be iframe id, title
+            etc
+        """
+        locator = eda_lex_locators["eda_setup"]["frame"].format(value)
+        time.sleep(4)
```
skristem, line 43: can u wait for the frame instead of just sleep here

robot/EDA/resources/locators_50.py
```diff
@@ -25,6 +25,12 @@
         "tab": "//div[@class='uiTabBar']/ul[@class='tabs__nav']/li[contains(@class,'uiTabItem')]/a[@class='tabHeader']/span[contains(text(), '{}')]",
         "spl-tab": "//div[@class='slds-tabs_default']//ul[@class='slds-tabs_default__nav']/li[contains(@class,'slds-tabs_default__item')]/a[text()= '{}']",
     },
+    "eda_setup": {
+        "custom_settings": "//a[text()='{}']",
```
skristem, line 29: since this is a very common locator, you might want to move it to the outer dictionary. Just a suggestion

robot/EDA/resources/CourseConnectionsSettingsPageObject.py
rhakeshj: Yeah I thought about it. But this case is unique here. In this page, this checkbox enables the button to initiate a batch/queue kinda process. Once that is done and if the user moves out of the screen, the check box automatically flips to unchecked. So for every iteration this checkbox will load as unchecked avoiding the necessary to check for it

robot/EDA/resources/CourseConnectionsSettingsPageObject.py
```diff
@@ -42,6 +49,20 @@ def set_enable_course_connections(self):
...
+        """ Selects the backfill checkbox """
+        locator = eda_lex_locators["eda_settings_cc"]["backfill_checkbox"]
+        self.selenium.wait_until_page_contains_element(locator)
+
+        for i in range(3):
+            i += 1
```
Safiyab, line 58: You don't have to increment, the for loop will take care of it

robot/EDA/resources/SetupPageObject.py
```diff
@@ -13,10 +15,34 @@ def _go_to_page(self):
...
+        locator = eda_lex_locators["eda_setup"]["settings_action_button"].format(button)
+        self.selenium.wait_until_page_contains_element(locator)
+        self.selenium.wait_until_element_is_visible(locator)
+        self.selenium.click_element(locator)
+
     def perform_a_quick_find_on(self, item):
         """ Searches for the given item through Quick Find on Setup page """
         self.eda.populate_placeholder("Quick Find", item)
 
+    def select_frame(self, value):
```
Safiyab, line 38: This is a more generic keyword. I think you should move it to EDA.py

robot/EDA/tests/browser/settings/hierarchy_settings.robot
```diff
@@ -0,0 +1,33 @@
...
+    Select accounts contacts checkbox
+    ...         Account Types with Multi-Addresses Enabled=Household Account
+    Click action button on EDA settings page    Save
+    Load Page Object                     Home                          Setup
+    Go to custom settings setup
+    Select frame        Custom Settings ~ Salesforce - Developer Edition
+    Click custom settings       Hierarchy Settings
+    Select frame        Custom Setting Definition ~ Salesforce - Developer Edition
+    Click custom settings action button     Manage
+    Select frame        Custom Setting Hierarchy Settings ~ Salesforce - Developer Edition
```
Safiyab, line 27: I think this all can be one keyword like: `Go to hierarchy settings page`
You don't need to show this level of detail in a test for a setup keyword. Especially the select and unselect frame needs to be hidden in the implementation.

robot/EDA/tests/browser/settings/hierarchy_settings.robot
```diff
@@ -0,0 +1,33 @@
...
+    Verify setup owner section      No records to display.
+    Unselect frame
+    Go to custom settings setup
```
Safiyab, line 30: Why is this the last step in the file ?

Safiyab (approved):
Made some suggestions. Can you please take a look and do the updates (they are nice-to have but not necessarily blocking)

robot/EDA/resources/CourseConnectionsSettingsPageObject.py
rhakeshj: That makes the iterator variable unused (other than we can use it for logging the iteration). We could use while for retries but again it will go undefined limits of attempt until the condition passes

robot/EDA/tests/browser/settings/hierarchy_settings.robot
rhakeshj: Agreed. I will move these under one keyword.

robot/EDA/tests/browser/settings/hierarchy_settings.robot
rhakeshj: This is added to make sure it leaves out of the frames (like switch to default frame)

robot/EDA/resources/SetupPageObject.py
rhakeshj: @skristem This has been updated and moved to EDA.py as suggested

robot/EDA/resources/locators_50.py
rhakeshj: Already had a similar locator in eda dictionary and made use of it and this is removed now

robot/EDA/resources/CourseConnectionsSettingsPageObject.py
rhakeshj: Removed the iterator increment and replaced with logging

robot/EDA/resources/SetupPageObject.py
rhakeshj: Done

robot/EDA/tests/browser/settings/hierarchy_settings.robot
rhakeshj: Moved to a common keyword now

robot/EDA/tests/browser/settings/hierarchy_settings.robot
rhakeshj: Moved this step to Test Teardown

robot/EDA/resources/locators_50.py
skristem: I still see the locator

robot/EDA/tests/browser/settings/course_connections_backfill.robot
```diff
@@ -7,20 +7,39 @@ Library         cumulusci.robotframework.PageObjects
...
+    Click action button on EDA settings page    Save
+    Select course connections subtab            Backfill
+    Select backfill checkbox
+    Click backfill button
+    Verify backfill toast message       The backfill was queued successfully.
+    Reload Page
+    Go to EDA settings tab          Course Connections
+    Select course connections subtab            Backfill
+    Verify backfill checkbox value              false
+
```
skristem, line 42: I suggest adjusting the formatting for easy readability in both this and below test

skristem (approved):
minor suggestions looks good otherwise

robot/EDA/tests/browser/settings/hierarchy_settings.robot
```diff
@@ -0,0 +1,35 @@
...
+    ...         Account Types with Multi-Addresses Enabled=Household Account
+    Click action button on EDA settings page    Save
+    Load Page Object                     Home                          Setup
+    Go to custom settings setup
+    Select hierarchy settings        Custom Settings ~ Salesforce - Developer Edition
+    ...                             Hierarchy Settings
+    Wait until loading is complete
+    Go to hierarchy settings view       Custom Setting Definition ~ Salesforce - Developer Edition
+    ...                                 Manage
+    Wait until loading is complete
```
Safiyab, line 30: I think you can hide the frame value and the waits too. This is what I had in mind:
```
Go to custom settings in setup
Select hierarchy settings           mode=Manage
```

You can play around with the level of details here. But leave this for another day and another PR.
This test looks and works great.

Safiyab (approved):

dgamboni (approved):
Functional review looks great, thank you!

spinkelman (approved):
After Functional Review, This item is approved.

PR #1398, opened 2020-11-20
[EDA] Fix to resolve update admin profile record type  insertion

jofsky:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -221,9 +221,9 @@ tasks:
                 - record_type: "Account.HH_Account"
                 - record_type: "Account.Sports_Organization"
                 - record_type: "Account.University_Department"
-                - record_type: "{managed}Attribute__c.Credential"
+                - record_type: "{managed}Attribute__c.{managed}Credential"
                   default: true
-                - record_type: "{managed}Attribute__c.Student_Characteristic"
+                - record_type: "{managed}Attribute__c.{managed}Student_Characteristic"
                 - record_type: "{managed}Course_Enrollment__c.Default"
                 - record_type: "{managed}Course_Enrollment__c.Faculty"
                 - record_type: "{managed}Course_Enrollment__c.Student"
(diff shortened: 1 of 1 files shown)


cumulusci.yml
```diff
@@ -221,9 +221,9 @@ tasks:
                 - record_type: "Account.HH_Account"
                 - record_type: "Account.Sports_Organization"
                 - record_type: "Account.University_Department"
-                - record_type: "{managed}Attribute__c.{managed}Credential"
+                - record_type: "{managed}Attribute__c.Credential"
```
jstvz, line 224: I don't think this is the culprit. If you look below: https://github.com/SalesforceFoundation/EDA/blob/51829b6dd849392e241131106769fcc0151bc094/cumulusci.yml#L300-L308

I missed the old `skip_record_types` option. Looks like it would stop after setting visibility for `Account`: https://github.com/SalesforceFoundation/EDA/blob/dbd4f51eb9f55ac19e5f046aa90f9dce05658c57/tasks/salesforce.py#L101-L102

What you want is the old behavior, so you'll should restore the `{managed}` token here and update `config_packaging` to:

```yaml
    config_packaging:
        # description: Configure packaging org for upload after package metadata is deployed
        steps:
            1:
                # Configure update_admin_profile
                options:
                    record_types:
                        - record_type: "Account.Administrative"
                          default: true
                        - record_type: "Account.Academic_Program"
                        - record_type: "Account.Business_Organization"
                        - record_type: "Account.Educational_Institution"
                        - record_type: "Account.HH_Account"
                        - record_type: "Account.Sports_Organization"
                        - record_type: "Account.University_Department"
```

jstvz (changes requested):

jstvz (approved):

PR #1404, opened 2020-11-23
[EDA] Config_qa flow modification to enable 2gp feature testing

jofsky:
Modified the 3rd and 5th steps of `config_qa` flow to only run when EDA is not an installed package in the org.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -28,7 +28,6 @@ tasks:
         options:
             namespace_inject: $project_config.project__package__namespace
             path: unpackaged/config/dev
-            unmanaged: True
 
     delete_dev_config:
         description: Removes Development configuration. Sets page layouts, compact layouts to system defaults. Removes record type visibilites.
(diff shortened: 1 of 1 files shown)


cumulusci.yml
```diff
@@ -222,9 +221,6 @@ tasks:
                 - record_type: "Account.HH_Account"
                 - record_type: "Account.Sports_Organization"
                 - record_type: "Account.University_Department"
-                - record_type: "{managed}Attribute__c.{managed}Credential"
```
davidmreed, line 225: Is this change intended to be merged?

jofsky: This is a question I have for @jstvz

jstvz: We should restore these.

jstvz (approved):

MatthewBlanski (approved):

Issue #1405, opened 2020-11-24
[EDA ] New Error Messages for  invalidContactFieldInAffiliationMapping and  invalidRecordTypeInAffiliationMapping


m-kinkade:
**Note: FOR INTERNAL USE ONLY. For support questions, please use the EDA Chatter group in [The Power of Us Hub](https://powerofus.force.com).
For feature requests, please use the Ideas tab in [The Power of Us Hub](https://powerofus.force.com).**

* **Do you want to request a *feature* or report a *bug*?**
Feature
* **What is the current behavior?**
When affiliation mappings with invalid primary affiliation fields or record types exist and a user creates or updates a Contact invalid mapped fields populated, affiliations for the invalid mappings will not be generated, but the user will not be notified of the issue.

* **What is the expected behavior?**
When affiliation mappings with invalid primary affiliation fields or record types exist, affiliations will not be created and users will be notified of the mapping responsible for the error when attempting to create or update Contacts with invalid affiliation mappings.

* **What is the motivation / use case for changing the behavior?**
To improve the current user experience.

* **Acceptance Criteria**

  - **Given**: I have an affiliation mapping with an invalid mapped Contact Primary Affiliation field on Contact that is not an Account Lookup Field (such as a checkbox)
  - **When**: I try to create a contact with the mapped field populated
  - **Then**: I receive an error telling me which mapping is invalid.

  - **Given**: I have an affiliation mapping with an invalid mapped Account Record Type
  - **When**: I try to create a contact with the mapped field populated
  - **Then**: I receive an error telling me which mapping is invalid.

m-kinkade:
[W-8449077](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008jE4eIAE/view)

kporray1:
This issue was delivered in Summer '21 / SFDO Release 232

PR #1407, opened 2020-11-26
[EDA ] New Error Messages for  invalidContactFieldInAffiliationMapping and  invalidRecordTypeInAffiliationMapping

m-kinkade:
# Critical Changes

# Changes
- We now provide custom error messages to indicate that a Contact is being created or updated with a value in one of the Primary Affiliation fields on Contact that has an invalid value for Account Record Type or Contact Primary Affl Field under EDA Settings, click Affiliations | Affiliation Mappings. The message provides information to help users correct such errors.

# Issues Closed
Closes Issue #1405 

# New Metadata
### Custom Labels
invalidContactFieldInAffiliationMapping
invalidRecordTypeInAffiliationMapping

# Deleted Metadata

# Testing Notes
https://salesforce.quip.com/VgRQAkbuVQqN


src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -767,6 +767,274 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    }
+
+    /*************************************************************************************************************
+    * @description Test case to verify that inserting a contact with a primary affiliation field populated with record type
+    * validation enabled, a valid Contact field mapping and invalid Record Type mapping throws an error.
+    **************************************************************************************************************/
+    @isTest 
+    private static void insertContactAfflInvalidRecordTypeError(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+        //insert testContactsList;
```
xinnanlisf: is this commented out on purpose?

src/classes/SRVC_Contact_PrimaryAffiliations.cls
```diff
@@ -221,10 +224,15 @@ public virtual with sharing class SRVC_Contact_PrimaryAffiliations {
         Set<Id> accountIds = new Set<Id>();
 
         for (MODL_PrimaryAfflMapper eachModel: afflMappingsModelsList) {
-
+            //try{
```
xinnanlisf, line 227: Should we get rid of these commented out code?

src/classes/SRVC_Contact_PrimaryAffiliations.cls
```diff
@@ -170,27 +174,26 @@ public virtual with sharing class SRVC_Contact_PrimaryAffiliations {
 
         try{
             for (MODL_PrimaryAfflMapper eachModel: afflMappingsModelsList) {
-
                 if (newContact.get(eachModel.getContactFieldApiName()) == oldContact.get(eachModel.getContactFieldApiName())) {
                     continue;
                 }
-
+
```
xinnanlisf, line 180: looks like adding spacing here

xinnanlisf (changes requested):
Great work @m-kinkade! especially thank you for being a hawkeye and fixing all the typos in test classes!
Left some minor complains. I also do have a concern about extending the constructor by adding a new parameter. Do you happen to know if people are using that model or writing their own unit tests? It looks like you are fixing EDA's unit tests too. Which can be a risk because It might break unit tests for dependent packages and worse case their logic. Should we keep the old constructor and create a new one? Let me know if I am overthinking. Also, cc @MatthewBlanski

MatthewBlanski (commented):
> Great work @m-kinkade! especially thank you for being a hawkeye and fixing all the typos in test classes!
> Left some minor complains. I also do have a concern about extending the constructor by adding a new parameter. Do you happen to know if people are using that model or writing their own unit tests? It looks like you are fixing EDA's unit tests too. Which can be a risk because It might break unit tests for dependent packages and worse case their logic. Should we keep the old constructor and create a new one? Let me know if I am overthinking. Also, cc @MatthewBlanski

In this case we don't need to worry about dependent packages for our models. Our models are not global and deliberately so! Being able to extend and add to them is one of their benefits. The models are also the appropriate place to handle the information for the error messages, so I feel like this is the right decision.

In this case, the constructors modified appear to be those in which we need to have said error messages. In this pattern it is superior to do so as we have extensive unit tests that detect undesired behavior and provide a list of what we need to change. Were it for an additional unnecessary parameter that had a default, my gut would be to add another constructor.

xinnanlisf:
> > Great work @m-kinkade! especially thank you for being a hawkeye and fixing all the typos in test classes!
> > Left some minor complains. I also do have a concern about extending the constructor by adding a new parameter. Do you happen to know if people are using that model or writing their own unit tests? It looks like you are fixing EDA's unit tests too. Which can be a risk because It might break unit tests for dependent packages and worse case their logic. Should we keep the old constructor and create a new one? Let me know if I am overthinking. Also, cc @MatthewBlanski
> 
> In this case we don't need to worry about dependent packages for our models. Our models are not global and deliberately so! Being able to extend and add to them is one of their benefits. The models are also the appropriate place to handle the information for the error messages, so I feel like this is the right decision.
> 
> In this case, the constructors modified appear to be those in which we need to have said error messages. In this pattern it is superior to do so as we have extensive unit tests that detect undesired behavior and provide a list of what we need to change. Were it for an additional unnecessary parameter that had a default, my gut would be to add another constructor.

Sweet! Thank you Matt!

xinnanlisf (dismissed):
Looks good. Approved!

src/classes/SRVC_Contact_PrimaryAffiliations.cls
```diff
@@ -87,27 +88,30 @@ public virtual with sharing class SRVC_Contact_PrimaryAffiliations {
...
-                    throw new ERR_ExceptionHandler.AffAccountRecordTypeMappingsException(Label.afflAccoutMappingError);
+                    // mapped contact field is valid, but value provided is not a Lookup to a contact Id
+                    contact.addError(eachModel.getInvalidErrorMessage());
+                    throw new ERR_ExceptionHandler.AffAccountRecordTypeMappingsException(eachModel.getInvalidErrorMessage());
                 } catch (System.StringException stringExc) {
-                    Contact.addError(stringExc.getMessage());
-                    throw new ERR_ExceptionHandler.AffAccountRecordTypeMappingsException(Label.afflAccoutMappingError);
+                    // mapped contact field is valid, but value provided is not a string 
+                    contact.addError(eachModel.getInvalidErrorMessage());
+                    throw new ERR_ExceptionHandler.AffAccountRecordTypeMappingsException(eachModel.getInvalidErrorMessage());
```
xinnanlisf, line 114: Have a quick question on this. Looks like we will never throw this exception right? We are adding error to the contact trigger?

m-kinkade: I think this exception could occur when the value in getContactFieldApiName is not a valid ID?

MatthewBlanski: That's accurate. Depending on the type of field passed in there can be a number of different explicit exceptions we are catching for.

MatthewBlanski (approved):

juliesnow:
Updated description based on discussion w/ Merle.

src/classes/SRVC_Contact_PrimaryAffiliations.cls
```diff
@@ -170,27 +174,25 @@ public virtual with sharing class SRVC_Contact_PrimaryAffiliations {
...
                 if (newContact.get(eachModel.getContactFieldApiName()) == oldContact.get(eachModel.getContactFieldApiName())) {
                     continue;
                 }
-
+               
                 if (newContact.get(eachModel.getContactFieldApiName()) != null) {
-                    if (eachModel.getIsValid() == false){
-                        newContact.addError(Label.afflAccoutMappingError);
-                        throw new ERR_ExceptionHandler.AffAccountRecordTypeMappingsException(Label.afflAccoutMappingError);
+                    if (eachModel.getIsValid() == false || String.isBlank(eachModel.getInvalidErrorMessage()) == false){
```
xinnanlisf, line 182: Want to make sure: based on eda style guide, should we use `myString.isBlank()` instead?

xinnanlisf: Also, we are throwing exceptions when model is valid but there is invalid error message? That doesn't sound so right?

m-kinkade: I think the syntax for the isBlank() method of the String class is String.isBlank(myString).  I tried myString.isBlank() and I get an error.

xinnanlisf: yeah we will need to update the style guide then. Also, really up to you, but isNotBlank == true would be more readable?

src/classes/UTIL_Describe.cls
```diff
@@ -1068,4 +1068,20 @@ public class UTIL_Describe {
...
         return sObjectTypeToChildRelationshipNames.get(sObjectTypeName);
     }
+
+    /**
+    * @description Determine if a specified field is a lookup to the specified SObjectType
+    * @param  fieldDescribe The Schema.DescribeFieldResult of a specified field
+    * @param  sObjectType   The SObjectType for the lookup object
+    * @return True if the specified field is a lookup to the specified SObjectType, false otherwise
+    */ 
+    public static Boolean isLookupFieldTo(Schema.DescribeFieldResult fieldDescribe, SObjectType sObjectType){
```
xinnanlisf, line 1078: Also, at the end of this line, need an extra spacing between ) and {. Also applies to other lines in this block.

xinnanlisf (changes requested):

src/classes/SRVC_Contact_PrimaryAffiliations.cls
MatthewBlanski: isBlank(thing) == false may be clear enough. Do we want to add a local variable to accomplish this? Is it adding enough value to be incorporated and named?

xinnanlisf: > Also, we are throwing exceptions when model is valid but there is invalid error message? That doesn't sound so right?
What to bring this up again. Do we want to allow model with isValid =true but invalidErrorMsg not blank to throw exceptions?
@m-kinkade @MatthewBlanski

m-kinkade: In this scenario, the isValid flag is used to indicate whether the Record Type mapping in affiliation mappings is a valid Account Record Type when validation is enabled.  The error message variable is used to capture any errors associated with the model (in this case, invalid Record Type or Contact field mapping).  Maybe we could rename isValid to be more specific, like isRecordTypeValid instead?

xinnanlisf: @MatthewBlanski curious what's your opinion on this.

MatthewBlanski: For the purposes of this PR, let's leave off of that variable. Eventually we will want to modify that value to capture a variety of states. We may wish to do this within a context of when we address a mapper when we are working off of the Affiliation object.

If you think the LOE is low, then please, go ahead and rename it.

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
     ****************************************************** FUNCTIONAL TESTS ***************************************************
     **************************************************************************************************************************/
 
+
+    /**************************************************************************************************************************************
+    * @description Helper method to generate error message associated with a mapping to a field that does not lookup to Account.
+    * @param  affiliationMapping An affiliation mapping with an invalid lookup field mapping.
+    * @return The expected error message associated with the generated model.
+    **************************************************************************************************************************************/
+    private static String getExpectedInvalidFieldMappingErrorMessage(Affl_Mappings__c affiliationMapping){
```
MatthewBlanski, line 777: This is a good example of a helper method. The comments were useful and meaningful during the code review.

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * validation enabled, a valid Contact field mapping and invalid Record Type mapping throws an error.
+    **************************************************************************************************************/
+    @isTest 
+    private static void insertContactAfflInvalidRecordTypeError(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = true));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 811: In these situations, we can use `UTIL_Namespace.StrTokenNSPrefix(String)` to namespace prefix an SObject or field. I prefer using this as it leverages the caching and other handling in UTIL_Namespace. If there is a readability issue, then local variables might assist. This isn't a required change here, but something I'd prefer for us to do in the future since we already have the utility in place to handle it.

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+        }
+
+        Test.startTest();
+        try {
+            insert testContactsList;
+        } catch (System.DmlException err){
+            System.assertEquals(true, err.getMessage().contains(CON_PrimaryAffls_TEST.getExpectedInvalidRTMappingErrorMessage(mappings[0])), 'Invalid Record Type mapping should throw an error when validation is enabled.');
+        }
+        
+        Test.stopTest();
```
MatthewBlanski, line 833: Two things. First, we can laser-focus the test block for metrics. I only validatated that this was possible recently. Second, we should put a `System.assert(false,String)` just before the catch block to have this test fail if we don't have any exception at all.
```suggestion
        try {
            Test.startTest();
            insert testContactsList;
            Test.stopTest();
        } catch (System.DmlException err){
            System.assertEquals(true, err.getMessage().contains(CON_PrimaryAffls_TEST.getExpectedInvalidRTMappingErrorMessage(mappings[0])), 'Invalid Record Type mapping should throw an error when validation is enabled.');
        }
```

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    }
+
+    /*************************************************************************************************************
+    * @description Test case to verify that inserting a Contact with a primary affiliation field populated with record type
+    * validation enabled, a valid Contact field mapping that is not a lookup to Account and valid Record Type mapping throws an error.
+    **************************************************************************************************************/
+    @isTest 
+    private static void insertContactAfflInvalidFieldMappingError(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
```
MatthewBlanski, line 842: At some later date we may want to move bulk testing into our unmanaged unit tests. This is a good practice for now so we can make certain we're not hitting CPU timeouts with this cantankerous code.

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+        for (Contact con : testContactsList){
+            con.Gender__c = 'Decline to State';
+        }
+
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = true));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 850: Again about UTIL_Namespace.StrTokenNSPrefix(String). I'd like us to get into the habit of using it.

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * an error when the mapped field is populated.
+    **************************************************************************************************************/
+    @isTest 
+    private static void insertContactAfflInvalidRTAndInvalidFieldMappingFieldPopulatedInvalidValidationOffFieldError(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = false));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 908: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+
+        List<Affl_Mappings__c> mappings = new List<Affl_Mappings__c>();
+        mappings.add(new Affl_Mappings__c(Name = 'Business Organization', Account_Record_Type__c = 'Not a Valid RecordType', Primary_Affl_Field__c = prefix + 'FirstName'));
+        mappings.add(new Affl_Mappings__c(Name = 'Household Account', Account_Record_Type__c = 'HH_Account', Primary_Affl_Field__c =  prefix + 'Primary_Household__c'));
+        insert mappings;
+
+        Test.startTest();
+        try {
+            insert testContactsList;
```
MatthewBlanski, line 917: Likewise here for putting the test block on this specific line and `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * throw an error when the mapped field is not populated.
+    **************************************************************************************************************/
+    @isTest 
+    private static void insertContactAfflInvalidRTAndInvalidFieldMappingFieldNotPopulatedNoError(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = false));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 937: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+
+        Test.startTest();
+        insert testContactsList;       
+        Test.stopTest();
+
+        List<Contact> testContactResultsList = [SELECT Id, Primary_Organization__c
+                                                FROM Contact 
+                                                WHERE Id IN :testContactsList];
+
+        System.assertEquals(100, testContactResultsList.size());
```
MatthewBlanski, line 952: Do we want to validate expected field values here in a loop? It might add value.

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
     ****************************************************** FUNCTIONAL TESTS ***************************************************
     **************************************************************************************************************************/
 
+
+    /**************************************************************************************************************************************
```
MatthewBlanski, line 772: For this class, do we want to consider using `Database.insert(List<SObject>,false)` and `SObject.getErrors()` to detect that the error is specifically added to each contact? It may be a more specific test and add value.

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * throw an error when the mapped field is not populated.
+    **************************************************************************************************************/
+    @isTest 
+    private static void insertContactAfflInvalidRTAndInvalidFieldMappingFieldNotPopulatedPositive(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+       
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = false));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 967: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+
+        Test.startTest();
+            insert testContactsList;
+        Test.stopTest();
+
+        List<Contact> testContactResultsList = [SELECT Id, Primary_Organization__c
+                                                FROM Contact 
+                                                WHERE Id IN :testContactsList];
+
+        System.assertEquals(100, testContactResultsList.size());
```
MatthewBlanski, line 982: Do we want to validate expected field values here in a loop? It might add value.

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    **************************************************************************************************************/
+    @isTest 
+    private static void updateContactAfflInvalidRecordTypeError(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+        insert testContactsList;
+
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = true));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 997: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+        Account businessAccount1 = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe_API.getBizAccRecTypeID())[0];
+        insert businessAccount1;
+
+        for (Contact con : testContactsList){
+            con.Primary_Organization__c = businessAccount1.Id;
+        }
+
+        Test.startTest();
+        try {
+            update testContactsList;
```
MatthewBlanski, line 1014: Another consideration of the single-line test block and the `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    **************************************************************************************************************/
+    @isTest 
+    private static void updateContactAfflInvalidFieldMappingError(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+        insert testContactsList;
+
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = true));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1034: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+        insert businessAccount1;
+
+        for (Contact con : testContactsList){
+            con.Primary_Organization__c = businessAccount1.Id;
+            con.Gender__c = 'Decline to State';
+        }
+
+        Test.startTest();
+        try {
+            update testContactsList;
```
MatthewBlanski, line 1052: Another consideration of the single-line test block and the `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    **************************************************************************************************************/
+    @isTest 
+    private static void updateContactAfflInvalidRTAndInvalidFieldMappingFieldPopulatedInvalidRecordTypeError(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+        insert testContactsList;
+
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = true));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1073: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+        mappings.add(new Affl_Mappings__c(Name = 'Household Account', Account_Record_Type__c = 'HH_Account', Primary_Affl_Field__c =  prefix + 'Primary_Household__c'));
+        insert mappings;
+
+        for (Contact con : testContactsList){
+            con.Gender__c = 'Decline to State';
+        }
+
+        Test.startTest();
+        try {
+            update testContactsList;
```
MatthewBlanski, line 1086: Another consideration of the single-line test block and the `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+        for (Contact con : testContactsList){
+            con.FirstName = '';
+        }
+        insert testContactsList;
+
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = true));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1110: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    **************************************************************************************************************/
+    @isTest 
+    private static void updateContactAfflInvalidRTAndValidFieldMappingFieldNotPopulatedNoValidationPositive(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+        insert testContactsList;
+
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = false));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1151: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    **************************************************************************************************************/
+    @isTest 
+    private static void updateContactAfflInvalidRTAndInValidFieldMappingFieldNotPopulatedNoValidationPositive(){
+        List<Contact> testContactsList = UTIL_UnitTestData_TEST.getMultipleTestContacts(100);
+        insert testContactsList;
+
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = false));
+
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1192: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -768,51 +768,492 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+        mappings.add(new Affl_Mappings__c(Name = 'Household Account', Account_Record_Type__c = 'HH_Account', Primary_Affl_Field__c =  prefix + 'Primary_Household__c'));
+        insert mappings;
+
+        for (Contact con : testContactsList){
+            con.Gender__c = 'Decline to State';
+        }
+
+        Test.startTest();
+        try {
+            update testContactsList;
```
MatthewBlanski, line 1205: Another consideration of the single-line test block and the `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * @description Tests to ensure that when a Contact is inserted with a primary affiliation field populated, but the 
+    * mapping is to a field that is not of type lookup to Account, an invalidContactFieldInAffiliationMapping error is thrown.
+    **************************************************************************************************************/
+    @isTest
+    private static void insertContactWithValidRecordTypeMappingAndInvalidFieldMappingToAccountLookup() {
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = true));       
+        
+        // insert mappings
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1402: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+            Contact con = UTIL_UnitTestData_TEST.getUniqueContact(i);
+            con.ReportsToId = reportToContact.Id;
+            contactsToBeInserted.add(con);
+        }
+
+        String errorMessageText = CON_PrimaryAffls_TEST.getinvalidContactFieldInAffiliationMappingMessage('ReportsToId');
+
+        Test.startTest();
+        try{
+            insert contactsToBeInserted;
```
MatthewBlanski, line 1423: Another consideration of the single-line test block and the `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+            Contact con = UTIL_UnitTestData_TEST.getUniqueContact(i);
+            con.ReportsToId = reportToContact.Id;
+            contactsToBeInserted.add(con);
+        }
+
+        String errorMessageText = CON_PrimaryAffls_TEST.getInvalidRecordTypeInAffiliationMappingMessage('Business_Organization_Invalid');
+
+        Test.startTest();
+        try{
+            insert contactsToBeInserted;
```
MatthewBlanski, line 1462: Another consideration of the single-line test block and the `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * mapping is to a field that is not of type lookup to Account, Record Type mapping is invalid and validation is 
+    * enabled, an invalidRecordTypeInAffiliationMapping error is thrown.
+    **************************************************************************************************************/
+    @isTest
+    private static void insertContactWithInvalidRTMappingAndInvalidFieldMappingToAccountLookupValidationOn() {
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = true));       
+        
+        // insert mappings
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1441: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+            Contact con = UTIL_UnitTestData_TEST.getUniqueContact(i);
+            con.ReportsToId = reportToContact.Id;
+            contactsToBeInserted.add(con);
+        }
+
+        String errorMessageText = CON_PrimaryAffls_TEST.getinvalidContactFieldInAffiliationMappingMessage('ReportsToId');
+
+        Test.startTest();
+        try{
+            insert contactsToBeInserted;
```
MatthewBlanski, line 1501: Another consideration of the single-line test block and the `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * mapping is to a field that is not of type lookup to Account, Record Type mapping is invalid and validation is 
+    * disabled, an invalidContactFieldInAffiliationMapping error is thrown.
+    **************************************************************************************************************/
+    @isTest
+    private static void insertContactWithInvalidRTMappingAndInvalidFieldMappingToAccountLookupValidationOff() {
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = false));       
+        
+        // insert mappings
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1480: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * mapping is to a field that is not of type lookup to Account, Record Type mapping is invalid and validation is 
+    * disabled, no error is thrown and the insert completes as expected.
+    **************************************************************************************************************/
+    @isTest
+    private static void insertContactWithNoPrimaryAfflFieldPopulated() {
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = false));       
+        
+        // insert mappings
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1519: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * @description Tests to ensure that when a Contact is updated with a primary affiliation field populated, but the 
+    * mapping is to a field that is not of type lookup to Account, an invalidContactFieldInAffiliationMapping error is thrown.
+    **************************************************************************************************************/
+    @isTest
+    private static void updateContactWithValidRecordTypeMappingAndInvalidFieldMappingToAccountLookup() {
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = true));       
+        
+        // insert mappings
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1560: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+
+        for (Contact con : contactsToBeInserted){
+            con.ReportsToId = reportToContact.Id;
+        }
+
+        String errorMessageText = CON_PrimaryAffls_TEST.getinvalidContactFieldInAffiliationMappingMessage('ReportsToId');
+
+        Test.startTest();
+        try{
+            update contactsToBeInserted;
```
MatthewBlanski, line 1586: Another consideration of the single-line test block and the `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * mapping is to a field that is not of type lookup to Account, Record Type mapping is invalid and validation is 
+    * enabled, an invalidRecordTypeInAffiliationMapping error is thrown.
+    **************************************************************************************************************/
+    @isTest
+    private static void updateContactWithInvalidRTMappingAndInvalidFieldMappingToAccountLookupValidationOn() {
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = true));       
+        
+        // insert mappings
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1604: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+
+        for (Contact con : contactsToBeInserted){
+            con.ReportsToId = reportToContact.Id;
+        }
+
+        String errorMessageText = CON_PrimaryAffls_TEST.getInvalidRecordTypeInAffiliationMappingMessage('Business_Organization_Invalid');
+
+        Test.startTest();
+        try{
+            update contactsToBeInserted;
```
MatthewBlanski, line 1630: Another consideration of the single-line test block and the `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+
+        for (Contact con : contactsToBeInserted){
+            con.ReportsToId = reportToContact.Id;
+        }
+
+        String errorMessageText = CON_PrimaryAffls_TEST.getinvalidContactFieldInAffiliationMappingMessage('ReportsToId');
+
+        Test.startTest();
+        try{
+            update contactsToBeInserted;
```
MatthewBlanski, line 1674: Another consideration of the single-line test block and the `System.assert(false,String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * mapping is to a field that is not of type lookup to Account, Record Type mapping is invalid and validation is 
+    * disabled, an invalidContactFieldInAffiliationMapping error is thrown.
+    **************************************************************************************************************/
+    @isTest
+    private static void updateContactWithInvalidRTMappingAndInvalidFieldMappingToAccountLookupValidationOff() {
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = false));       
+        
+        // insert mappings
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1648: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/CON_PrimaryAffls_TEST.cls
```diff
@@ -952,12 +1389,348 @@ private with sharing class CON_PrimaryAffls_TEST {
...
+    * mapping is to a field that is not of type lookup to Account, Record Type mapping is invalid and validation is 
+    * disabled, no error is thrown and the update completes as expected.
+    **************************************************************************************************************/
+    @isTest
+    private static void updateContactWithNoPrimaryAfflFieldPopulated() {
+        UTIL_CustomSettings_API.getSettingsForTests(new Hierarchy_Settings__c(Affiliation_Record_Type_Enforced__c = false));       
+        
+        // insert mappings
+        String namespace = UTIL_Namespace.getNamespace();
+        String prefix = (String.isNotBlank(namespace)) ? namespace + '__' : '';
```
MatthewBlanski, line 1692: Another `UTIL_Namespace.StrTokenNSPrefix(String)`

src/classes/MAPR_PrimaryAffiliations_TEST.cls
```diff
@@ -574,11 +599,135 @@ private class MAPR_PrimaryAffiliations_TEST {
...
+        MAPR_PrimaryAffiliations_TEST.insertAfflMappingsWithInaccurateRecordTypeNameAndNonLookupPrimaryFieldLabels();
+        MAPR_PrimaryAffiliations maprPrimaryAfflInstance = new STUB_PrimaryAffiliationValidationOff();
+
+        List<Affl_Mappings__c> primaryAfflMappings = UTIL_CustomSettingsFacade.getPrimaryAfflMappings();
+
+        Test.startTest();
+            MODL_PrimaryAfflMapper primaryAfflModel = maprPrimaryAfflInstance.getModelFromMapping(primaryAfflMappings[0]);
+        Test.stopTest();
+
+        String prefix = UTIL_Namespace.StrTokenNSPrefix('');
```
MatthewBlanski, line 616: This is the first step in applying the prefix and is a superior use of the utilities at our disposal. I would prefer that we use this around the specific String in question getting prefixed.

src/classes/MAPR_PrimaryAffiliations_TEST.cls
```diff
@@ -574,11 +599,135 @@ private class MAPR_PrimaryAffiliations_TEST {
...
+        MAPR_PrimaryAffiliations_TEST.insertAfflMappingsWithValidRecordTypeNameAndNonLookupPrimaryFieldLabels();
+        MAPR_PrimaryAffiliations maprPrimaryAfflInstance = new STUB_PrimaryAffiliationValidationOff();
+
+        List<Affl_Mappings__c> primaryAfflMappings = UTIL_CustomSettingsFacade.getPrimaryAfflMappings();
+
+        Test.startTest();
+            MODL_PrimaryAfflMapper primaryAfflModel = maprPrimaryAfflInstance.getModelFromMapping(primaryAfflMappings[0]);
+        Test.stopTest();
+
+        String prefix = UTIL_Namespace.StrTokenNSPrefix('');
```
MatthewBlanski, line 641: Another `UTIL_Namespace.StrTokenNSPrefix(String)` on the specific String in question.

src/classes/MAPR_PrimaryAffiliations_TEST.cls
```diff
@@ -574,11 +599,135 @@ private class MAPR_PrimaryAffiliations_TEST {
...
+        MAPR_PrimaryAffiliations_TEST.insertAfflMappingsWithInaccurateRecordTypeNameAndNonLookupPrimaryFieldLabels();
+        MAPR_PrimaryAffiliations maprPrimaryAfflInstance = new STUB_PrimaryAffiliationValidationOn();
+
+        List<Affl_Mappings__c> primaryAfflMappings = UTIL_CustomSettingsFacade.getPrimaryAfflMappings();
+
+        Test.startTest();
+            MODL_PrimaryAfflMapper primaryAfflModel = maprPrimaryAfflInstance.getModelFromMapping(primaryAfflMappings[0]);
+        Test.stopTest();
+
+        String prefix = UTIL_Namespace.StrTokenNSPrefix('');
```
MatthewBlanski, line 666: Another `UTIL_Namespace.StrTokenNSPrefix(String)` on the specific String in question.

src/classes/MAPR_PrimaryAffiliations_TEST.cls
```diff
@@ -574,11 +599,135 @@ private class MAPR_PrimaryAffiliations_TEST {
...
+        MAPR_PrimaryAffiliations_TEST.insertAfflMappingsWithInaccurateRecordTypeNameAndAccuratePrimaryFieldLabels();
+        MAPR_PrimaryAffiliations maprPrimaryAfflInstance = new STUB_PrimaryAffiliationValidationOn();
+
+        List<Affl_Mappings__c> primaryAfflMappings = UTIL_CustomSettingsFacade.getPrimaryAfflMappings();
+
+        Test.startTest();
+            MODL_PrimaryAfflMapper primaryAfflModel = maprPrimaryAfflInstance.getModelFromMapping(primaryAfflMappings[0]);
+        Test.stopTest();
+
+        String prefix = UTIL_Namespace.StrTokenNSPrefix('');
```
MatthewBlanski, line 691: Another `UTIL_Namespace.StrTokenNSPrefix(String)` on the specific String in question.

src/classes/MAPR_PrimaryAffiliations_TEST.cls
```diff
@@ -574,11 +599,135 @@ private class MAPR_PrimaryAffiliations_TEST {
...
+        MAPR_PrimaryAffiliations_TEST.insertAfflMappingsWithValidRecordTypeNameAndNonLookupPrimaryFieldLabels();
+        MAPR_PrimaryAffiliations maprPrimaryAfflInstance = new STUB_PrimaryAffiliationValidationOn();
+
+        List<Affl_Mappings__c> primaryAfflMappings = UTIL_CustomSettingsFacade.getPrimaryAfflMappings();
+
+        Test.startTest();
+            MODL_PrimaryAfflMapper primaryAfflModel = maprPrimaryAfflInstance.getModelFromMapping(primaryAfflMappings[0]);
+        Test.stopTest();
+
+        String prefix = UTIL_Namespace.StrTokenNSPrefix('');
```
MatthewBlanski, line 716: Another `UTIL_Namespace.StrTokenNSPrefix(String)` on the specific String in question.

src/classes/MODL_PrimaryAfflMapper.cls
```diff
@@ -52,7 +53,13 @@ public class MODL_PrimaryAfflMapper {
     * @param settingsRecordTypeValue 'Account Record Type' value from EDA Settings.
     * @param isValid Boolean flag to indicate whether the Record Type Affiliation maps to a valid Account record type.
     *******************************************************************************************************/ 
-    public MODL_PrimaryAfflMapper(Schema.RecordTypeInfo accountRecordTypeInfo, Schema.DescribeFieldResult contactDescribeFieldResult, String settingsContactFieldValue, String settingsRecordTypeValue, Boolean isValid) {
+    public MODL_PrimaryAfflMapper(Schema.RecordTypeInfo accountRecordTypeInfo,
```
MatthewBlanski, line 56: Expanding the constructors here was the correct call. We now have another required field for a valid mapping that allows us to inform the user of the specific validation failure.

src/classes/UTIL_Describe.cls
```diff
@@ -1068,4 +1068,20 @@ public class UTIL_Describe {
...
         return sObjectTypeToChildRelationshipNames.get(sObjectTypeName);
     }
+
+    /**
+    * @description Determine if a specified field is a lookup to the specified SObjectType
+    * @param  fieldDescribe The Schema.DescribeFieldResult of a specified field
+    * @param  sObjectType   The SObjectType for the lookup object
+    * @return True if the specified field is a lookup to the specified SObjectType, false otherwise
+    */ 
+    public static Boolean isLookupFieldTo(Schema.DescribeFieldResult fieldDescribe, SObjectType sObjectType) {
```
MatthewBlanski, line 1078: This is a good use of the tools at our disposal and hybridizing it with the requirements. Since we cache the mappings, the use of this is also efficient and performant.

src/classes/UTIL_Describe_TEST.cls
```diff
@@ -362,4 +362,52 @@ public with sharing class UTIL_Describe_TEST {
...
+
+    /**
+    * @description Test to ensure isLookupFieldTo() returns true when a specified field is of type lookup to the 
+    * specified SObject type.
+    */ 
+    @isTest 
+    private static void isLookupToAccountTrue(){
+
+        SObject conObj = new Contact(LastName = 'testCon');
+        Schema.DescribeFieldResult acctFieldDesc = conObj.getSObjectType().getDescribe().fields.getMap().get('AccountId').getDescribe();
```
MatthewBlanski, line 374: Within the context of this test, why not use a pre-existing combination of functions to get this information from UTIL_Describe? This is an effective and specific test, so there is value in keeping it isolated in the way you have built it as well.

src/classes/UTIL_Describe_TEST.cls
```diff
@@ -362,4 +362,52 @@ public with sharing class UTIL_Describe_TEST {
...
+    @isTest 
+    private static void isLookupToAccountFalse(){
+        SObject conObj = new Contact(LastName = 'testCon');
+        Schema.DescribeFieldResult reportsToFieldDesc = conObj.getSObjectType().getDescribe().fields.getMap().get('ReportsToId').getDescribe();
+        
+        Test.startTest();
+        Boolean isLookupResult = UTIL_Describe.isLookupFieldTo(reportsToFieldDesc, Account.SObjectType);
+        Test.stopTest();
+
+        System.assertEquals(false, isLookupResult, 'Should return false since reportsToFieldDesc field on contact is lookup to Contact, not Account.');
```
MatthewBlanski, line 396: These messages for assertions are clear and precise. They are exemplary.

MatthewBlanski (changes requested):
After manually testing the behavior against the spec, everything does work as expected. I encountered an odd piece of behavior that keyed back into the Affiliation's portion of the trigger code that does not impact this PR.

Of the comments I included, the only necessary changes are adding a `System.assert(false,String)` in the locations where we expect an exception so we validate that those tests fail in the expected fashion and do not pass as a false positive.

For the other comments, please balance the value of making those changes against the level of effort in implementing them as they are nice to haves.

PR #1432, opened 2020-12-09
MODL, MAPR, SRVC classes for Feature Parameter

skosgi1:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/FeatureParameterDateModel.cls
+++ b/src/classes/FeatureParameterDateModel.cls
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2020, Salesforce.org
+    Copyright (c) 2021, Salesforce.org
     All rights reserved.
     Redistribution and use in source and binary forms, with or without
     modification, are permitted provided that the following conditions are met:
--- a/src/classes/FeatureParameterDateModel.cls
+++ b/src/classes/FeatureParameterDateModel.cls
@@ -33,10 +33,10 @@
 */
 public with sharing class FeatureParameterDateModel extends FeatureParameterModel{
 
-    public Datetime value { get; private set; }
+    public Date value { get; private set; }
 
     /* Constructor to set the properties */
-    public FeatureParameterDateModel(String name, Datetime value) {
+    public FeatureParameterDateModel(String name, Date value) {
         super(name);
         this.value = value;
     }
--- a/src/classes/FeatureParameterDateModel.cls
+++ b/src/classes/FeatureParameterDateModel.cls
@@ -1,0 +1,52 @@
+/*
+    Copyright (c) 2020, Salesforce.org
+    All rights reserved.
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+    * Redistributions of source code must retain the above copyright
+      notice, this List of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this List of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of Salesforce.org nor the names of
+      its contributors may be used to endorse or promote products derived
+      from this software without specific prior written permission.
+    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+    FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+    INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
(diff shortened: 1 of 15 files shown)


src/classes/FeatureParameterModel.cls
```diff
@@ -0,0 +1,50 @@
...
+        this.name = name;
+    }
+
+    /******************************************************************************
+    * @description Virtual method to be overridden by child classes
+    * @return null
+    ******************************************************************************/
+    public virtual String convertToString() {
+        return null;
+    }
```
MatthewBlanski, line 49: We're getting code coverage hits on this. Let's convert this to an abstract method.
```suggestion
    public abstract String convertToString() {}
```

src/classes/FeatureParameterDateModel.cls
```diff
@@ -0,0 +1,52 @@
...
+/**
+* @author Salesforce.org
+* @date 2021
+* @group 
+* @group-content ../../ApexDocContent/
+* @description Model class for Feature Parameter Date that extends MODL_FeatureParameter
+*/
+public with sharing class FeatureParameterDateModel extends FeatureParameterModel{
+
+    public Datetime value { get; private set; }
```
MatthewBlanski, line 36: We have to go date on this since datetime isn't a supported type for Feature Parameter. Only Date is.
```suggestion
    public Date value { get; private set; }
```

src/classes/FeatureParameterDateModel.cls
```diff
@@ -0,0 +1,52 @@
...
+
+    /* Constructor to set the properties */
+    public FeatureParameterDateModel(String name, Datetime value) {
```
MatthewBlanski, line 39: Same with here for Date
```suggestion
    public FeatureParameterDateModel(String name, Date value) {
```

src/classes/FeatureParameterDateModel_TEST.cls
```diff
@@ -0,0 +1,67 @@
...
+private class FeatureParameterDateModel_TEST {
+
+    /*********************************************************************************************************************
+    * @description Test method to verify FeatureParameterDateModel assigns values to properties
+    **********************************************************************************************************************/
+    @isTest
+    private static void FeatureParameterDateModelConstructor() {
+
+        Test.startTest();
+            FeatureParameterDateModel featureParamDateModel = new FeatureParameterDateModel('testFeature', System.now());
```
MatthewBlanski, line 44: We'll have to use Date.today() instead of System.now().
```suggestion
            FeatureParameterDateModel featureParamDateModel = new FeatureParameterDateModel('testFeature', Date.today());
```

src/classes/FeatureParameterMapper.cls
```diff
@@ -0,0 +1,69 @@
...
+        return instance;
+    }
+
+    /*************************************************************************************************************
+    * @description Method to return a featureParameterModel
+    * @Params featureParameter The feature Parameter to return the model for
+    * @return featureParamDateModel The model for the featureParameter with with value set to the current datetime
+    **************************************************************************************************************/
+    public FeatureParameterModel getFeatureParameterModel(String featureParameter) {
+        FeatureParameterDateModel featureParamDateMdl = new FeatureParameterDateModel(featureParameter, System.now());
```
MatthewBlanski, line 66: Date conversion here.
```suggestion
        FeatureParameterDateModel featureParamDateMdl = new FeatureParameterDateModel(featureParameter, Date.today());
```

src/classes/FeatureParameterMapper_TEST.cls
```diff
@@ -0,0 +1,83 @@
...
+    ***************************************************************************************************************************/
+    @isTest
+    private static void getFeatureParameterModel() {
+
+        Test.startTest();
+            FeatureParameterModel featureParamModel = FeatureParameterMapper.getInstance().getFeatureParameterModel('testFeatureParam');
+        Test.stopTest();
+
+        System.assertEquals('testFeatureParam', featureParamModel.name, 'Feature Parameter Name should be the one which was passed');
+        System.assertEquals(System.now(), ((featureParameterDateModel)featureParamModel).value, 'Feature Parameter value should be returned');
```
MatthewBlanski, line 80: Likewise here. We may want to make this method specific to an expected FeatureParameterDateModel.
```suggestion
        System.assertEquals(Date.today(), ((featureParameterDateModel)featureParamModel).value, 'Feature Parameter value should be returned');
```

src/classes/FeatureParameterService_TEST.cls
```diff
@@ -0,0 +1,96 @@
...
+
+        System.assertEquals(FeatureParameterMapper.getInstance(), featureParamMapper, 'Should return the instance of FeatureParameterMapper');
+
+    }    
+
+    /**************************************************************************************************************************
+    * @description Test method to verify that the getFeatureParameterModel method returns FeatureParameterModel
+    ***************************************************************************************************************************/
+    @isTest
+    private static void getFeatureParameterModel() {
```
MatthewBlanski, line 87: Let's change this to use a stubbed FeatureParameterMapper that always returns an expected value, which in this case would be a new instance of a FeatureParameterDateModel. We can then check against that to ensure an encapsulated passthrough.

MatthewBlanski (changes requested):
A few changes requested. We should switch to an abstract method for convertToString() on FeatureParameterModel instead of virtual. This does force the issue of implementation of that method in dependent classes, which I am comfortable with.

We will need to convert over to a Date instead of DateTime to match the exact datatype Feature Parameters use.

Last, let's start stubbing tests on the Feature Parameter Service. That will survive any changes we make into the future and meaningfully tests that we get what we are expecting from the mapper.

skosgi1:
> A few changes requested. We should switch to an abstract method for convertToString() on FeatureParameterModel instead of virtual. This does force the issue of implementation of that method in dependent classes, which I am comfortable with.
> 
> We will need to convert over to a Date instead of DateTime to match the exact datatype Feature Parameters use.
> 
> Last, let's start stubbing tests on the Feature Parameter Service. That will survive any changes we make into the future and meaningfully tests that we get what we are expecting from the mapper.

Thanks for the feedback @MatthewBlanski , I have addressed the above. Can you please do a second review? Thanks.

MatthewBlanski (approved):
This is now good to go for merge

PR #1444, opened 2020-12-11
[EDA] Running a Settings Health Check: Date Feature Parameter

skosgi1:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


src/classes/FeatureParameterMapper_TEST.cls
```diff
@@ -66,18 +66,174 @@ private class FeatureParameterMapper_TEST {
...
+    * @description Test method to verify that the getFeatureParameterModel method returns a FeatureParameterModel
+    ***************************************************************************************************************************/
+    @isTest
+    private static void getFeatureParameterModelSuccess() {
+
+        Date featureParameterDate = Date.valueOf('2015-1-1');
+        String featureParameterName = 'SettingsHealthCheckLastRunDate';
+
+        // Set value for the Feature Parameter
+        System.FeatureManagement.setPackageDateValue(featureParameterName, featureParameterDate);
```
skosgi1, line 151: @MatthewBlanski wanted to check if we can set the featureParameter value here in the test method or do we want call the update updateFromFeatureParameterModel using stub and then do a getFeatureParameterModel()?

src/classes/FeatureParameterMapper.cls
```diff
@@ -63,7 +109,71 @@ public virtual with sharing class FeatureParameterMapper {
...
+
+    }
+
+    /*************************************************************************************************************
+    * @description Returns the enum equivalent of a String if found else returns null
+    * @Params toBeConvertedString The String that needs to be converted to Enum
+    * @return TelemetryParameterName enum equivalent of the String passed
+    **************************************************************************************************************/
+    @TestVisible
+    private TelemetryParameterName convertToEnum(String toBeConvertedString) {
```
skosgi1, line 162: @MatthewBlanski slightly updated the conversion and error handling w.r.t to the SPEC, wanted to check with you on this if this is ok?

src/lwc/healthCheck/healthCheck.js
```diff
@@ -1,24 +1,22 @@
...
-    @wire(getHealthCheckViewModel)
-    healthCheckViewModel({error, data}){
-        if (data){
-            this.lastRunDate = data.lastRunDate;
-        } else if(error){
+    @wire(getHealthCheckViewModel) healthCheckViewModel;
 
-        }
-    }
+    @wire(updateHealthCheckLastRunDate) healthCheckLastRunDate;
```
MatthewBlanski, line 21: Are we using this anywhere? I don't believe we need this wire or method.

unpackaged/gated/healthCheck/lwc/healthCheck/healthCheck.js
```diff
@@ -9,16 +10,14 @@ export default class HealthCheck extends LightningElement {
...
-    @wire(getHealthCheckViewModel)
-    healthCheckViewModel({error, data}){
-        if (data){
-            this.lastRunDate = data.lastRunDate;
-        } else if(error){
+    @wire(getHealthCheckViewModel) healthCheckViewModel;
 
-        }
-    }
+    @wire(updateHealthCheckLastRunDate) healthCheckLastRunDate;
```
MatthewBlanski, line 22: Likewise here. I do not believe we need this.

src/classes/HealthCheckVMapper.cls
```diff
@@ -88,25 +88,31 @@ public virtual with sharing class HealthCheckVMapper {
     * @description Retrieve a view model representing a health check to run.
     * @return A HealthCheckVModel representing details of a health check to be run.
     **********************************************************************************/
-    public virtual HealthCheckVModel getHealthCheckViewModel(){
-        FeatureParameterDateModel featureParamDateModel = 
-           (FeatureParameterDateModel)this.locateFeatureParameterService().getFeatureParameterModel();
-        String lastRunDate = featureParamDateModel.convertToString(); 
+    public virtual HealthCheckVModel getHealthCheckViewModel() {
```
MatthewBlanski, line 91: I may have missed this in a previous review. This class uses a number of tabs. Please fix the spacing issues to use four spaces.

MatthewBlanski (changes requested):
There are a number of changes, some having to do with genericization and another having to do with changing a method signature that seems to be a better pattern in the code review so we can get a theoretically arbitrary feature parameter from the feature parameter service.

A few of the others have to do with LWC skilling up and how we can and cannot access methods.

src/classes/FeatureParameterMapper_TEST.cls
```diff
@@ -66,18 +66,150 @@ private class FeatureParameterMapper_TEST {
...
+    @isTest
+    private static void getFeatureParameterModelEnumNotDate() {
+        try {
+            Test.startTest();
+                FeatureParameterModel featureParamModel = FeatureParameterMapper.getInstance().getFeatureParameterModel('HasUserManagedTDTM');
+                System.assert(false, 'When Passed parameter string is not an enum an error is thrown');
+            Test.stopTest();
+        } catch (FeatureParameterMapper.FeatureParameterMapperException ex) {
+            System.assert(ex.getMessage().contains('Feature Parameter'));
+            System.assert(ex.getMessage().contains('is not defined.'));
```
skosgi1, line 111: @MatthewBlanski would this be an ideal way to handle asserts? Do we want to rearrange the exception in main class to accommodate the entire string in 1 assert?

skosgi1:
> There are a number of changes, some having to do with genericization and another having to do with changing a method signature that seems to be a better pattern in the code review so we can get a theoretically arbitrary feature parameter from the feature parameter service.
> 
> A few of the others have to do with LWC skilling up and how we can and cannot access methods.

Thanks for the review @MatthewBlanski , I have addressed the feedback, can you please give it another shot? Thx.

MatthewBlanski (changes requested):
A few more changes requested on this. We're getting closer. There are some suggestions for stubs and mocks. There are other suggestions for how to go about implementation in a manner that adds value.

src/classes/FeatureParameterMapper_TEST.cls
```diff
@@ -66,18 +66,126 @@ private class FeatureParameterMapper_TEST {
...
+          System.assert(ex.getMessage().contains('Feature Parameter ' + featureParameter + ' is not defined.'));
+        }
+
+    }
+
+    /********************************************************************************************************************************************
+    * @description Test method to verify that the updateFromFeatureParameterModel method updates the feature parameter date
+    *********************************************************************************************************************************************/
+     @isTest
+     private static void updateFromFeatureParameterModelPoistiveTest() {
```
MatthewBlanski, line 173: A simple typo
```suggestion
     private static void updateFromFeatureParameterModelPositiveTest() {
```

src/classes/HealthCheckVMapper.cls
```diff
@@ -73,40 +73,39 @@ public virtual with sharing class HealthCheckVMapper {
...
+            String lastRunDate = featureParamDateModel.convertToString();
 
-        for (HealthCheckDefinitionModel healthChkDefModel : healthCheckDefinitionList){
-            //Generate view model for current health check definition model
-            HealthCheckDefinitionVModel healthChkDefViewModel = new HealthCheckDefinitionVModel(healthChkDefModel.name,
-                                                                                                healthChkDefModel.className,
-                                                                                                healthChkDefModel.namespace);
+            if (lastRunDate == '12/31/2015') {
+                lastRunDate = '';
+            }
```
MatthewBlanski, line 88: Since Date.format() returns an arbitrary format based on the user's locale and other settings, we cannot guarantee the string output. It would suggest setting lastRunDate to an empty string. Then, if featureParamDateModel.value is not that a strongly-typed date of 12/31/2015, set lastRunDate to featureParamDateModel.convertToString().

MatthewBlanski (changes requested):
A few minor changes

src/classes/HealthCheckVMapper.cls
```diff
@@ -73,40 +73,39 @@ public virtual with sharing class HealthCheckVMapper {
...
 
-        Integer numberOfRuns = healthCheckDefinitionList.size();
+            Integer numberOfRuns = healthCheckDefinitionList.size();
+
+            return new HealthCheckVModel(lastRunDate, numberOfRuns, healthCheckDefViewModelsList);
+        }
+        catch (FeatureParameterMapper.FeatureParameterMapperException e) {
+            throw new AuraHandledException(Label.stgHealthCheckErrorLastRunDate);
+        }
```
MatthewBlanski, line 108: @skosgi1 Please make a unit test for this using a stub.

src/classes/HealthCheckVMapper_TEST.cls
```diff
@@ -119,6 +119,25 @@ public class HealthCheckVMapper_TEST {
...
+        FeatureParameterMapper.instance = new STUB_FeatureParameterMapperError();
+        HealthCheckDefinitionService.instance = new STUB_HealthCheckDefinitionService();
+
+        try {
+            Test.startTest();
+                HealthCheckVModel healthChkViewMod = HealthCheckVMapper.getInstance().getHealthCheckViewModel();
+            Test.stopTest();
+            System.assert(false, 'Should throw an exception');
+        } catch (AuraHandledException ex) {
+            System.assertEquals('System.AuraHandledException', ex.getTypeName(), 'AuraHandledException should be thrown when exception occurs.');
```
MatthewBlanski, line 137: This is the right play here since the exception in this case is a passthrough with an arbitrary message.

MatthewBlanski (approved):

PR #1464, opened 2020-12-23
Record Type Service

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


unpackaged/test/account_model_health_checks/pre/classes/RecordTypeMapper_Unmanaged_TEST.cls
```diff
@@ -0,0 +1,58 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2020
+* @group
+* @group-content ../../ApexDocContent/Tests.htm
+* @description Unmanaged test class for RecordTypeMapper
+*/
+@isTest
+private class RecordTypeMapper_Unmanaged_TEST {
```
MatthewBlanski, line 35: This class is not added to the unmanaged tests portion of our flows. It also does not compile.

unpackaged/test/account_model_health_checks/pre/classes/RecordTypeMapper_Unmanaged_TEST.cls
```diff
@@ -0,0 +1,58 @@
...
+    * Make sure that no inactive record type is included in this map.
+    ***************************************************************************************************************************/
+    @isTest
+    private static void getModelByIdMapForSObjectTypeAccountInactiveRecordTypeTest() {
+        Test.startTest();
+            Map<Id, RecordTypeModel> recordTypeModelById = RecordTypeMapper.getInstance().getModelByIdMapForSObjectType(Account.getSobjectType());
+        Test.stopTest(); 
+        
+        List<RecordTypeModel> recordTypeModels = new List<RecordTypeModel>();
+        for (RecordTypeModel recordTypeModel : recordTypeModelById.values) {
```
MatthewBlanski, line 48: This is a method and not a property.
```suggestion
        for (RecordTypeModel recordTypeModel : recordTypeModelById.values()) {
```

unpackaged/test/account_model_health_checks/pre/classes/RecordTypeMapper_Unmanaged_TEST.cls
```diff
@@ -0,0 +1,58 @@
...
+            if (recordTypeModel.developerName = 'An_Inactive_Record_Type') {
```
MatthewBlanski, line 49: This should be an equality operator
```suggestion
            if (recordTypeModel.developerName == 'An_Inactive_Record_Type') {
```

unpackaged/test/account_model_health_checks/pre/classes/RecordTypeMapper_Unmanaged_TEST.cls
```diff
@@ -0,0 +1,58 @@
...
+                recordTypeModels.add(recordTypeModel);
+            }
+        }
+
+        system.assertEquals(0, recordTypeModels.size());
```
MatthewBlanski, line 54: The desired state is that we find the inactive record type.
```suggestion
        system.assertEquals(1, recordTypeModels.size());
```

src/classes/RecordTypeService_TEST.cls
```diff
@@ -0,0 +1,146 @@
...
+
+        Test.startTest();
+            Map<Id, RecordTypeModel> recordTypeModelById = RecordTypeService.getInstance().getModelByIdMapForSObjectType(Account.getSobjectType());
+            RecordTypeModel recordTypeModel = RecordTypeService.getInstance().getRecordTypeModelFromMap('000000000000000000', recordTypeModelById);
+        Test.stopTest();
+
+        System.assertEquals('000000000000000000', recordTypeModel.Id);
+        System.assertEquals('Household', recordTypeModel.name);
+        System.assertEquals('Household Account', recordTypeModel.developerName);
+        System.assertEquals('hed', recordTypeModel.namespace);
```
MatthewBlanski, line 124: Is Active is in the spec.
```suggestion
        System.assertEquals(true, recordTypeModel.isActive);
```

src/classes/RecordTypeService_TEST.cls
```diff
@@ -0,0 +1,146 @@
...
+    * @description Test method to verify that a valid record type model is returned.
+    ***************************************************************************************************************************/
+    @isTest
+    private static void getRecordTypeModelFromMapValidIdInMapTest() {
+        RecordTypeMapper.instance = new STUB_RecordTypeMapper();
+
+        Test.startTest();
+            Map<Id, RecordTypeModel> recordTypeModelById = RecordTypeService.getInstance().getModelByIdMapForSObjectType(Account.getSobjectType());
+            RecordTypeModel recordTypeModel = RecordTypeService.getInstance().getRecordTypeModelFromMap('000000000000000000', recordTypeModelById);
+        Test.stopTest();
```
MatthewBlanski, line 119: We should be careful about including only the specific method we are testing inside of our test blocks so we can isolate those metrics from any test setup or asserts.
```suggestion
        Map<Id, RecordTypeModel> recordTypeModelById = RecordTypeService.getInstance().getModelByIdMapForSObjectType(Account.getSobjectType());
        
        Test.startTest();
            RecordTypeModel recordTypeModel = RecordTypeService.getInstance().getRecordTypeModelFromMap('000000000000000000', recordTypeModelById);
        Test.stopTest();
```

src/classes/RecordTypeService_TEST.cls
```diff
@@ -0,0 +1,146 @@
...
+    * @description Test method to verify that null is returned for invalid id
+    ***************************************************************************************************************************/
+    @isTest
+    private static void getRecordTypeModelFromMapInvalidIdTest() {
+        RecordTypeMapper.instance = new STUB_RecordTypeMapper();
+
+        Test.startTest();
+            Map<Id, RecordTypeModel> recordTypeModelById = RecordTypeService.getInstance().getModelByIdMapForSObjectType(Account.getSobjectType());
+            System.assertEquals(null, RecordTypeService.getInstance().getRecordTypeModelFromMap('invalid', recordTypeModelById));
+        Test.stopTest();
```
MatthewBlanski, line 106: We should be careful about including only the specific method we are testing inside of our test blocks so we can isolate those metrics from any test setup or asserts.
```suggestion
        Test.startTest();
            Map<Id, RecordTypeModel> recordTypeModelById = RecordTypeService.getInstance().getModelByIdMapForSObjectType(Account.getSobjectType());
        Test.stopTest();
        
        System.assertEquals(null, RecordTypeService.getInstance().getRecordTypeModelFromMap('invalid', recordTypeModelById));
```

MatthewBlanski (changes requested):
This captures the overall intent of the spec, but does have some necessary changes and requested changes

- The isActive portion of the spec was replaced with namespace for RecordTypeModels. This will need to be fixed.
- The unmanaged unit test class does not compile and is not added to the yml
- There is a generic exception catch in RecordTypeService.getRecordTypeModelFromMap that we do not have test coverage for and may be made from an overabundance of caution that could result in unexpected behavior not being caught by the test.
- Test blocks should be isolated from any asserts or test setup.
- Ideally we add messages explaining what each assert does.

In general, we want to avoid catching for Exception and focus on catching for specific exceptions. This allows us to detect unexpected behavior without having it fail silently. We should deviate from this in two cases as we have currently tacitly defined through prior practices: if we want to embed that exception into an aura handled exception for a more elegant display in the UI or if we want to embed it into a custom exception that retains the original information and allows us to more gracefully present the exception by appending relevant information.

In the case of the test blocks, those should be used to isolate the single methodcall that we perform, which allows it to operate on an entirely separate set of limits that we can use to extract metrics on the actual CPU time used to the actual number of SOQL queries used for that specific test and not its setup or asserts. This will be pivotal in getting metrics on performance in the future and in detecting spikes that might indicate that we have introduced unexpected performance-impacting changes.

MatthewBlanski (approved):

src/classes/RecordTypeModel.cls
```diff
@@ -0,0 +1,47 @@
...
+/**
+* @author Salesforce.org
+* @date 2020
+* @group
+* @group-content ../../ApexDocContent/HealthCheck.htm
+* @description Model class for Record Type
+*/
+public with sharing class RecordTypeModel {
+
+    public Id id;
```
m-kinkade, line 36: @MatthewBlanski do we need getters and setters for these variables?

MatthewBlanski: Yes, done.

PR #1466, opened 2020-12-23
Model, Mapper, Service for picklist entries

skosgi1:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/PicklistEntryMapper_TEST.cls
+++ b/src/classes/PicklistEntryMapper_TEST.cls
@@ -42,11 +42,11 @@ private class  PicklistEntryMapper_TEST {
     private static void getInstanceNew() {
 
         Test.startTest();
-            PicklistEntryMapper relationshipLookupMapperInstance = PicklistEntryMapper.getInstance();
+            PicklistEntryMapper picklistEntryMapperInstance = PicklistEntryMapper.getInstance();
         Test.stopTest();
 
-        System.assertEquals(PicklistEntryMapper.instance, relationshipLookupMapperInstance, 'Instance of mapper class returned should match static instance variable.');
-        System.assertEquals(true, relationshipLookupMapperInstance != null, 'Instance of mapper class should not be null.');
+        System.assertEquals(PicklistEntryMapper.instance, picklistEntryMapperInstance, 'Instance of mapper class returned should match static instance variable.');
+        System.assertEquals(true, picklistEntryMapperInstance != null, 'Instance of mapper class should not be null.');
     }
 
     /**************************************************************************************************************************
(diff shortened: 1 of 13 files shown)


src/classes/PicklistEntryMapper_TEST.cls
```diff
@@ -0,0 +1,95 @@
...
+    private static void getModelsByNameForSObjectField() {
+
+        String prefix = UTIL_Namespace.StrTokenNSPrefix('');
+        String objectName = prefix + 'Relationship__c';
+        String fieldName = prefix + 'Type__c';
+
+        Test.startTest();
+            Map<String, PicklistEntryModel> picklistEntryModelByPicklistEntryName = 
+                PicklistEntryMapper.getInstance().getModelsByNameForSObjectField(UTIL_Describe.getObjectDescribe(objectName).getSObjectType(), 
+                                                                                 UTIL_Describe.getFieldDescribe(objectName, fieldName).getSObjectField());
```
skosgi1, line 82: @MatthewBlanski Wanted to check with you if this is the right way to test? Thx.

src/classes/PicklistEntryMapper_TEST.cls
```diff
@@ -0,0 +1,95 @@
...
+
+    /**************************************************************************************************************************
+    * @description Test method to verify that the getModelsByNameForSObjectField method returns the picklist entry models
+    ***************************************************************************************************************************/
+    @isTest
+    private static void getModelsByNameForSObjectField() {
+
+        String prefix = UTIL_Namespace.StrTokenNSPrefix('');
+        String objectName = prefix + 'Relationship__c';
+        String fieldName = prefix + 'Type__c';
```
MatthewBlanski, line 77: We can use this method directly without plusses here. This should be our canonical way to use the existing tool to namespace-safe things.
```suggestion
        String objectName = UTIL_Namespace.StrTokenNSPrefix('Relationship__c');
        String fieldName = UTIL_Namespace.StrTokenNSPrefix('Type__c');
```

src/classes/PicklistEntryMapper_TEST.cls
```diff
@@ -0,0 +1,95 @@
...
+
+        Test.startTest();
+            Map<String, PicklistEntryModel> picklistEntryModelByPicklistEntryName = 
+                PicklistEntryMapper.getInstance().getModelsByNameForSObjectField(UTIL_Describe.getObjectDescribe(objectName).getSObjectType(), 
+                                                                                 UTIL_Describe.getFieldDescribe(objectName, fieldName).getSObjectField());
+        Test.stopTest();
+
+        List<Schema.PicklistEntry> picklistEntriesList = UTIL_Describe.getFieldPicklistEntries(prefix + 'Relationship__c', prefix + 'Type__c');
```
MatthewBlanski, line 85: Why not use the local variables you defined above for this? That way it will stay consistent.
```suggestion
        List<Schema.PicklistEntry> picklistEntriesList = UTIL_Describe.getFieldPicklistEntries(objectName,fieldName);
```

MatthewBlanski (changes requested):
A few minor changes for some reordering in the unit tests and practice with using EDA's internal tools. It's nitpick level. The actual code is perfect to spec.

MatthewBlanski (approved):

PR #1467, opened 2020-12-25
[EDA] Introducing prettier linting via github actions for code format…

jofsky:
…ting standardization

# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -1,0 +1,13 @@
+name: Formatting
+on: [push]
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: '14'
+      - name: prettier
+        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/prettier.yml
```diff
@@ -0,0 +1,13 @@
...
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: '14'
+      - name: prettier
+        run: |
+          npx prettier "**/*.{js,html,md,yml,yaml,json}" --check
```
MatthewBlanski, line 13: Given the scope of what Prettier is doing here and that JS touches our custom settings UI which is notably taciturn, I believe that our standard practice is to have that QA'd. Can we exclude js for this PR to make the matter more cut and dry? I'd like to talk to the rest of the team about prettier and the options we have there before we make that step into any of our managed code. I also want to make certain that we have the bandwidth to appropriately test those changes.

@davidmreed @jofsky

PR #1469, opened 2020-12-31
Reciprocal Relationship Settings Health Check Card

MatthewBlanski:
Added Reciprocal Relationships Health Check Card and mocks
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


src/labels/CustomLabels.labels
```diff
@@ -1087,6 +1087,30 @@
         <shortDescription>Field label for Accounts and Contacts Settings</shortDescription>
         <value>Disable Preferred Phone Enforcement</value>
     </labels>
+    <labels>
+        <fullName>stgHCPicklistValueInactive</fullName>
```
m-kinkade, line 1091: Do we need to include translation files for these labels?

m-kinkade (commented):
Mind blown.  This looks great.  I'm not sure if we plan to include translation files in a separate PR...if not, those will need to be included for the new custom labels :)

src/labels/CustomLabels.labels
MatthewBlanski: Good point. We really do.

PR #1473, opened 2021-01-01
[EDA] Default Account Model Health Check, Admin RT Check and Household RT check

m-kinkade:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/AccountModelHealthCheckVMapper.cls
+++ b/src/classes/AccountModelHealthCheckVMapper.cls
@@ -65,6 +65,24 @@ public virtual with sharing class AccountModelHealthCheckVMapper {
         AccountModelSettingsModel acctModSettingMod = this.locateAccountModelSettingsService().getAccountModelSettingsModel();
         Map<Id, RecordTypeModel> accountRTModelsById = this.locateRecordTypeService().getModelByIdMapForSObjectType(Account.SObjectType);
 
+        List<HealthCheckGroupVModel> accountHlthCheckGroupVModelsList = new List<HealthCheckGroupVModel>();
+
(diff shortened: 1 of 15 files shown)


src/classes/AccountModelHealthCheckVMapper.cls
```diff
@@ -65,55 +65,71 @@ public virtual with sharing class AccountModelHealthCheckVMapper {
...
+
+    /**********************************************************************************
+    * @description Generates a HealthCheckGroupVModel comprised of the combined health 
+    * check results for Default Account, Administrative and Household accounts. 
+    * @param  healthCheckGroupVModelsList A list of HealthCheckGroupVModels
+    * @return A HealthCheckGroupVModel representing the health check results for 
+    * Default, Administrative and Household Accounts.
+    **********************************************************************************/ 
+    @TestVisible
+    private HealthCheckGroupVModel combineAccountHealthCheckGroupViewModels(List<HealthCheckGroupVModel> healthCheckGroupVModelsList){
```
MatthewBlanski, line 85: This is an elegant solution that made the PR much more readable. It also isolated a unit test scenario.

MatthewBlanski (changes requested):
This is a good PR and all of the new code is functional. We currently are not getting double quotes for an empty or null record type, which was an oversight in the design for the model.

Please go into the Record Type Model and have the record types become "" when the string is blank. Feel free to use a ternary for this.

MatthewBlanski (approved):

PR #1484, opened 2021-01-05
Account Settings Health Check

MatthewBlanski:
# Critical Changes

# Changes
- Added a Health Check for the Default Account Model, Administrative Account Record Type, and the Household Account Record Type

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


cumulusci.yml
```diff
@@ -406,7 +417,7 @@ flows:
             6:
                 task: deploy_qa_config
             7:
-                task: deploy_health_check_config 
+                task: deploy_health_check_config
```
jofsky, line 420: Will this need to be added to K12?

jofsky (approved):
Looks good I'm assuming all the files in the `unpackaged/test_unpackaged/account_model_health_checks/ is intentionally not namespace tokenized to support the unmanaged test scenarios correct? @MatthewBlanski

cumulusci.yml
MatthewBlanski: No. This task will disappear before we go to a managed version. Right now it's a workaround to expose the top level health check component when working in an unmanaged context without having it exposed in our managed code.

src/classes/AccountModelHealthCheckVMapper.cls
```diff
@@ -0,0 +1,394 @@
...
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @group
```
xinnanlisf, line 30: minor but since we have fixed the other HC pr. the group here should be Health Check

src/classes/AccountModelHealthCheckVMapper.cls
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        return instance;
+    }
+
+    /**********************************************************************************
+    * @description Retrieve the results for a for an AccountModel health check.
+    * @return A HealthCheckGroupVModel representing a HealtCheckDescription for Account.
+    **********************************************************************************/ 
+    public virtual HealthCheckGroupVModel getHealthCheckGroup() {
```
xinnanlisf, line 64: The way how we implement this class is nice and clean. However, I am seeing some inconsistency among these feature vmapper classes. Should we consider to unify how we implement these classes to ensure readability? Also, I am sensing some duplicate code here, which can be a headache of maintenance if we need to make any change moving forward.

src/classes/AccountModelHealthCheckVMapper.cls
```diff
@@ -0,0 +1,394 @@
...
+        AccountModelSettingsModel acctModSettingMod = this.locateAccountModelSettingsService().getAccountModelSettingsModel();
+        Map<Id, RecordTypeModel> accountRTModelsById = this.locateRecordTypeService().getModelByIdMapForSObjectType(Account.SObjectType);
+
+        List<HealthCheckGroupVModel> accountHlthCheckGroupVModelsList = new List<HealthCheckGroupVModel>();
```
xinnanlisf, line 68: Not a big of fan of 'Hlth' but just my opinion. Also I feel like we correct this to use full words in some other PRs.

src/classes/AccountModelHealthCheckVMapper_TEST.cls
```diff
@@ -0,0 +1,1100 @@
...
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @group
```
xinnanlisf, line 30: Same here for adding 'Health Check' as group. I am assuming this would happen for most of the classes in this PR.

src/classes/AccountModelSettingsService.cls
```diff
@@ -0,0 +1,76 @@
...
+    * @description Empty constructor is not public for Singleton Pattern
+    **********************************************************************/
+    protected AccountModelSettingsService() {}
+
+    /*****************************************************************************
+    * @description Static method to get the current instance for Singleton pattern
+    * @return The instance of AccountModelSettingsService.
+    ******************************************************************************/
+    public static AccountModelSettingsService getInstance() {
+
```
xinnanlisf, line 52: Not a big deal but it seems that we leave an extra spacing for most of our getInstance() methods. Maybe this is a copy-paste thing?

src/classes/HealthCheckGroupController.cls
```diff
@@ -0,0 +1,59 @@
+/*
```
xinnanlisf, line 1: I am really curious. This class is in 230 already. Why this PR is trying to add it as a new class again?

MatthewBlanski: That's a good question. It looks like we had a hiccup with feature/230

MatthewBlanski: It looks like it's resolved now

xinnanlisf (approved):
Nicely done! I left a few comments but really just for general discussions and minor things. And I believe @m-kinkade  has quietly got them fixed :)

src/classes/AccountModelHealthCheckVMapper.cls
```diff
@@ -0,0 +1,394 @@
...
+
+    /**********************************************************************************
+    * @description Generates a HealthCheckGroupVModel comprised of the combined health 
+    * check results for Default Account, Administrative and Household accounts. 
+    * @param  healthCheckGroupVModelsList A list of HealthCheckGroupVModels
+    * @return A HealthCheckGroupVModel representing the health check results for 
+    * Default, Administrative and Household Accounts.
+    **********************************************************************************/ 
+    @TestVisible
+    private HealthCheckGroupVModel combineAccountHealthCheckGroupViewModels(List<HealthCheckGroupVModel> healthCheckGroupVModelsList) {
```
xinnanlisf, line 86: This method sounds like one we can extract and reused by all vmappers? maybe we can have a HealthCheckVMapper and let other vmappers to inherit this method?

src/classes/AccountModelHealthCheckVMapper.cls
```diff
@@ -0,0 +1,394 @@
...
+        if (instance == null) {
+            instance = new AccountModelHealthCheckVMapper();
+        }
+
+        return instance;
+    }
+
+    /**********************************************************************************
+    * @description Retrieve the results for a for an AccountModel health check.
+    * @return A HealthCheckGroupVModel representing a HealtCheckDescription for Account.
```
xinnanlisf, line 62: Shall we make the description more exact and specific. It is health check for account model not account. I am feeling this will cause a bit of confusion and misleading.

src/classes/AccountModelHealthCheckVMapper.cls
MatthewBlanski: Let's explore the level of effort on it. I'd like to see this used as it is a cleaner design than the one I applied in the Reciprocal Relationship Health Check. In an ideal world this would live in an abstract class that these classes depend on.

src/classes/AccountModelHealthCheckVMapper.cls
MatthewBlanski: Now that we have concrete code and some patterns, I think there is value in extracting it into abstract classes that these classes extend. Let's be certain we don't impact our release for this given the hard timelines we are on. If we can fit it in while compassionately handing it off to QA, then let's do so.

xinnanlisf (changes requested):
let's update the variable names to avoid abbreviations.

xinnanlisf:
> let's update the variable names to avoid abbreviations.

nvm let's fix these later.

PR #1497, opened 2021-01-09
[EDA] Affiliation Mapping Health Check Card AND [EDA] Affiliation Mappings Record Type Health Check

m-kinkade:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


MatthewBlanski (changes requested):
You are using the branch from Account Model as your base branch from this PR, please do not remove it as upon merge the Account Model work will be purged.

src/lwc/healthCheckRun/healthCheckRun.html
```diff
@@ -1,32 +1,5 @@
...
-                    </div>
-                    <div class="slds-col 
-                                slds-size_1-of-1 
-                                slds-text-title">
-                        {labelReference.stgHealthCheckTitle}
-                    </div>
-                </div>
-            </div>
-        </h3>
-
```
MatthewBlanski, line 29: Please do not remove this for now. We are commenting it out. See the PR for tab opening.

src/classes/UTIL_Describe.cls
```diff
@@ -1079,4 +1079,29 @@ public class UTIL_Describe {
...
+        return true;
+    }
+
+    /*******************************************************************************************************
+    * @description Retrieve a map of all fields for a specified SObject by field name.
+    * @param  sObjectName The name of an SObject type
+    * @return A Map<String, Schema.SObjectField> if the specified SObject name exists, null otherwise.
+    *******************************************************************************************************/
+    public static Map<String, Schema.SObjectField> getAllFieldsForSObject(String sObjectName){
+        return Schema.getGlobalDescribe().get(sObjectName)?.getDescribe().fields.getMap();
```
MatthewBlanski, line 1105: Let's use our UTIL_Describe functionality to its fullest here.
```suggestion
        UTIL_Describe.objectDescribes(sObjectName);
        return objectDescribesget(sObjectName).fields.getMap();
```

src/classes/LookupFieldMapper.cls
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**********************************************************************************************************
+    * @description Retrieve a map of LookupField models representing fields on a specified owner SObject type
+    * that is a lookup to a specified target SObject type by field label.
+    * @param  sObjectType  The SObject type where the field originates
+    * @param  lookupToSObjectType The SObject type where the field lookup targets
+    * @return Map<String, LookupFieldModel>
+    **********************************************************************************************************/ 
+    public virtual Map<String, LookupFieldModel> getModelByLabelMapForSObjectType(SObjectType sObjectType, SObjectType lookupToSObjectType) {
```
MatthewBlanski, line 90: I would prefer we abandon this method and swap it out for method in the service that effectively does the following:

- Arguments: Map<String,LookupFieldModel>
- Returns: Map<String,LookupFieldModel>
- Intialize a return map Map<String,LookupFieldModel>
- For each LookupFieldModel in the map's values
-- add the model to the return map by its label
-return the map to return

Ultimately I don't want us to hit the schema for field describes twice if we don't have to.

src/classes/LookupFieldMapper.cls
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**********************************************************************************************************
+    * @description Retrieve a map of LookupField models representing fields on a specified owner SObject type
+    * that is a lookup to a specified target SObject type by field name.
+    * @param  sObjectType  The SObject type where the field originates
+    * @param  lookupToSObjectType The SObject type where the field lookup targets
+    * @return Map<String, LookupFieldModel>
+    **********************************************************************************************************/ 
+    public virtual Map<String, LookupFieldModel> getModelByNameMapForSObjectType(SObjectType sObjectType, SObjectType lookupToSObjectType) {
```
MatthewBlanski, line 66: This method will work for our front end. We may want to refactor this to take advantage of the caches in UTIL_Describe at some point.

You may want to investigate using `UTIL_Describe.getFieldsOfType` for this to shorten the list down and iterate over it again to ensure caching.

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -0,0 +1,199 @@
...
+*/
+/**
+* @author Salesforce.org
+* @date 2021
+* @group Health Check
+* @group-content ../../ApexDocContent/HealthCheck.htm
+* @description Generate a HealthCheckGroup view model representing the results of
+* an Affiliation Mappings health check.
+*/
+public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
```
MatthewBlanski, line 35: Are we delaying implementing the tests for this until we have actual health checks? I don't see a test class for this.

MatthewBlanski (changes requested):
There are a few changes, mostly to do with not requiring schema and rerunning get describes.

src/classes/AffiliationMappingsHealthCheckVMapper.cls
m-kinkade: Using the name as is and adding _TEST exceeded the character limit for class names.  I had to shorten to: AfflMappingsHealthCheckVMapper.cls instead

MatthewBlanski: In cases like these, truncate at the end rather than the beginning so they are adjacent. Would you kindly make that change?

src/lwc/healthCheckRun/healthCheckRun.html
m-kinkade: This seems to be code that was duplicated from the healthCheckHighlightsPanel.html markup (which I may have neglected to remove after copy/past).  I'm not sure it belongs in the healthCheckRun component?

MatthewBlanski: Oh, very well. Carry on then!

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -0,0 +1,196 @@
...
+    public static AffiliationMappingsHealthCheckVMapper getInstance() {
+        if (instance == null) {
+            instance = new AffiliationMappingsHealthCheckVMapper();
+        }
+
+        return instance;
+    }
+
+    /**********************************************************************************
+    * @description Retrieve the results for a for an AccountModel health check.
```
xinnanlisf, line 60: Seems like a typo here.

MatthewBlanski (approved):
All concerns addressed. Functionality looks good and works!

xinnanlisf:
Shall we convert this from draft to ready for review?

MatthewBlanski (approved):

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -0,0 +1,304 @@
...
+            ),
+            'Pass',
+            Label.stgHealthCheckStatusPassed,
+            '',
+            '',
+            new List<HealthCheckItemVModel>()
+        );
+    }
+
+    /**********************************************************************************
```
xinnanlisf, line 199: @m-kinkade 
Should this description be more specific about valid record type?

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -0,0 +1,304 @@
...
+            parentHealthCheckGroupVModel.healthCheckItemList.addAll(childAffiliationMappingsHealthCheckResults.healthCheckItemList);
+        }
+
+        parentHealthCheckGroupVModel.updateHealthCheckStatus();
+        parentHealthCheckGroupVModel.healthCheckItemList.sort();
+
+        return parentHealthCheckGroupVModel;
+    }
+
+    /**********************************************************************************
```
xinnanlisf, line 95: Love this list! Although we will have to maintain it if we are adding more health checks.

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -0,0 +1,304 @@
...
+        if (instance == null) {
+            instance = new AffiliationMappingsHealthCheckVMapper();
+        }
+
+        return instance;
+    }
+
+    /**********************************************************************************
+    * @description Retrieve the results for a for an AffiliationMappings health check.
+    * @return A HealthCheckGroupVModel representing a HealtCheckDescription for Account.
```
xinnanlisf, line 61: @m-kinkade assuming this is a typo too?

src/classes/AffiliationMappingsHealthCheckVMapper.cls
m-kinkade: @xinnanlisf  I agree, I'm not sure if it's worth the maintenance, but I found it helpful for now.  I am happy to go with whatever the team decides is the best approach.

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -0,0 +1,305 @@
...
+            0,
+            new List<String>(),
+            new List<HealthCheckItemVModel>()
+        );
+
+        Map<Id, RecordTypeModel> accRTModelsById = this.locateRecordTypeService().getModelByIdMapForSObjectType(Account.SObjectType);
+        List<AffiliationMappingsModel> affiliationMappingsModelsList = this.locateAffiliationMappingsService().getAffiliationMappingsModels();
+
+        for (AffiliationMappingsModel affiliationMappingsModel : affiliationMappingsModelsList) {
+            HealthCheckGroupVModel childAffiliationMappingsHealthCheckResults = this.getAffiliationMappingsHealthCheckResults(
```
xinnanlisf, line 77: I am not entirely comfortable with this line. Do you need to add more parameters when adding new health checks? Wil that break our unit tests?

MatthewBlanski: This should retrieve the aggregate group for an individual affiliation mapping

m-kinkade: @xinnanlisf yes, I added the necessary parameters in the subsequent PRs for the additional health checks.  The unit tests will need to be updated accordingly.

xinnanlisf: yeah that seems like a pain when adding new health checks and a new parameter is required. We will have to update tons of lines in unit tests.

MatthewBlanski: Let's add the extra parameters in this PR and provide empty maps, etc to them for now. That way we don't have to worry about it as much downstream and we're in good shape for our unit tests.

MatthewBlanski (approved):
All tests pass! The code is clean, prerequisite arguments added for other PRs!

PR #1502, opened 2021-01-12
Update EDA Settings Reciprocal Relationship to use Partner and add a Partner row

MatthewBlanski:
# Critical Changes

# Changes
- We've updated the reciprocal relationship settings for Husband and Wife to use the existing picklist value Partner instead of Spouse. We've also added a new reciprocal setting for Partner. If you're starting with this version of EDA, these changes are included automatically. If you started with an earlier version, see [Keep Up with Relationship Updates](https://powerofus.force.com/s/article/EDA-Relationship).

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


skosgi1 (approved):
Changes look good and happy path testing passes.

spinkelman:
QA Passes - Will merge when item clears Merge Conflict @MatthewBlanski

PR #1506, opened 2021-01-12
student faculty record types exist, are active and are not equal

skosgi1:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


src/classes/CourseConSettingHealthCheckVMapper.cls
```diff
@@ -62,35 +62,59 @@ public virtual with sharing class CourseConSettingHealthCheckVMapper {
...
-
         CourseConnectionSettingsModel courseConnSettingsModel = this.locateCourseConnectionSettingsService().getCourseConnectionSettingsModel();
 
         String objectName = UTIL_Namespace.StrTokenNSPrefix('Course_Enrollment__c');
         SObjectType courseConnSObjectType = UTIL_Describe.getObjectDescribe(objectName).getSObjectType();
         Map<Id, RecordTypeModel> courseConnRTModelsById = this.locateRecordTypeService().getModelByIdMapForSObjectType(courseConnSObjectType);
 
-        HealthCheckGroupVModel courseConnEnableCheckResult = this.getCourseConnectionsNotEnabledHealthCheckItems(courseConnSettingsModel);
+        List<HealthCheckGroupVModel> courseConSettingtHlthCheckGroupVModelsList = new List<HealthCheckGroupVModel>();
+        courseConSettingtHlthCheckGroupVModelsList.add(this.getCourseConnectionsNotEnabledHealthCheckItems(courseConnSettingsModel));
```
MatthewBlanski, line 72: You'll need to do more with this specific item as it should exclude the following three if it is in a warning status. This is missing the requirement of skipping the next three checks if it comes back as a warning.

MatthewBlanski (changes requested):
This is missing one portion for the spec that needs to be resolved before pushing it.

MatthewBlanski (approved):
These changes check out.

PR #1526, opened 2021-01-20
Affiliation Mappings Health Checks: Contact Primary Field, Auto-Enrollment Status and Role

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -198,6 +257,325 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
...
+                new List<HealthCheckItemVModel>()
+            ));
+
+            healthCheckGroupVModel.totalChecks++;
+            healthCheckGroupVModel.updateHealthCheckStatus();
+
+            return healthCheckGroupVModel;
+        }
+
+        if (picklistEntryModelByName.keySet().contains(affiliationMappingsModel.autoProgramEnrollmentStatus) == false) {
```
MatthewBlanski, line 306: This should evaluate to the exact same situation. We don't need to perform an extra check here.
```suggestion
        if (picklistEntryModelByName.keySet().contains(affiliationMappingsModel.autoProgramEnrollmentStatus) == false) {
```

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -198,6 +257,325 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
...
+                new List<HealthCheckItemVModel>()
+            ));
+
+            healthCheckGroupVModel.totalChecks++;
+            healthCheckGroupVModel.updateHealthCheckStatus();
+
+            return healthCheckGroupVModel;
+        }
+
+        if (picklistEntryModelByName.keySet().contains(affiliationMappingsModel.autoProgramEnrollmentRole) == false) {
```
MatthewBlanski, line 428: We don't need to perform an extra check here.
```suggestion
        if (picklistEntryModelByName.keySet().contains(affiliationMappingsModel.autoProgramEnrollmentRole) == false) {
```

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -198,6 +257,325 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
...
+                    ),
+                    new List<HealthCheckItemVModel>()
+                ));
+
+                healthCheckGroupVModel.totalChecks++;
+                healthCheckGroupVModel.updateHealthCheckStatus();
+                
+                return healthCheckGroupVModel;
+            }
+        }
```
MatthewBlanski, line 554: We can isolate the if statement here and avoid nesting. This also can help readability as it becomes more clear that we are defaulting over to the other contact field model.

Another option that I prefer would be to make a helper method just like in the record type service where we get the lookupfieldmodel by the two maps. This might make code readability even better and isolate the get by name or label behavior within the service class. When we start settings work, we may rely on that method in the service class, so doing so will be defensive design.

```suggestion
        if (accountlookupFieldOnContactFieldModelFromName == null) {
            LookupFieldModel accountlookupFieldOnContactFieldModelFromLabel = this.locateLookupFieldService().getLookupFieldModelFromLabelMap(
                affiliationMappingsModel.primaryAffiliationField,
                accountlookupFieldOnContactFieldModelByLabel);
        }
            
        if (accountlookupFieldOnContactFieldModelFromLabel == null) {
            healthCheckGroupVModel.healthCheckItemList.add(new HealthCheckItemVModel(
                '03',
                Label.stgHCAfflMapContactPrimaryFieldTitle,
                'Fail',
                Label.stgHealthCheckStatusFailed,
                String.format(
                    Label.stgHCAfflMapContactPrimaryFieldInvalid, 
                    new List<Object>{affiliationMappingsModel.primaryAffiliationField}
                ),
                String.format(
                    Label.stgHCAfflMapContPrimaryFieldInvalidFix, 
                    new List<Object>{affiliationMappingsModel.primaryAffiliationField}
                ),
                new List<HealthCheckItemVModel>()
            ));

            healthCheckGroupVModel.totalChecks++;
            healthCheckGroupVModel.updateHealthCheckStatus();
            
            return healthCheckGroupVModel;
        }
```

src/classes/RecordTypeService.cls
```diff
@@ -58,27 +58,81 @@ public virtual with sharing class RecordTypeService {
...
         } catch (System.StringException ex) {
             return null;
         }
     }
 
+    /******************************************************************************
+    * @description Retrieve a record type model for a specified record type name or label.
+    * @param recordTypeNameOrLabel A string representing the name or label for a record type
+    * @param recordTypeModelByName A map of record type models by name
+    * @param recordTypeModelByName A map of record type models by label
```
MatthewBlanski, line 89:
```suggestion
    * @param recordTypeModelByLabel A map of record type models by label
```

MatthewBlanski (changes requested):
This is much improved. There are a few small things that could use tweaking, but overall this is in good shape.

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -101,32 +129,29 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
     * Contact Primary Affiliation Field Uniqueness
     * Auto-Enrollment Status
     * Auto-Enrollment Role
-    * @param  affiliationMappingsModel A model representing an affiliation mapping in EDA Settings
-    * @param  recordTypeModelsById A map of record type models by Id representing all Account record types specified in the org.
+    * @param affiliationMappingsModel A model representing an affiliation mapping in EDA Settings
+    * @param recordTypeModelsByName A map of record type models by developer name representing all Account record types specified in the org.
+    * @param recordTypeModelsByLabel A map of record type models by label name representing all Account record types specified in the org.
```
MatthewBlanski, line 134: These parameters are no longer passed

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -101,32 +129,29 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
...
-
-        RecordTypeModel recordTypeModel = 
-            this.locateRecordTypeService().getRecordTypeModelFromMap(
-                affiliationMappingsModel.accountRecordTypeId, 
-                recordTypeModelsById
-            );
-
+    private virtual HealthCheckGroupVModel getAffiliationMappingsHealthCheckResults(
+        AffiliationMappingsModel affiliationMappingsModel,
+        RecordTypeModel accountRecordTypeModel,
```
MatthewBlanski, line 143: This is a clear and descriptive variable name.

MatthewBlanski (approved):
Woot! Here we go!

PR #1532, opened 2021-01-22
Account Record Type is Mapped to More than One Contact Primary Affiliation Field

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/AffiliationMappingsHealthCheckVMapper.cls
+++ b/src/classes/AffiliationMappingsHealthCheckVMapper.cls
@@ -73,6 +73,8 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
             this.locateRecordTypeService().getRecordTypeModelByName(accRecordTypeModelsById);
         Map<String, RecordTypeModel> accountRecordTypeModelsByLabel = 
             this.locateRecordTypeService().getRecordTypeModelByLabel(accRecordTypeModelsById);
+        Set<String> duplicateAccountRecordTypeFieldValues = 
+            this.locateAffiliationMappingsService().getDuplicateAccountRecordTypeFieldValuesFromList(affiliationMappingsModelsList);
         Map<String, LookupFieldModel> accountLookupsOnContactFieldModelsByName = 
             this.locateLookupFieldService().getModelByNameMapForSObjectType(Contact.SObjectType, Account.SObjectType);
         Map<String, LookupFieldModel> accountLookupsOnContactFieldModelsByLabel =
@@ -104,7 +106,8 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
                 accountRecordTypeModel,
                 accountlookupFieldOnContactFieldModel,
                 affilationRolePicklistEntryModelByName,
-                affilationStatusPicklistEntryModelByName
+                affilationStatusPicklistEntryModelByName,
+                duplicateAccountRecordTypeFieldValues
             );
 
             // Add results from child health check group results to parent
@@ -143,7 +146,8 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
         RecordTypeModel accountRecordTypeModel,
         LookupFieldModel accountlookupFieldOnContactFieldModel,
         Map<String, PicklistEntryModel> affilationRolePicklistEntryModelByName,
-        Map<String, PicklistEntryModel> affilationStatusPicklistEntryModelByName
+        Map<String, PicklistEntryModel> affilationStatusPicklistEntryModelByName,
+        Set<String> duplicateAccountRecordTypeFieldValues
     ) {
         HealthCheckGroupVModel parentHealthCheckGroupVModel = new HealthCheckGroupVModel();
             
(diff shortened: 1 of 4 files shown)


xinnanlisf:
@MatthewBlanski I believe this is ready for review! Due to respect of merge conflicts, I will have the code in space once this is merged. Let me know if any feedback!

xinnanlisf:
> @MatthewBlanski I believe this is ready for review! Due to respect of merge conflicts, I will have the code in space once this is merged. Let me know if any feedback!

I mean place, not space.

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -577,6 +592,108 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
...
+    * @description Retrieves a HealthCheckGroup view model representing the health check
+    * results for Account Record Types specified.  Determines whether the Account 
+    * record type included in an affiliation mapping is unique.
+    * @param  affiliationMappingsModel A model representing an affiliation mapping in EDA Settings
+    * @param  accountRecordTypeModel A model representing the Account record type specified for an affiliation mapping
+    * @param  duplicateAccountRecordTypeFieldValues A list of duplicate accountRecordTypeFieldValues retrieved from Affiliation Mappings
+    * @return A HealthCheckGroupVModel
+    **********************************************************************************/ 
+    @TestVisible
+    private virtual HealthCheckGroupVModel getAccountRecordTypeUniquenessHealthCheckGroupVModel(
```
MatthewBlanski, line 606: @xinnanlisf This will cover duplicates for a not found scenario. However, it is not covering duplicates when the same record type is used but is referenced by API name and the other is referenced by label. In order to achieve this, I recommend having the service return a data structure that contains the combination of duplicate record type models and duplicate record type labels. You will need a means to compare record type models, which you should be able to do based on the record type model's id.

src/classes/AffiliationMappingsService.cls
```diff
@@ -64,6 +64,29 @@ public virtual with sharing class AffiliationMappingsService {
         return this.locateAffiliationMappingsMapper().getAffiliationMappingsModel();
     }
 
+    /***************************************************************************
+    * @description Retrieves a set of duplicate accountRecordTypeFieldValues from a list of Affiliation Mapping models
+    * @param affiliationMappingsModelList The affiliation mapping model list to
+    * retrieve ids from
+    * @return The set of duplicate accountRecordTypeFieldValues.
+    ***************************************************************************/
+    public Set<String> getDuplicateAccountRecordTypeFieldValuesFromList(
```
MatthewBlanski, line 73: This is a good way to ensure you do not loop through the same record type twice. Consider changing this method to handle record type model resolution based on the name and label maps and returning a structured data type with duplicate record type models and duplicate values in the mappings.

MatthewBlanski (changes requested):
There are some small holes in the way we are determining duplicates in this instance, introduced by the complexity that we support both names and labels.

src/classes/AffiliationMappingsHealthCheckVMapper.cls
xinnanlisf: @MatthewBlanski good catch! Hesitating to add more complexity to the data model, does this work? I am adding the method to the vmapper layer instead of the service so that we can utilize those two maps.
`private Set<Id> getDuplicateAccountRecordTypeIds(
        List<AffiliationMappingsModel> affiliationMappingsModelsList,
        Map<String, RecordTypeModel> accountRecordTypeModelsByName,
        Map<String, RecordTypeModel> accountRecordTypeModelsByLabel 
    ) {
        Set<Id> accountRecordTypeIds = new Set<Id>();
        Set<Id> duplicateRecordTypeIds = new Set<Id>();

        for (AffiliationMappingsModel affiliationMappingsModel : affiliationMappingsModelsList) {
            RecordTypeModel accountRecordTypeModel = 
                this.locateRecordTypeService().getRecordTypeModelFromNameAndLabelMaps(
                    recordTypeNameOrLabel, 
                    accountRecordTypeModelsByName, 
                    accountRecordTypeModelsByLabel
                );
            
            if (accountRecordTypeModel == null) {
                continue;
            } else {
                if (accountRecordTypeIds.contains(accountRecordTypeModel.Id)) {
                    duplicateRecordTypeIds.add(accountRecordTypeModel.Id);
                } else {
                    accountRecordTypeIds.add(accountRecordTypeModel.Id);
                }
            }
        }

        return duplicateRecordTypeIds;
    }`

And then we check if duplicateRecordTypeIds.contains(accountRecordTypeModel.Id)

MatthewBlanski: That looks good to me

src/classes/AffiliationMappingsService.cls
MatthewBlanski: Invalidated by previous suggestion

src/classes/AffiliationMappingsHealthCheckVMapper.cls
xinnanlisf: ok on it!

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -613,4 +732,42 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
...
+        List<AffiliationMappingsModel> affiliationMappingsModelsList,
+        Map<String, RecordTypeModel> accountRecordTypeModelsByName,
+        Map<String, RecordTypeModel> accountRecordTypeModelsByLabel 
+    ) {
+        Set<Id> accountRecordTypeIds = new Set<Id>();
+        Set<Id> duplicateRecordTypeIds = new Set<Id>();
+
+        for (AffiliationMappingsModel affiliationMappingsModel : affiliationMappingsModelsList) {
+            RecordTypeModel accountRecordTypeModel = 
+                this.locateRecordTypeService().getRecordTypeModelFromNameAndLabelMaps(
```
MatthewBlanski, line 754: Using this pre-existing method is a wonderfully elegant solution.

MatthewBlanski (approved):
This is functional and elegant!

PR #1535, opened 2021-01-23
Contact Primary Affiliation Field is Mapped to More than Account Record Type Labels

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/src/classes/AffiliationMappingsHCVMapper_TEST.cls
+++ b/src/classes/AffiliationMappingsHCVMapper_TEST.cls
@@ -1114,7 +1114,7 @@ public with sharing class AffiliationMappingsHCVMapper_TEST {
         AffiliationMappingsHCVMapper_TEST.validateHCGroupAssertions(
             healthCheckGroupVModelResult,
             Label.stgHealthCheckAffiliationMappings,
-            Label.stgHealthCheckStatusPassed,
+            Label.stgHealthCheckStatusFailed,
             0,
             1,
             0,
@@ -1181,7 +1181,7 @@ public with sharing class AffiliationMappingsHCVMapper_TEST {
         AffiliationMappingsHCVMapper_TEST.validateHCGroupAssertions(
             healthCheckGroupVModelResult,
             Label.stgHealthCheckAffiliationMappings,
-            Label.stgHealthCheckStatusPassed,
+            Label.stgHealthCheckStatusFailed,
             0,
             1,
             0,
@@ -2791,7 +2791,7 @@ public with sharing class AffiliationMappingsHCVMapper_TEST {
                         'test key 4', 
                         'test setting', 
                         'Fail', 
-                        Label.stgHealthCheckStatusPassed,
+                        Label.stgHealthCheckStatusFailed,
                         'test description', 
                         'test fix',
                         new List<HealthCheckItemVModel>()
--- a/src/classes/AffiliationMappingsHCVMapper_TEST.cls
+++ b/src/classes/AffiliationMappingsHCVMapper_TEST.cls
@@ -1170,7 +1170,7 @@ public with sharing class AffiliationMappingsHCVMapper_TEST {
                     'Valid Contact Lookup Field Name',
                     'Valid Contact Lookup Field Label'
                 ),
-                new Set<String>{'Valid Contact Lookup Field NameValid Contact Lookup Field Label'}
+                new Set<String>{'Valid Contact Lookup Field Name'}
             );
         Test.stopTest();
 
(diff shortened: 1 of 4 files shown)


xinnanlisf:
@MatthewBlanski the last piece of affiliation mappings puzzles. I tweaked the lookup model a bit to enable duplicate detection. Let me know if any feedback.

src/classes/LookupFieldModel.cls
```diff
@@ -34,9 +34,11 @@
 public with sharing class LookupFieldModel {
     public String name; //developer name
     public String label;
+    public String key;
 
     public LookupFieldModel(String name, String label) {
         this.name = name;
         this.label = label;
+        this.key = name + label;
```
MatthewBlanski, line 42: The name of a LookupFieldModel is guaranteed to be unique for a given sObjectField. Let's use that instead.

src/classes/AffiliationMappingsHealthCheckVMapper.cls
```diff
@@ -770,4 +865,42 @@ public virtual with sharing class AffiliationMappingsHealthCheckVMapper {
...
+
+    /**********************************************************************************
+    * @description Retrieves a set of duplicate lookup field model keys in affiliation mappings
+    * @param  affiliationMappingsModelsList A list of models representing affiliation mappings in EDA Settings
+    * @param  accountLookupsOnContactFieldModelsByName A map of lookup field model by name
+    * @param  accountLookupsOnContactFieldModelsByLabel A map of lookup field model by label
+    * @return A set of duplicate lookup field model keys
+    **********************************************************************************/
+    @TestVisible
+    private Set<String> getDuplicateLookupFieldModelKeys(
```
MatthewBlanski, line 877: You can preserve this method and use the name of the lookup field model. When lacking an Id, a name suffices as our ultimate goal is to retrieve a unique identifier for a lookup field model for a given sObjectField, which is Relationship Type in this case.

src/classes/AffiliationMappingsHCVMapper_TEST.cls
```diff
@@ -138,7 +138,7 @@ public with sharing class AffiliationMappingsHCVMapper_TEST {
         AffiliationMappingsHealthCheckVMapper AffiliationMappingsHealthCheckVMapper =
             AffiliationMappingsHealthCheckVMapper.getInstance();
 
-        Integer numberOfHealthChecks = 5; //TODO: Update when adding new Health Check
+        Integer numberOfHealthChecks = 6; //TODO: Update when adding new Health Check
```
MatthewBlanski, line 141: Let's change this over to a note.
```suggestion
        Integer numberOfHealthChecks = 6; //NOTE: Update when adding new Health Check
```

src/classes/AffiliationMappingsHCVMapper_TEST.cls
```diff
@@ -194,7 +194,7 @@ public with sharing class AffiliationMappingsHCVMapper_TEST {
         AffiliationMappingsHealthCheckVMapper AffiliationMappingsHealthCheckVMapper =
             AffiliationMappingsHealthCheckVMapper.getInstance();
 
-        Integer numberOfHealthChecks = 5; //TODO: Update when adding new health checks
+        Integer numberOfHealthChecks = 6; //TODO: Update when adding new health checks
```
MatthewBlanski, line 197: Likewise here on changing this to a note.
```suggestion
        Integer numberOfHealthChecks = 6; //NOTE: Update when adding new health checks
```

MatthewBlanski (changes requested):
A few minor changes to avoid some extra complexity that I don't believe we need to introduce given the final state of the code and one cleanup task prior to final commit. This is clean and elegant otherwise.

xinnanlisf:
@MatthewBlanski thanks for the feedback. The PR should be ready again!

src/classes/LookupFieldModel_TEST.cls
```diff
@@ -48,6 +48,6 @@ public with sharing class LookupFieldModel_TEST {
         Test.stopTest();
 
         System.assertEquals('TestName', lookupFieldModel.name, 'Name should be set.');
-        System.assertEquals('Test Label', lookupFieldModel.label, 'Label should be set.');
+		System.assertEquals('Test Label', lookupFieldModel.label, 'Label should be set.');
```
MatthewBlanski, line 51: Spacing fix.
```suggestion
        System.assertEquals('Test Label', lookupFieldModel.label, 'Label should be set.');
```

MatthewBlanski (approved):

PR #1548, opened 2021-02-03
QA Org 2gp runs install scripts

MatthewBlanski:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes
Run qa_org_2gp against a prerelease org shape

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -331,6 +331,14 @@ flows:
                 task: update_admin_profile
             3:
                 task: execute_install_apex
+            4:
+                task: deploy_dev_config
+            5:
+                task: execute_qa_apex
+            6:
+                task: deploy_qa_config
+            7:
+                flow: eda_settings_unmanaged
 
     config_apextest:
         # description: Configure an org to run apex tests after package metadata is deployed.
@@ -507,12 +515,6 @@ flows:
         steps:
             2:
                 flow: config_2gp
-            3:
-                task: None
-            4:
-                flow: config_qa
-            5:
-                task: snapshot_changes
 
     upgraded_org:
         description: Simulates an existing EDA customer org that has been push-upgraded from the latest production release of EDA to the current beta, including any dependencies. Deploys the unmanaged metadata from the production release.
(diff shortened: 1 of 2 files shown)


cumulusci.yml
```diff
@@ -499,6 +503,17 @@ flows:
                 options:
                     namespaced: False
 
+    qa_org_2gp:
+        steps:
+            2:
+                flow: config_2gp
+            3:
+                task: None
```
skosgi1, line 511: @MatthewBlanski since we are overwriting step 2 with a new task, can we also overwrite step 3 with config_qa instead of None?

cumulusci.yml
```diff
@@ -499,6 +503,17 @@ flows:
...
+            4:
+                flow: config_qa
```
skosgi1, line 513: @MatthewBlanski do we need config_qa again here since deploy_post and update_admin_profile are part of  config_2gp?

skosgi1: Ah nvm, I see coonfig_qa has other stuff too.

skosgi1: Can we move, step 3 to 7 in config_qa to config_2gp, so that we don't have to run deploy_post and update_admin_profile twice(once in each flow)?

skosgi1 (changes requested):
@MatthewBlanski one question with respect to running tasks twice. Everything else looks good!

xinnanlisf:
It looks good. I second Samsheeth's suggestion on avoiding running duplicate flows/tasks.

cumulusci.yml
MatthewBlanski: My thoughts on this were to not replicate a second flow for this specific scenario and let us continue on config_qa. The bug with the installer will be fixed soon so having a double non-destructive run for the time being felt like the lowest LOE. If we want to go that route we can, though it will need some additional cleanup.

skosgi1 (approved):
Looks great, thanks for making the changes!

davidmreed:
@MatthewBlanski  You may or may not wish to hold this PR. We are releasing a fix to CumulusCI today that will have 2GP packages run install scripts correctly.

xinnanlisf:
DO NOT merge until further notice.

spinkelman:
QA Testing Passes - Will merge on Monday

PR #1583, opened 2021-02-24
Internal Execute Install Script Fix

MatthewBlanski:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

- This fixes an issue with namespacing in dev org, 2gp installer, and execute install apex needing a persistent unmanaged install context to use until the batch jobs complete.


xinnanlisf (approved):
the fix looks good. However, it looks like we are getting rid of config_2gp. Is it on purpose?

MatthewBlanski:
@Xinnan Li <xinnan.li@salesforce.com> that is accurate. We no longer need
config_2gp because the default config_2gp flow and config_qa now serves our
needs.
Matthew Blanski
Lead Member of Technical Staff
Salesforce - Salesforce.org
C: 773-326-7278

cumulusci.yml
```diff
@@ -91,6 +91,14 @@ tasks:
             path: unpackaged/config/spe
             namespace_inject: $project_config.project__package__namespace
 
+    deploy_install_emulator:
+        group: "EDA: Metadata"
+        description: Deploys "Manage Encryption Keys" permission set.
```
davidmreed, line 96: This description came from a different task.

cumulusci.yml
```diff
@@ -91,6 +91,14 @@ tasks:
...
+        class_path: cumulusci.tasks.salesforce.Deploy
+        options:
+            path: unpackaged/config/install_emulator
+            namespace_inject: $project_config.project__package__namespace
```
davidmreed, line 100: This option is no longer required - CumulusCI can infer it for you now.

cumulusci.yml
```diff
@@ -478,48 +474,14 @@ flows:
         steps:
             1:
                 task: execute_enable_record_type_validation
-                options:
-                    namespaced: True
```
davidmreed, line 482: 🎉

unpackaged/config/dev/package.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Package xmlns="http://soap.sforce.com/2006/04/metadata">
     <types>
-        <members>*</members>
+        <members>%%%NAMESPACE%%%HEDA</members>
```
davidmreed, line 4: Is this CustomApp identical to the one we package in terms of tabset, currently? I would expect a deployment to fail in a managed context if the tabset differs. Are you seeing different behavior?

cumulusci.yml
MatthewBlanski: This seems to be fixed from a merge!

unpackaged/config/dev/package.xml
MatthewBlanski: It is. Also we do not see a deployment failure. App tabs can be appended to. I haven't tested for what occurs when we try to remove tabs, but I did want to change this to be explicit. We've been running this for a long time in the [QA config.](https://github.com/SalesforceFoundation/EDA/blob/main/unpackaged/config/qa/package.xml)

The only reason I am not recommending we make a different app when performing dev/qa work is for ease of lqa and the objects being in "one place"

davidmreed: Sounds good. I'll go ahead and approve, just calling it out as a potential risk if there are future modifications. (Fairly sure it will fail if tabs are removed).

davidmreed (approved):

PR #1609, opened 2021-03-10
[EDA] Settings Navigation - Unstyled

MatthewBlanski:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/lwc/edaSettingsNavigationSection/edaSettingsNavigationSection.html
```diff
@@ -0,0 +1,16 @@
+<template>
+    <div class="slds-nav-vertical__section">
```
MatthewBlanski, line 2: Calling out that normally we would use a vertical navigation component. Instead, we are appending to the design pattern used for slds vertical navigation alongside our UI innovation on settings, which will need to treat individual menu items as one comprehensive whole. We may reverse course before the end of the release, but I wanted to call out why we are not using that base component.

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.html
```diff
@@ -0,0 +1,12 @@
+<template>
+    <div class="slds-grid slds-wrap">
+        <nav class="slds-nav-vertical slds-nav-vertical_compact">
+            <template for:each={edaNavigationViewModel.navigationSections} for:item="navigationSection">
+                <c-eda-settings-navigation-section
```
MatthewBlanski, line 5: I ended up having to break this out in order to handle the active state of a page from a data-driven perspective.

force-app/main/default/lwc/edaSettings/edaSettings.html
```diff
@@ -1,11 +1,11 @@
 <template>
     <template if:true={currentUserHasAccess}>
         <div class="slds-grid slds-theme_default eda-height_full">
-            <div class="slds-col slds-no-flex slds-size_4-of-12 slds-scrollable_y eda-height_full">
-                {labelReference.settingsNavigation}
+            <div class="slds-col slds-p-top_large slds-no-flex slds-size_4-of-12 slds-scrollable_y eda-height_full">
```
xinnanlisf, line 4: the paddling looks alright but should we consider having a consistent way to have paddings and margins aligned? Also, I am curious about eda-height_full. Do we really need a custom css?

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
```diff
@@ -0,0 +1,39 @@
...
+            },
+        ],
+    };
+
+    @api setActivePage(pageName) {
+        this.settingsNavigationViewModel.navigationSections.forEach((navigationSection) => {
+            if (navigationSection.pageName === pageName) {
+                navigationSection.isActive = true;
+            } else {
+                navigationSection.isActive = undefined;
```
xinnanlisf, line 35: Why are we setting isActive as undefined instead of false?

xinnanlisf (commented):
Still in progress going through the front end mainly. However I would like to submit the comments for ongoing discussion purpose.

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
```diff
@@ -0,0 +1,39 @@
...
+    labelReference = {
+        accountModelSettingsTitle,
+    };
+
+    settingsNavigationViewModel = {
+        navigationSections: [
+            {
+                label: this.labelReference.accountModelSettingsTitle,
+                page: "accountmodelsettings",
+                sectionkey: "accountmodelsettings",
```
xinnanlisf, line 14: very minor thing but sectionKey?

xinnanlisf (approved):
Overall it looks good except some minor variable name thing and general questions about component styling. I am going to approve this thou.

force-app/main/default/lwc/edaSettings/edaSettings.html
MatthewBlanski: For the padding, that's something we'll want to look at once we have a 'full' UI and establish where the padding does and doesn't go relative to what we have displayed. My understanding is that's usually done at a 'fit and finish' stage once we have the components together.

Unfortunately, we do need the 100% height for the time being, though if you can find an slds class that is supported and documented that handles the height, that would be the superior solution.

force-app/main/default/lwc/edaSettings/edaSettings.html
```diff
@@ -1,11 +1,11 @@
...
+                <c-eda-settings-navigation onsettingsnavigation={handleSettingsNavigation}></c-eda-settings-navigation>
             </div>
             <div class="slds-col slds-no-flex slds-size_8-of-12 slds-scrollable_y eda-height_full">
-                <template if:true={settingsPageToDisplay.accountModelSettings}>
+                <template if:true={settingsPageToDisplay.accountmodelsettings}>
```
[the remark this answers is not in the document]
MatthewBlanski: Largely to obviate case sensitivity issues. This is controlled by a map that uses undefined or true to take advantage of truthiness. Note that template if:true handles that well. My understanding is undefined/true is a superior pattern to true/false in these cases.

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
MatthewBlanski: Not all navigation sections are navigable. Key can be a keyword in some instances, so sectionkey allows us to have a unique identifier on it.

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
MatthewBlanski: This is following truthy/falsy. More importantly, this is a format of defining variables that can be important when dealing with some attributes in lwc, such as disabled on input affordances. In a 'fun' experience with said attributes:

- When a disabled attribute is undefined, the component is considered to not be disabled.
- When disabled === false, the component is disabled. This was the headscratcher moment for me.
- When disabled === true, the component is disabled.
- Thus, when disabled is defined, the component is considered disabled. Fun with Javascript!

force-app/main/default/lwc/edaSettings/edaSettings.html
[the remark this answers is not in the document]
xinnanlisf: I do understand js is case sensitive but accountModelSettings is still a property name and should follow the code style?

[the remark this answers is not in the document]
MatthewBlanski: Good point. Changed.

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
MatthewBlanski: Changed to id in the full nav.

xinnanlisf:
Closing this one since we are working on styled PR.

PR #1611, opened 2021-03-10
[EDA] People & Groups: Account Model Settings Frontend

m-kinkade:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/classes/AccountModelSettingsController.cls
```diff
@@ -0,0 +1,58 @@
...
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+/**
+ * @author Salesforce.org
+ * @date 2021
+ * @group Health Check
```
MatthewBlanski, line 30:
```suggestion
 * @group EDA Settings
```

force-app/main/default/classes/AccountModelSettingsController_TEST.cls
```diff
@@ -0,0 +1,146 @@
...
+            );
+        }
+    }
+
+    /**************************************************************************************************************************************
+     * @description Stub class to simulate the response from AccountModelSettingsVMapper to return a valid AccountModelSettingsVModel
+     **************************************************************************************************************************************/
+    private class STUB_ExceptionAccountModelSettingsVMapper extends AccountModelSettingsVMapper {
+        public override AccountModelSettingsVModel getAccountModelSettingsViewModel() {
+            throw new TestException('Test exception!');
```
MatthewBlanski, line 140: You can always use a standard exception like a NoAccessException, but that does run the risk of us catching it later. This may be a better solution than doing so.

This is a perfectly-functional piece of code and does not require changes.

force-app/main/default/classes/AccountModelSettingsVMapper.cls
```diff
@@ -58,27 +58,81 @@ public virtual with sharing class AccountModelSettingsVMapper {
...
+
+        Map<Id, RecordTypeModel> activeAccountRecordTypeModelsById = this.locateRecordTypeService()
+            .getModelByIdMapForSObjectType(Account.SObjectType);
+
+        // filter out inactive record types
+        for (Id accountRecordTypeModelId : activeAccountRecordTypeModelsById.keySet()) {
+            if (activeAccountRecordTypeModelsById.get(accountRecordTypeModelId).isActive == false) {
+                activeAccountRecordTypeModelsById.remove(accountRecordTypeModelId);
+            }
+        }
```
MatthewBlanski, line 80: As we're going to be performing this in multiple places, this is a good candidate to move to the RecordTypeService, which should instantiate a new map and use map.put() only on active record type models.

force-app/main/default/classes/AccountModelSettingsVMapper.cls
```diff
@@ -58,27 +58,81 @@ public virtual with sharing class AccountModelSettingsVMapper {
...
+
+        List<ComboboxValueVModel> activeRecordTypeComboboxValueVModelsList = new List<ComboboxValueVModel>();
+
+        // populate available combobox options with active Account record types
+        for (RecordTypeModel activeAccountRecordTypeModel : activeAccountRecordTypeModelsById.values()) {
+            activeRecordTypeComboboxValueVModelsList.add(
+                new ComboboxValueVModel(activeAccountRecordTypeModel.name, activeAccountRecordTypeModel.id)
+            );
+        }
```
MatthewBlanski, line 89: This is a good candidate to be split out into its own method so we can confirm that view model conversion is as expected.

force-app/main/default/classes/AccountModelSettingsVMapper_TEST.cls
```diff
@@ -138,10 +135,158 @@ public with sharing class AccountModelSettingsVMapper_TEST {
...
+     * value set to the current field value specified in Hierarchy Settings for Account Record Types and a list of comboboxVModels that
+     * correspond to all active Account Record Types in the org.
+     ***********************************************************************************************************************************/
+    @isTest
+    private static void getAccountModelSettingsViewModelValid() {
+        RecordTypeMapper.instance = new STUB_RecordTypeMapper();
+
+        Hierarchy_Settings__c hierarchySettings = UTIL_CustomSettings_API.getSettingsForTests(
+            new Hierarchy_Settings__c(
+                Account_Processor__c = '000000000000000001',
```
MatthewBlanski, line 155: That's a clever way to use hybrid settings to get everything you need out of one mapper.

force-app/main/default/classes/AccountModelSettingsVMapper_TEST.cls
```diff
@@ -138,10 +135,158 @@ public with sharing class AccountModelSettingsVMapper_TEST {
...
+    }
+
+    /***********************************************************************************************************************************
+     * @description Test method to verify that the getAccountModelSettingsViewModel returns an AccountModelSettingsVModel with the selected
+     * value set to the value specified in the associated Hierarchy Settings fields and an empty list of comboboxVModels when there are
+     * no active Account Record Types in the org.
+     ***********************************************************************************************************************************/
+    @isTest
+    private static void getAccountModelSettingsViewModelWithNoRecordTypes() {
+        // TODO: what happens if there are zero Account record types in the org?
```
MatthewBlanski, line 204: Change this to EDA being wholly broken without record types in the org, therefore consideration of that scenario is unnecessary for unit tests.
```suggestion
        // TODO: what happens if there are zero Account record types in the org?
```

force-app/main/default/classes/AccountModelSettingsVModel_TEST.cls
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * @author Salesforce.org
+ * @date 2021
+ * @group
+ * @group-content ../../ApexDocContent/Tests.htm
+ * @description Test class for AccountModelSettingsVModel
+ */
+@isTest
+private with sharing class AccountModelSettingsVModel_TEST {
+    private static Integer loopCounter = 5;
```
MatthewBlanski, line 36: Thank you for using the loop Counter, this will help immensely in the future if we want to do bulk testing overrides.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -7,12 +7,81 @@
...
             >
-                <!-- Body of Account Model Settings page (populates Slot in child component) -->
-                <lightning-button
-                    label="Test Button"
-                    disabled={affordancesDisabled}
-                ></lightning-button>
             </c-settings-save-canvas>
+
+            <!-- Body of Account Model Settings page (populates Slot in child component) -->
+            <template if:true={accountModelSettingsViewModel}>
```
MatthewBlanski, line 13: This is a good solution given how we are populating this variable. Since a lightning spinner for this is not in the spec, that is a valid concern. Depending on page load times we may still need to incorporate one. We can save that as a nice to have.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -7,12 +7,81 @@
...
+                        </div>
+                        <div class="slds-text-body_small slds-text-color_weak slds-p-horizontal_large">
+                            {labelReference.stgHelpAccountModel}
+                        </div>
+                    </div>
+
+                    <div class="slds-col slds-size_1-of-3 slds-col_bump-right slds-p-right_x-large">
+                        <lightning-combobox
+                            name="defaultAccountModel"
+                            label="Default Account Record Type"
```
MatthewBlanski, line 28: This label is hardcoded and should be changed to use {labelReference.stgAccModelTitle}

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -7,12 +7,81 @@
...
+                            value={accountModelSettingsViewModel.defaultAccountRecordType.value}
+                            placeholder=""
```
MatthewBlanski, line 30: This wasn't directly in the spec, but we need to use the Select... as the placeholder for these. It came in from docs post-spec

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -7,12 +7,81 @@
...
+                            {labelReference.stgAccModelTitle}
+                        </div>
+                        <div class="slds-text-body_small slds-text-color_weak slds-p-horizontal_large">
+                            {labelReference.stgHelpAccountModel}
+                        </div>
+                    </div>
+
+                    <div class="slds-col slds-size_1-of-3 slds-col_bump-right slds-p-right_x-large">
+                        <lightning-combobox
+                            name="defaultAccountModel"
```
MatthewBlanski, line 27: Keep these names in mind for navigation purposes. Also consider adding a data-qa-locator attribute by the same name so that QA can quickly navigate with robot. Again, this is a new item that came up from a discussion on Friday, so you could not have known about it beforehand.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -7,12 +7,81 @@
...
                 onsettingssavecompleted={handleSettingsSaveCompleted}
             >
-                <!-- Body of Account Model Settings page (populates Slot in child component) -->
-                <lightning-button
-                    label="Test Button"
-                    disabled={affordancesDisabled}
-                ></lightning-button>
             </c-settings-save-canvas>
+
+            <!-- Body of Account Model Settings page (populates Slot in child component) -->
```
MatthewBlanski, line 12: Note we're not actually using the slot here! We should.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -7,12 +7,81 @@
...
+            <template if:true={accountModelSettingsViewModel}>
+                <div class="slds-grid slds-grid_vertical-align-start slds-wrap">
+                    <!-- Default Account Record Type Setting -->
+                    <div class="slds-col slds-size_2-of-3 slds-p-bottom_x-large">
```
MatthewBlanski, line 16: This is a good choice of sizing and styling. Using a vertical align for this was a great choice as well. We may want to consider changing the saving component to define the grid to follow this format instead.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -7,12 +7,81 @@
...
+                        </div>
+                        <div class="slds-text-body_small slds-text-color_weak slds-p-horizontal_large">
+                            {labelReference.stgHelpAdminRecType}
+                        </div>
+                    </div>
+
+                    <div class="slds-col slds-size_1-of-3 slds-col_bump-right slds-p-right_x-large">
+                        <lightning-combobox
+                            name="adminAccountModel"
+                            label="Administrative Account Record Type"
```
MatthewBlanski, line 51: Same here for the label. We could also consider using a labeled by for this, indicating the heading is the label.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -7,12 +7,81 @@
...
+                            value={accountModelSettingsViewModel.adminAccountRecordType.value}
+                            placeholder=""
```
MatthewBlanski, line 53: Likewise for the placeholder

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -7,12 +7,81 @@
...
+                    <div class="slds-col slds-size_1-of-3 slds-col_bump-right slds-p-right_x-large">
+                        <lightning-combobox
+                            name="defaultAccountModel"
+                            label="Default Account Record Type"
+                            value={accountModelSettingsViewModel.defaultAccountRecordType.value}
+                            placeholder=""
+                            options={accountModelSettingsViewModel.defaultAccountRecordType.options}
+                            onchange={handleDefaultAccountModelChange}
+                            disabled={affordancesDisabled}
+                            variant="label-hidden"
```
MatthewBlanski, line 34: Good use of variants. TIL.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.js
```diff
@@ -1,9 +1,20 @@
...
+import getAccountModelSettingsViewModel from "@salesforce/apex/AccountModelSettingsController.getAccountModelSettingsViewModel";
+
 import stgAccountModelSettingsTitle from "@salesforce/label/c.stgAccountModelSettingsTitle";
+import stgAccModelTitle from "@salesforce/label/c.stgAccModelTitle";
+import stgHelpAccountModel from "@salesforce/label/c.stgHelpAccountModel";
+import stgAdminAccountRecordType from "@salesforce/label/c.stgAdminAccountRecordType";
+import stgHelpAdminRecType from "@salesforce/label/c.stgHelpAdminRecType";
+import stgAccountRecordTypeSupportsHHAddress from "@salesforce/label/c.stgAccountRecordTypeSupportsHHAddress";
+import stgHelpHouseholdRecType from "@salesforce/label/c.stgHelpHouseholdRecType";
+import AfflMappingsDescription from "@salesforce/label/c.AfflMappingsDescription";
```
MatthewBlanski, line 12: We don't use this label anywhere
```suggestion
import AfflMappingsDescription from "@salesforce/label/c.AfflMappingsDescription";
```

force-app/main/default/lwc/accountModelSettings/accountModelSettings.js
```diff
@@ -14,8 +25,56 @@ export default class AccountModelSettings extends LightningElement {
 
     labelReference = {
```
MatthewBlanski, line 26: Now that we're in the 'real' aspect of this, let's consider placing all variables above properties and methods.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.js
```diff
@@ -14,8 +25,56 @@ export default class AccountModelSettings extends LightningElement {
...
+        stgAccountRecordTypeSupportsHHAddress,
+        stgHelpHouseholdRecType,
     };
 
+    @wire(getAccountModelSettingsViewModel)
+    accountModelSettingsViewModel({ error, data }) {
+        if (data) {
+            this.accountModelSettingsViewModel = data;
+        } else if (error) {
+            console.log("error retrieving accountmodelsettingsvmodel");
```
MatthewBlanski, line 41: Please comment this out for now. Error handling for display will need new mocks and a different pass.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.js
```diff
@@ -14,8 +25,56 @@ export default class AccountModelSettings extends LightningElement {
...
+        }
+    }
+
+    handleDefaultAccountModelChange(event) {
+        // add updated setting to hierarchySettingsChanges object
+        let hierarchySettingsChange = {
+            settingsType: "string",
+            settingsName: "Account_Processor__c".toLowerCase(),
```
MatthewBlanski, line 49: Do we need lower case here? Legitimate question, I'm not sure either way, but if we don't need to process this further, then let's not.

force-app/main/default/lwc/settingsSaveCanvas/settingsSaveCanvas.html
```diff
@@ -1,37 +1,33 @@
 <template>
-    <div class="slds-grid slds-wrap">
-        <div class="slds-col slds-text-heading_small">
-            <span>{componentTitle}</span>
-        </div>
+    <div class="slds-grid slds-grid_vertical-align-start">
```
MatthewBlanski, line 2: It's good to feel comfortable updating this component now that we're hitting real styling. We should use its slot!

force-app/main/default/lwc/settingsSaveCanvas/settingsSaveCanvas.html
```diff
@@ -1,37 +1,33 @@
...
+                ></lightning-button>
+                <lightning-button
+                    variant="brand"
+                    label={labelReference.settingsButtonSave}
+                    title={labelReference.settingsButtonSave}
+                    onclick={handleSaveClick}
+                    class="slds-p-right_x-small"
+                ></lightning-button>
+            </div>
+            <!-- </span> -->
```
MatthewBlanski, line 30:
```suggestion
```

MatthewBlanski (changes requested):
A few small changes requested. Overall this is an excellent foray into our first 'real' settings work. The save slot was not used and a few small comments and labels and 'new' requirements and patterns.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
m-kinkade: @MatthewBlanski  should this be included as a custom label?

MatthewBlanski: Yes. stgOptSelect

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -7,12 +7,81 @@
...
-                    disabled={affordancesDisabled}
-                ></lightning-button>
             </c-settings-save-canvas>
+
+            <!-- Body of Account Model Settings page (populates Slot in child component) -->
+            <template if:true={accountModelSettingsViewModel}>
+                <div class="slds-grid slds-grid_vertical-align-start slds-wrap">
+                    <!-- Default Account Record Type Setting -->
+                    <div class="slds-col slds-size_2-of-3 slds-p-bottom_x-large">
+                        <div class="slds-text-title slds-text-color_default slds-p-left_small">
```
MatthewBlanski, line 17: Let's bump the font size up on this a little. Let's try using slds-text-title_bold

force-app/main/default/lwc/accountModelSettings/accountModelSettings.js
```diff
@@ -1,20 +1,84 @@
...
+        defaultAccountModelTitle: stgAccModelTitle,
+        defaultAccountModelDescription: stgHelpAccountModel,
+        adminAccountModelTitle: stgAdminAccountRecordType,
+        adminAccountModelDescription: stgHelpAdminRecType,
+        hhAccountModelTitle: stgAccountRecordTypeSupportsHHAddress,
+        hhAccountModelDescription: stgHelpHouseholdRecType,
+        comboboxPlaceholderText: stgOptSelect,
+        defaultAccountModelComboboxId: "defaultAccountModel",
+        adminAccountModelComboboxId: "adminAccountModel",
+        hhAccountModelComboboxId: "hhAccountModel",
```
MatthewBlanski, line 31: Oooh, elegant. These aren't technically labels. Maybe move these to another data structure? This is a great way to make it obvious that these are important.

MatthewBlanski (changes requested):
One nitpick. Let's move those ids into another data structure that isn't label reference that is explicitly for capturing the navigation id and reference ids for each input. Maybe navigationReference affordanceIdReference? @xinnanlisf can probably help with naming here too. I'm not happy with either of the suggestions I've just made.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.js
m-kinkade: How about inputAttributeReference?  @xinnanlisf what are your thoughts?

force-app/main/default/lwc/settingsSaveCanvas/settingsSaveCanvas.html
```diff
@@ -1,37 +1,33 @@
 <template>
-    <div class="slds-grid slds-wrap">
-        <div class="slds-col slds-text-heading_small">
-            <span>{componentTitle}</span>
+    <div class="slds-grid slds-grid_vertical-align-start">
+        <div class="slds-col slds-size--1-of-2 slds-text-title slds-text-title_bold slds-p-bottom_xx-large">
```
MatthewBlanski, line 3: I've been thinking about this styling. Let's try using `slds-text-heading_small` on this one so it stands out from the others. It took me a while to determine exactly where we wanted this to sit, but this is closer to the styling. We should be able to use this moving forward.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.js
```diff
@@ -8,30 +8,29 @@ export default class AccountModelSettings extends LightningElement {
...
+
     handleSettingsSaveCompleted(event) {
         this.affordancesDisabledToggle = false;
     }
 
-    refreshAllApex() {
-    }
-}
\ No newline at end of file
+    refreshAllApex() {}
```
MatthewBlanski: This is why we're running slow! We'll need to cover this in the next WI. We need this to refresh apex so we get the 'current' settings.

MatthewBlanski (approved):

force-app/main/default/lwc/accountModelSettings/accountModelSettings.js
```diff
@@ -1,20 +1,87 @@
...
 
-    labelReference = {
-        stgAccountModelSettingsTitle,
-    };
+    @wire(getAccountModelSettingsViewModel)
+    accountModelSettingsViewModel({ error, data }) {
+        if (data) {
+            this.accountModelSettingsViewModel = data;
+        } else if (error) {
+            //console.log("error retrieving accountmodelsettingsvmodel");
```
xinnanlisf, line 49: Saw this while working on patterns. Is this on purpose? @m-kinkade

m-kinkade: @xinnanlisf yes, for now.  I'm not sure how we want to handle an error in this scenario so it will be addressed in a future iteration.

force-app/main/default/classes/HierarchySettingsService.cls
```diff
@@ -70,11 +70,20 @@ public virtual with sharing class HierarchySettingsService {
...
+
         try {
             // Update Hierarchy Settings fields with single values
             for (
                 String changedSettingName : hierarchySettingsChangesVModel.settingsSingleValueBySettingsName.keySet()
             ) {
+                //changedSettingName = UTIL_Namespace.StrTokenNSPrefix(changedSettingName);
+
                 Schema.DescribeFieldResult fieldDescribeResult = this.locateHierarchySettingsMapper()
                     .getHierarchySettingsField(changedSettingName);
```
MatthewBlanski, line 88: We can leverage UTIL_Namespace at this juncture without adding a large layer of complexity and using it for the field setting below. Pre-conversion can be a good option, but let's handle namespacing a bit more locally. You would have to use settingsField for the actual put operations below, and replicate that local variable setting, but I feel like the performance difference between those checks is minimal at best with the benefit of reducing complexity.

```suggestion
        try {
            // Update Hierarchy Settings fields with single values
            for (
                String changedSettingName : hierarchySettingsChangesVModel.settingsSingleValueBySettingsName.keySet()
            ) {
                String settingsField = UTIL_Namespace.StrTokenNSPrefix(changedSettingName);

                Schema.DescribeFieldResult fieldDescribeResult = this.locateHierarchySettingsMapper()
                    .getHierarchySettingsField(settingsField);
```

force-app/main/default/classes/HierarchySettingsService_TEST.cls
```diff
@@ -187,14 +187,11 @@ public with sharing class HierarchySettingsService_TEST {
         );
 
         Map<String, String> singleValueSettingUpdatesByNameInvalid = new Map<String, String>{
-            UTIL_Namespace.StrTokenNSPrefix('Account_Processor_Invalid__c') => 'Updated Value'
+            'Account_Processor_Invalid__c' => 'Updated Value'
```
MatthewBlanski, line 190: The side effect is using the namespace token prefix in our tests. This is a tradeoff I feel is relevant at this stage, but worth the effort.

force-app/main/default/lwc/settingsSaveCanvas/settingsSaveCanvas.js
```diff
@@ -33,11 +33,11 @@ export default class SettingsSaveCanvas extends LightningElement {
     @api
     handleHierarchySettingsChange(hierarchySettingsChange) {
         if (hierarchySettingsChange.settingsType === "string") {
-            hierarchySettingsChanges.settingsSingleValueBySettingsName[hierarchySettingsChange.settingsName] =
-                hierarchySettingsChange.settingsValue;
+            this.hierarchySettingsChanges.settingsSingleValueBySettingsName[hierarchySettingsChange.settingsName] =
+                hierarchySettingsChange.settingsValue; // hierarchy settingsName is prefixed with hed__ in namespaced context...how to handle this?
```
MatthewBlanski, line 37: This fix handles it
```suggestion
                hierarchySettingsChange.settingsValue;
```

force-app/main/default/lwc/settingsSaveCanvas/settingsSaveCanvas.js
```diff
@@ -67,7 +67,8 @@ export default class SettingsSaveCanvas extends LightningElement {
                     this.displayNoAccessError(exceptionType, errorMessage);
                 }
 
-                if (exceptionType === "HierarchySettingsMapper.InvalidSettingsException") {
+                if (exceptionType.includes("HierarchySettingsService.InvalidSettingsException")) {
```
MatthewBlanski, line 70: Good fix on the includes.

force-app/main/default/classes/HierarchySettingsService.cls
```diff
@@ -113,18 +122,59 @@ public virtual with sharing class HierarchySettingsService {
...
         return this.locateHierarchySettingsMapper().updateHierarchySettings(hierarchySettingsModelToUpdate);
     }
 
+    /******************************************************************************************************************************
+     * @description Append the hed__ prefix to all hierarchy settings field names specified in the settingsSingleValueBySettingsName
+     * and settingsListSettingsName of a HierarchySettingsChangesVModel when handled in a managed context.
+     * @param  hierarchySettingsChangesVModelToNamespace A HierarchySettingsChangesVModel with fields requiring a namespace prefix
+     * @return A HierarchySettingsChangesVModel
+     ******************************************************************************************************************************/
+    private HierarchySettingsChangesVModel addNamespacePrefixToHierarchySettingsChangesFields(
```
MatthewBlanski, line 146: This method seems a little heavy, let alone by making extra loops for the task at hand. Also, if we are to perform something of this nature in this fashion, I suggest instantiating a new HierarchySettingsChangesVModel and return it that model to avoid side effect.

MatthewBlanski (approved):

PR #1612, opened 2021-03-11
[EDA] Settings Navigation - Styled

MatthewBlanski:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.css
```diff
@@ -0,0 +1,3 @@
+.eda-height_full {
```
MatthewBlanski, line 1: This is only because vertical stretch was not working appropriately.

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
```diff
@@ -0,0 +1,113 @@
+import { LightningElement, api, track } from "lwc";
+import accountModelSettingsTitle from "@salesforce/label/c.stgAccountModelSettingsTitle";
+
+export default class EdaSettingsNavigation extends LightningElement {
+    labelReference = {
+        accountModelSettingsTitle,
+    };
+
+    @track settingsNavigationViewModel = {
```
MatthewBlanski, line 9: Tracked so we can see internal changes

force-app/main/default/lwc/edaSettingsNavigationSection/edaSettingsNavigationSection.css
```diff
@@ -0,0 +1,7 @@
+.eda-nav-menu-item_link:hover {
```
MatthewBlanski, line 1: Fixes a blue hover issue. Couldn't do it with default classes but using lwc tokens is an acceptable approach

force-app/main/default/lwc/edaSettingsNavigationSection/edaSettingsNavigationSection.css
```diff
@@ -0,0 +1,7 @@
...
+    color: var(--lwc-colorTextActionLabelActive);
+}
+
+.eda-nav-is-active {
```
MatthewBlanski, line 5: Necessary to set background color for UI

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
```diff
@@ -0,0 +1,112 @@
...
+        navigationSections: [
+            {
+                label: "Setup Home",
+                page: "setuphome",
+                sectionkey: "setuphome",
+            },
+            {
```
m-kinkade, line 16: Do we need to translate the text for the navigation section labels?

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
```diff
@@ -0,0 +1,112 @@
...
+                sectionkey: "peopleandgroups",
+                navigationSubSections: [
+                    {
+                        label: this.labelReference.accountModelSettingsTitle,
+                        page: "accountmodelsettings",
+                        sectionkey: "accountmodelsettings",
+                        menuItems: [
+                            { label: "Default Account Model", sectionKey: "defaultaccountmodel" },
+                            {
+                                label: "Administrative Account Record Type",
```
m-kinkade, line 27: Same question regarding these labels...do they need to be translated?

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
```diff
@@ -0,0 +1,112 @@
...
+            },
+        ],
+    };
+
+    @api setActivePage(pageName) {
+        this.settingsNavigationViewModel.navigationSections.forEach((navigationSection) => {
+            if (navigationSection.page === pageName) {
+                navigationSection.isActive = true;
+            } else {
+                navigationSection.isActive = undefined;
```
m-kinkade, line 93: Just curious as to why we opted to set to undefined here instead of false?

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
MatthewBlanski: We will eventually. These are placeholders to be replaced as we go and add the actual pages. The page and sectionkey can remain untranslated as they are back-end

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
MatthewBlanski: Likewise on translations. These are placeholders for now to be replaced by the 'real' labels.

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
MatthewBlanski: This is following truthy/falsy. More importantly, this is a format of defining variables that can be important when dealing with some attributes in lwc, such as disabled on input affordances. In a 'fun' experience with said attributes:

- When a disabled attribute is undefined, the component is considered to not be disabled.
- When disabled === false, the component is disabled. This was the headscratcher moment for me.
- When disabled === true, the component is disabled.
- Thus, when disabled is defined, the component is considered disabled. Fun with Javascript!

force-app/main/default/lwc/edaSettings/edaSettings.html
```diff
@@ -2,11 +2,15 @@
...
-                class="slds-col slds-p-top_large slds-p-left_small slds-text-heading_small slds-no-flex slds-size_4-of-12 slds-scrollable_y eda-height_full"
+                class="slds-col slds-p-top_x-small slds-no-flex slds-size_4-of-12 slds-scrollable_y eda-height_full"
+                data-qa-locator="edaSettingsNavigation"
             >
-                {labelReference.settingsNavigation}
+                <c-eda-settings-navigation onsettingsnavigation={handleSettingsNavigation}></c-eda-settings-navigation>
             </div>
-            <div class="slds-col slds-no-flex slds-size_8-of-12 slds-scrollable_y eda-height_full">
+            <div
+                class="slds-col slds-no-flex slds-size_8-of-12 slds-scrollable_y eda-height_full"
```
xinnanlisf, line 11: For the 2nd column do we need unified padding? It looks like we are relying on child components to handle paddings, which can be hard to manage.

MatthewBlanski: So, there's an issue with handling that relative to the view boxes. The padding that we do have is for text and the navigation padding relies on an iteration, which is using top padding as section height padding vs subsection height padding vs menu item padding differs. This means that the interior padding needs to use a combination of padding rather than one. If you can suggest a resolution to that difficulty, I will run with it and use it. I just couldn't figure it out.

xinnanlisf: I am not sure if we should put css classes into child component however I feel like the key is to ensure the consistency. I have been seeing paddings classes in many places. I do want to take it back thought that we can rely on child components to decide their paddings. We can talk about it more if we got the chance.

xinnanlisf: The other thing I have noticed is that many slds classes we used in our code are not quite necessary. Some of them can inherit from parent and some of them do not make a difference. Maybe we should review it at some time.

MatthewBlanski: It's an interesting question. The main thing I am trying to preserve is an explicit callout to behavior that does not depend on a parent component. Within the same component we can get away with that, but I don't want to make assumptions on the child component's context.

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
```diff
@@ -0,0 +1,112 @@
...
+                navigationSubSections: [
+                    {
+                        label: this.labelReference.accountModelSettingsTitle,
+                        page: "accountModelSettings",
+                        id: "accountModelSettings",
+                        menuItems: [
+                            { label: "Default Account Model", id: "defaultaccountmodel" },
+                            {
+                                label: "Administrative Account Record Type",
+                                menuitemkey: "administrativeAccountRecordType",
```
m-kinkade, line 28: should "menuitemkey" be replaced with "id" to align with the other menuItems?

MatthewBlanski: Good catch! Done

xinnanlisf (approved):

force-app/main/default/lwc/edaSettingsNavigationSection/edaSettingsNavigationSection.html
```diff
@@ -0,0 +1,69 @@
...
+                >
+            </template>
+        </h2>
+
+        <!--Header Only Menu Items-->
+        <template if:true={hasMenuItems}>
+            <!--Base State-->
+            <template if:false={viewModel.isActive}>
+                <ul aria-describedby={viewModel.id}>
+                    <c-eda-settings-navigation-menu-items
```
m-kinkade, line 26: I think the child component needs to be wrapped in <li></li> since it is part of an unordered list element

m-kinkade (approved):
aria-current update is working as expected :)

davidmreed (approved):

spinkelman:
QA passes - Merging - Closing WI

PR #1617, opened 2021-03-13
[EDA] People & Groups: Account Auto-Deletion Frontend

m-kinkade:
# Critical Changes

# Changes

# Issues Closed

# New Metadata
### Apex Classes
AccountAutoDeletionSettingsVMapper
AccountAutoDeletionSettingsVMapper_TEST
DualListboxVModel
DualListboxVModel_TEST
DualListboxValueVModel
DualListboxValueVModel_TEST 

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


m-kinkade:
This PR is dependent on https://github.com/SalesforceFoundation/EDA/pull/1611

force-app/main/default/lwc/settingsSaveCanvas/settingsSaveCanvas.js
```diff
@@ -53,6 +54,7 @@ export default class SettingsSaveCanvas extends LightningElement {
                 if (result === true) {
                     // update successful
                     this.switchEditMode(false);
+                    this.showToast("success", "Save Complete", this.labelReference.successMessage);
```
MatthewBlanski, line 57: This will work for now. It's interesting how slow the delay on the toast is in comparison to the actual save resolving. I'm not too worried about race conditions on async refresh apex at this juncture.

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -89,6 +89,34 @@
...
+                    <div class="slds-grid slds-grid_vertical-align-start slds-wrap">
+                        <div class="slds-col slds-size_2-of-3 slds-p-bottom_small">
+                            <div class="slds-text-title slds-text-color_default slds-p-left_small slds-text-title_bold">
+                                {labelReference.accountAutoDeletionTitle}
+                            </div>
+                            <div class="slds-text-body_small slds-text-color_weak slds-p-horizontal_large">
+                                {labelReference.accountAutoDeletionDescription}
+                            </div>
+                        </div>
+                        <div class="slds-col slds-size_2-of-3 slds-p-bottom_x-large slds-p-left_large">
```
MatthewBlanski, line 104: Let's go 3 of 3 on dual listboxes. These components have a maximum size. Total nitpick here.
```suggestion
                        <div class="slds-col slds-size_3-of-3 slds-p-bottom_x-large slds-p-left_large">
```

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -89,6 +89,34 @@
                         </div>
                     </div>
                 </template>
+
+                <!-- Account Auto Deletion Settings -->
+                <template if:true={accountAutoDeletionSettingsViewModel}>
+                    <div class="slds-grid slds-grid_vertical-align-start slds-wrap">
+                        <div class="slds-col slds-size_2-of-3 slds-p-bottom_small">
```
MatthewBlanski, line 96: Keeping 2 of 3 here is a good idea since we're limiting the area to match for other descriptions and titles.

force-app/main/default/classes/AccountAutoDeletionSettingsVMapper.cls
```diff
@@ -0,0 +1,124 @@
...
+     *******************************************************************************************************/
+    public virtual DualListboxVModel getAccountAutoDeletionSettingsViewModel() {
+        // get current values set for deletion from hierarchy settings
+        AccountRecordTypeAutoDeletionModel currentAutoDeletionSettings = this.locateAccountRecordTypeAutoDeletionService()
+            .getAccountRecordTypeAutoDeletionModel();
+
+        // get all active account record types in org
+        Map<Id, RecordTypeModel> activeAccountRecordTypeModelsById = this.locateRecordTypeService()
+            .getActiveRecordTypeModelsById(
+                this.locateRecordTypeService().getModelByIdMapForSObjectType(Account.SObjectType).values()
```
MatthewBlanski, line 73: This is efficient. We tend to assign local variables here, but this is also refactor-safe. I'm happy keeping this as it is.

force-app/main/default/classes/AccountAutoDeletionSettingsVMapper.cls
```diff
@@ -0,0 +1,124 @@
...
+                    activeAccountRecordTypeModelsById.get(recordTypeId).id
+                )
+            );
+        }
+
+        // populate selected listbox - only requires the String representing the value of the selected option
+        for (
+            Id accountRecordTypeIdToDelete : (List<Id>) currentAutoDeletionSettings.accountRecordTypeIdsForDeletionWhenOrphanedList
+        ) {
+            if (activeAccountRecordTypeModelsById.containsKey(accountRecordTypeIdToDelete) == false) {
```
MatthewBlanski, line 94: This logic may be repeatable when we get to addresses and other settings that can get played with. Consider making a filter method into the record type service. Something like filterListForValidIds, which returns a list in the same order and uses a set of record type ids to validate that against. Leaving it this way, for now, is also fine, but that kind of behavior is not unique to this setting.

MatthewBlanski (changes requested):
I have one minor nitpick change on the styling for the area taken up by the dual listbox. All other changes are observations.

force-app/main/default/classes/AccountAutoDeletionSettingsVMapper.cls
```diff
@@ -0,0 +1,116 @@
...
+                    activeAccountRecordTypeModelsById.get(recordTypeId).name,
+                    activeAccountRecordTypeModelsById.get(recordTypeId).id
+                )
+            );
+        }
+
+        // populate selected listbox - only requires the String representing the value of the selected option
+        List<String> selectedOptionsDualListboxValuesList = (List<String>) this.locateRecordTypeService()
+            .filterListForValidIds(
+                (List<Id>) currentAutoDeletionSettings.accountRecordTypeIdsForDeletionWhenOrphanedList,
```
MatthewBlanski, line 92: Unfortunately, we can't guarantee that this is a list of ids. Let's pass in the string list and do a two step process in the actual method that skips a string if it is not an id and then skips if it is not in the set.

When I go and mangle the setting with this implementation it doesn't show up in the UI

force-app/main/default/classes/RecordTypeService.cls
```diff
@@ -86,6 +86,30 @@ public virtual with sharing class RecordTypeService {
...
+     * @return A List<Id> containing Ids that correspond to valid Record Type models and match a value in the specified list of Ids to find.
+     ***********************************************************************************************************************/
+    public virtual List<Id> filterListForValidIds(
+        List<Id> idsToFindList,
+        Map<Id, RecordTypeModel> validRecordTypeModelsById
+    ) {
+        List<Id> validIdsList = new List<Id>();
+
+        for (Id idToFind : idsToFindList) {
+            if (validRecordTypeModelsById.containsKey(idToFind) == false) {
```
MatthewBlanski, line 102: Considering this implementation, we could use `getRecordTypeModelFromMap` for consistent and reusable logic. @xinnanlisf, how do you feel about making this method lighter weight and relying on a `List<String>` and `Set<Id>` vs using a `List<String>` and `Map<Id,RecordType>` with `getRecordTypeModelFromMap`?

MatthewBlanski (changes requested):
The id filtering method is causing a non-obvious issue if I change the value in the back end to not an Id. By using strings and then detecting for an ID similar `RecordTypeService. getRecordTypeModelFromMap` we can obviate this issue.

force-app/main/default/classes/RecordTypeService.cls
m-kinkade: I prefer the lighter weight version, it seems more straightforward to me

MatthewBlanski: Make it so!

MatthewBlanski (approved):

PR #1643, opened 2021-03-25
Styling updates

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
```diff
@@ -1,123 +1,119 @@
 <template>
-    <div class="slds-grid--vertical slds-theme_default">
```
xinnanlisf, line 2: all right, github cant even figure out which line is removed...

Anyway, we don't need these css classes since inheriting from parent component.
Also, we can move slds-p-top_large slds-p-left_small to edaSettingsPage.html so that each page component wont have to add this anymore.

force-app/main/default/lwc/edaSettings/edaSettings.html
```diff
@@ -2,13 +2,13 @@
...
         <div class="slds-grid slds-theme_default eda-height_full">
             <div
-                class="slds-col slds-p-top_x-small slds-no-flex slds-size_4-of-12 slds-scrollable_y eda-height_full"
+                class="slds-col slds-no-flex slds-size_4-of-12 slds-scrollable_y slds-p-top_x-small"
                 data-qa-locator="edaSettingsNavigation"
             >
                 <c-eda-settings-navigation onsettingsnavigation={handleSettingsNavigation}></c-eda-settings-navigation>
             </div>
             <div
-                class="slds-col slds-no-flex slds-size_8-of-12 slds-scrollable_y eda-height_full"
```
xinnanlisf, line 11: we dont need eda-height_full here since upper level div is having it.

force-app/main/default/lwc/edaSettings/edaSettings.html
```diff
@@ -2,13 +2,13 @@
...
+                class="slds-col slds-no-flex slds-size_8-of-12 slds-scrollable_y slds-p-top_large slds-p-left_small"
```
xinnanlisf, line 11: adding paddings here and also move it to the end.

force-app/main/default/lwc/settingsSaveCanvas/settingsSaveCanvas.html
```diff
@@ -8,7 +8,6 @@
                     label={labelReference.settingsButtonEdit}
                     title={labelReference.settingsButtonEdit}
                     onclick={handleEditClick}
-                    class="slds-p-right_x-small"
```
xinnanlisf, line 11: removing these because based on the mocks, buttons should be aligned with affordance.

xinnanlisf:
@MatthewBlanski, I made a few tweaks to styling and left comments for those changes. With these changes, individual settings page does not need to worry about styling any more also given that we will create reusable components. All they need to do is adding a save canvas! Let me know what do you think!

force-app/main/default/lwc/accountModelSettings/accountModelSettings.html
MatthewBlanski: Have you tried rendering these on their own? We want them to be able to be inserted directly into modals.

force-app/main/default/lwc/edaSettings/edaSettings.html
MatthewBlanski: We do because each content area has its own definition of size. So what we're specifying in this instance is that the overall settings container needs to take up 100% of the height of the content area it's in and then within that the nav and page need to take up the height of the settings container. Then they need to be scrollable on the y axis.

force-app/main/default/lwc/edaSettings/edaSettings.html
MatthewBlanski: Adding the padding to start with here is a good idea. This should keep consistency throughout.

force-app/main/default/lwc/settingsSaveCanvas/settingsSaveCanvas.html
```diff
@@ -23,16 +22,15 @@
...
                 ></lightning-button>
             </div>
         </div>
     </div>
 
     <!-- Display Content from Parent Component Body -->
-    <div class="slds-theme_default eda-height_full">
-        <div class="slds-m-left_x-small">
-            <!-- Slot that will be populated with body from parent component -->
+    <div class="slds-theme_default eda-height_full slds-m-left_x-small">
```
xinnanlisf, line 31: adding extra divs here. Also move slds-m-left_x-small to upper level.

force-app/main/settingsComponentLibrary/lwc/settingsRowInput/settingsRowInput.html
```diff
@@ -1,11 +1,11 @@
 <template>
-    <div class="slds-grid slds-grid_vertical-align-start slds-wrap">
-        <div class="slds-col slds-size_2-of-3 slds-p-bottom_x-large">
+    <div class="slds-grid slds-grid_vertical-align-start slds-wrap slds-p-bottom_x-large">
```
xinnanlisf, line 2: having bottom padding for each 'setting' not each 'div' in reusable components.

force-app/main/settingsComponentLibrary/lwc/settingsRowDualListbox/settingsRowDualListbox.html
```diff
@@ -1,21 +1,23 @@
 <template>
-    <div class="slds-col slds-size_2-of-3 slds-p-bottom_x-large">
-        <div class="slds-text-title slds-text-color_default slds-p-left_small slds-text-title_bold">{title}</div>
-        <div class="slds-text-body_small slds-text-color_weak slds-p-horizontal_large">{description}</div>
-    </div>
+    <div class="slds-grid slds-grid_vertical-align-start slds-wrap slds-p-bottom_x-large">
```
xinnanlisf, line 2: aligned with the other reusable component.

xinnanlisf:
@MatthewBlanski this should be ready i think. I added more tweaks and also reverted my changes to account page since it will be updated by Merle's PR. Let me know what you think.

MatthewBlanski (changes requested):
Please apply the styling to Account Model and Contact Information pages.

force-app/main/default/lwc/settingsSaveCanvas/settingsSaveCanvas.html
xinnanlisf: Done! It looks like the sizes of cancel and save are different. I left them as they were since the mocks was showing the same way.

xinnanlisf:
@MatthewBlanski i believe the new stylings are in space for contact information and affiliations too. I am hesitating to update the account model one since we are merging naming component. It looks like we are adding some more reusable components and I do remember merging between child feature branches can be a headache sometimes. I am wondering if we can merge all these changes first and then once the account naming setting is there we can update it. I guess we can update naming branch too depends on which PR is merged first.

MatthewBlanski (approved):

PR #1675, opened 2021-04-12
[EDA] System Settings: System Tools Nav Section and Page

skosgi1:
# Critical Changes

# Changes

# Issues Closed

# New Metadata
**Custom Labels**

stgSystemToolsNav
stgSystemToolsTitle

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/lwc/systemTools/systemTools.html
```diff
@@ -0,0 +1,9 @@
+<template>
+    <div class="slds-grid--vertical slds-theme_default">
```
xinnanlisf, line 2: is there a reason why this is not using saving canvas?

skosgi1: There is no edit and save/cancel for this page @xinnanlisf

xinnanlisf (approved):

PR #1701, opened 2021-04-23
Auto Enrollment Mappings View

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
```diff
@@ -78,7 +81,7 @@ export default class EdaSettingsNavigation extends LightningElement {
             storeErrors: stgStoreErrorsTitle,
         },
         programSettings: {
-            programSettings: "Programs",
+            programSettings: stgProgramsSettingsNav,
             programAutoEnrollmentMappings: "Program Auto-Enrollment Mappings",
```
MatthewBlanski, line 85: Since we're adding program auto enrollment mappings, we should also be setting this label.

force-app/main/default/lwc/edaSettingsTable/edaSettingsTable.js
```diff
@@ -0,0 +1,22 @@
...
+    @api tabletitle;
+    @api tabledescription;
+    @api tabledata;
+    @api tablecolumns;
+
+    labelReference = {
+        newButtonLabel: stgBtnAddMapping,
+    };
+
+    handleNewRowClick(event) {
```
MatthewBlanski, line 15: If we are going to make this generic, then let's add in some events that we can use later.

MatthewBlanski (changes requested):
A few minor changes. One on label in the nav and then event creation and passing for the generic table component.

force-app/main/default/lwc/edaSettingsTable/edaSettingsTable.js
xinnanlisf: Should we add them in edit/delete/new items

force-app/main/default/lwc/edaSettingsNavigation/edaSettingsNavigation.js
xinnanlisf: Not sure if I am following. I thought we are adding auto enrollment to the same page as other program settings?

MatthewBlanski: I mean on line 85, you have a label for the table that is the title of the display. That is "Program Auto-Enrollment Mappings". So we should include that in the nav.

force-app/main/default/lwc/edaSettingsTable/edaSettingsTable.js
MatthewBlanski: If we are using a genericized table, then we should add genericized events as well. This can be a simple passthrough that replicates the important details. My understanding is that handleNewRowClick on its own doesn't bubble and is captured at the component level, so we should create a new custom event and dispatch it so we can handle it in the parent component.

The same would apply to the row action event. Replicate the relevant data and pass it upstream.

xinnanlisf:
@MatthewBlanski this should be ready for review again! Again I rolled back the reusable component for tables.

force-app/main/default/lwc/programSettings/programSettings.js
```diff
@@ -0,0 +1,123 @@
...
+        this.affordancesDisabledToggle = false;
+        this.refreshAllApex();
+    }
+
+    handleSettingsSaveCancel(event) {
+        this.refreshAllApex();
+    }
+
+    refreshAllApex() {
+        /*Promise.all([
```
MatthewBlanski, line 107: Please add the view refresh to this as this is a necessary component for "viewing"

MatthewBlanski (changes requested):

force-app/main/default/lwc/programSettings/programSettings.js
xinnanlisf: Done!

MatthewBlanski (approved):

xinnanlisf:
@MatthewBlanski i have fixed the empty values for role and status. Also fixed the tell me more hyper link fyi

spinkelman:
QA Testing Passes

PR #1707, opened 2021-04-26
[EDA]Lightning App

MatthewBlanski:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


lifewithryan (approved):
Happy Path testing seems to work. I didn't dig deep, the app loads, I am able to get to Settings, made a change and verified the setting saved by going to the old EDA Settings.

I'm not seeing anything that needs to be addressed. Only question would be is if we need to put a "loading" gif/image in our base settings component while the settings load. That's definitely minor though.

iskander-sfdc (approved):
Signed up a scratch org, new app and tab were created successfully

spinkelman:
QA Passes. Merging

PR #1711, opened 2021-04-28
[EDA]  Lead Conversion Account Naming

mrrigney:
# Critical Changes

# Changes

# Issues Closed
[[W-9021439]](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008xkZJIAY/view)

# New Metadata
**Classes**
ACCT_RenameLeadConvertedAccounts_TDTM

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


mrrigney:
@MatthewBlanski  Couple notes around this PR.  I didn't refactor UTIL_ACCT_Naming.defaultAccountName method only added 'theAccount' parm as we talked because turns out 'switch' statements are not available for record ID's in Apex.  This lessons the chance of breaking any existing code. Also, I wrote a bulk test in ACCT_RenameLeadConvertedAccounts_TEST.bulkConvertLeadAdminRecordType which uses a static 'leadCount' attribute but didn't it call anyplace in the unpackaged because I wasn't sure if we wanted to do this and where.  I should have time to add this and do any additional test methods you might suggest, just wanted to get the base PR to you for now.

mrrigney:
@MatthewBlanski I just came across an interesting thing I hadn't noticed before.  The method ACCT_IndividualAccounts_TDTM.handleInsertProcessing has always been renaming Household accounts during lead conversion as it calls the UTIL_ACCT_Naming.updateHHAccountsName method every time.  This does not happen for Administrative accounts which is why we started this whole process.  So now the question becomes, if admins don't select Household accounts to rename, should we remove this logic and use the company name?  Obviously this will add more complexity as we will have to determine lead conversion context from this place.

MatthewBlanski:
> @MatthewBlanski I just came across an interesting thing I hadn't noticed before. The method ACCT_IndividualAccounts_TDTM.handleInsertProcessing has always been renaming Household accounts during lead conversion as it calls the UTIL_ACCT_Naming.updateHHAccountsName method every time. This does not happen for Administrative accounts which is why we started this whole process. So now the question becomes, if admins don't select Household accounts to rename, should we remove this logic and use the company name? Obviously this will add more complexity as we will have to determine lead conversion context from this place.

@mrrigney  At this juncture in the process, altering any base functionality introduces a risk that I do not wish to introduce. That is further afield than refactoring code. I would rather go to production preserving and appending to existing behavior, even if one setting is slightly inaccurate. I am more comfortable logging a bug and moving on to readdress later.

mrrigney:
> > @MatthewBlanski I just came across an interesting thing I hadn't noticed before. The method ACCT_IndividualAccounts_TDTM.handleInsertProcessing has always been renaming Household accounts during lead conversion as it calls the UTIL_ACCT_Naming.updateHHAccountsName method every time. This does not happen for Administrative accounts which is why we started this whole process. So now the question becomes, if admins don't select Household accounts to rename, should we remove this logic and use the company name? Obviously this will add more complexity as we will have to determine lead conversion context from this place.
> 
> @mrrigney At this juncture in the process, altering any base functionality introduces a risk that I do not wish to introduce. That is further afield than refactoring code. I would rather go to production preserving and appending to existing behavior, even if one setting is slightly inaccurate. I am more comfortable logging a bug and moving on to readdress later.

@MatthewBlanski agreed thank you, will just want to document this

force-app/main/default/classes/ACCT_RenameLeadConvertedAccounts_TEST.cls
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * @author Salesforce.org
+ * @date 2011 (1.x)
+ * @group Accounts
+ * @group-content ../../ApexDocContent/Accounts.htm
+ * @description Tests for ACCT_RenameLeadConvertedAccounts_TDTM.
+ */
+@isTest
+private class ACCT_RenameLeadConvertedAccounts_TEST {
+    private static Integer leadCount = 5; // used to determine number of test contact records to create
```
MatthewBlanski, line 39: Not performing bulk tests in this scenario is acceptable. The most important test is that the functionality works over multiple records. I consider the 200 record test the lowest priority portion of this WI.

force-app/main/default/classes/UTIL_ACCT_Naming.cls
```diff
@@ -246,17 +246,17 @@ public class UTIL_ACCT_Naming {
...
     */
-    private static String defaultAccountName(List<Contact> cons, String accountNamingFormat) {
+    private static String defaultAccountName(List<Contact> cons, String accountNamingFormat, Account theAccount) {
         Set<String> householdLastNames = new Set<String>();
         String finalAccountName;
         String name = '';
         String strAnd = ' ' + Label.defaultNamingConnector + ' ';
         Integer counter = 0;
 
         for (Contact con : cons) {
```
MatthewBlanski, line 256: This portion of the method is in sore need of a refactor and currently attempts to perform too many operations. We have a number of unnecessary checks in it that are also muddying up the code.

That said, reviewing this has made me realize there was an architectural miss on this piece that the original architecture accounted for. `con.LastName + ' ' + System.label.DefaultAdminName` does not account for all of the varieties of an administrative account name.

MatthewBlanski (changes requested):
A few noted changes and unfortunately one major one.

force-app/main/default/classes/UTIL_ACCT_Naming.cls
MatthewBlanski: Ideally, we would use a number of guard clauses as two of the states stop after the first contact. I envision the logic flow as:

- if the list of contacts is empty, return null
- If the account is an administrative account
  - return `finalAccountName = con[0].LastName + ' ' + System.label.DefaultAdminName`
- if the account is not a household account
  - return `finalAccountName = con[0].LastName + ' ' + System.label.DefaultAccountName`
- if the account naming format is not `hhNameFormat`
  - return null
- Household Account record type and naming format is assumed from this point onward
- for the contact in the list of contacts (standard for loop)
  - Add the contact's Last Name to `name`
  - if we are not on the last contact
  - add `strAnd` to name
  - return `name + ' ' + System.label.DefaultHouseholdName`
  
  This is an exact duplicate of the logic occurring minus some unnecessary cycles. It's interesting that there's a logic flaw in default account name in which a household account may go null, but preserving existing behavior is the goal of a refactor on this method. Theoretically we could change `accountNamingFormat` to be a boolean since that is its only true purpose in this method.

MatthewBlanski:
> @MatthewBlanski You mentioned not performing bulk tests was acceptable, so I am testing that 5 records convert successfully in a batch, do you think that is that good enough for this pr?

5 record batches has been good enough in EDA in other features as well. I'm comfortable drawing the line there.

force-app/main/default/classes/UTIL_ACCT_Naming.cls
MatthewBlanski: If you could take a shot at refactoring it now, I would appreciate it. This method stands out like a sore thumb to me.

mrrigney:
@MatthewBlanski I refactored the defaultAccountName method and all of the test classes still pass.  There was one question I wasn't sure about, you mentioned 'if the account naming format is not hhNameFormat' What does that mean?  The settings will always have a hhNameFormat defined I believe and we know we are on a HH record type, so is there an additional check I am missing?  Believe everything is ready for full code review

MatthewBlanski:
> @MatthewBlanski I refactored the defaultAccountName method and all of the test classes still pass. There was one question I wasn't sure about, you mentioned 'if the account naming format is not hhNameFormat' What does that mean? The settings will always have a hhNameFormat defined I believe and we know we are on a HH record type, so is there an additional check I am missing? Believe everything is ready for full code review

There is a scenario within the method itself that does not necessitate that we are using the hhNameFormat and the Household Account Record type at the same time.

MatthewBlanski (approved):

paulmalenke (approved):

annejackson:
AC testing has passed. There were some bugs that have been written up to be triaged and fixed in future sprints (or that need more dev research to determine level of effort). 

@rhakeshj  and @spinkelman are doing some regression testing, will await their results before merging.

spinkelman:
After confirming with Anne, QA Passes - Merging

PR #1712, opened 2021-04-28
Auto Enrollment Edit Modal

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/aura/autoEnrollmentMappingModal/autoEnrollmentMappingModal.cmp
```diff
@@ -0,0 +1,17 @@
+<aura:component>
+    <aura:registerEvent name="modalDataChangeEvent" type="c:modalDataChangeEvent" />
+
+    <aura:attribute name="actionName" type="String" />
+    <aura:attribute name="accountRecordType" type="String" />
+    <aura:attribute name="autoProgramEnrollmentStatus" type="String" />
+    <aura:attribute name="autoProgramEnrollmentRole" type="String" />
```
MatthewBlanski, line 7: This looks like a good foundation.

force-app/main/default/aura/autoEnrollmentMappingModal/autoEnrollmentMappingModal.cmp
```diff
@@ -0,0 +1,17 @@
...
+    abc{!v.autoProgramEnrollmentStatus}
+    <c:autoEnrollmentMappingModalBody
+        actionName="{!v.actionName}"
+        accountRecordType="{!v.accountRecordType}"
+        autoProgramEnrollmentStatus="{!v.autoProgramEnrollmentStatus}"
+        autoProgramEnrollmentRole="{!v.autoProgramEnrollmentRole}"
+        onaccountrecordtypechange="{!c.handleAccountRecordTypeChange}"
+        onautoenrollmentmappingstatuschange="{!c.handleAutoEnrollmentMappingStatusChange}"
```
MatthewBlanski, line 15: You'll need to add a role change event handler here too.

force-app/main/default/aura/autoEnrollmentMappingModalOpener/autoEnrollmentMappingModalOpener.cmp
```diff
@@ -0,0 +1,28 @@
...
+    />
+    <aura:handler
+        name="modalFooterEventHandler"
+        event="c:customModalFooterEvent"
+        action="{!c.handleModalFooterEvent}"
+    />
+    <aura:registerEvent name="modalSaveEvent" type="c:modalSaveEvent" />
+    <lightning:overlayLibrary aura:id="edaOverlayLibrary" />
+
+    <aura:method name="openAutoEnrollmentMappingModal" action="{!c.openAutoEnrollmentMappingModal}">
```
MatthewBlanski, line 21: This is also a solid foundation

force-app/main/default/classes/AffiliationMappingsService.cls
```diff
@@ -101,6 +101,15 @@ public virtual with sharing class AffiliationMappingsService {
         this.locateAffiliationMappingsMapper().updateAffiliationMappings(afflMappingModel);
     }
 
+    /*****************************************************************************
+     * @description Checks user permissions and passes AffiliationMappingsModel to
+     * updateAffiliationMappings.
+     *****************************************************************************/
+    public void updateAutoEnrolllmentMapping(AffiliationMappingsModel affiliationMappingsModel) {
```
MatthewBlanski, line 108: You may need a slightly different entry method as an "update" may span two mappings.

For instance, if I update the Academic Program Mapping to instead be the record type for Business Organization, I'll need to determine the difference in those and update both records accordingly. The expected behavior for this would be to that Academic Program would have its auto enrollment, status, and role fields cleared while the Business Organization mapping would have its fields set to the new values.

The inability to select an unmapped account record type should protect us from almost all collision issues, at least until we separate the mappings from a data standpoint.

MatthewBlanski (commented):
We're off to a good start here. I've left a few comments on concerns I might have and I'm wondering now if ProgramSettingsVMapper is quite what we want. Perhaps AutoEnrollmentSettingsVMapper would have been the better choice. It's not too late to change it.

force-app/main/default/aura/autoEnrollmentMappingModal/autoEnrollmentMappingModal.cmp
xinnanlisf: yep i did, just havent pushed yet

force-app/main/default/classes/AffiliationMappingsService.cls
xinnanlisf: Making sure that i understand it correctly:

1) For auto enrollment editing, we are only updating status and role for 'enabled' auto enrollment and we don't allow users to update account record types.

2) Updating account record types should fall on affiliation mappings settings page and we might need to refactor the updating affiliations method so that we can update both affiliation mappings.

Either way I think we need to update the user stories too.

MatthewBlanski: Not quite on that. More that "editing" the Account Record type should work seamlessly from the user's standpoint, but what we need to do is to have it update 2 back end records instead of 1.

force-app/main/default/classes/AffiliationMappingsService.cls
```diff
@@ -101,6 +101,29 @@ public virtual with sharing class AffiliationMappingsService {
         this.locateAffiliationMappingsMapper().updateAffiliationMappings(afflMappingModel);
     }
 
+    /*****************************************************************************
+     * @description Checks user permissions and update the auto enrollment mapppings.
+     *****************************************************************************/
+    public void updateAutoEnrollmentMappings(
```
MatthewBlanski, line 107: You're on the right path here. We do need to update both instances, but only need to perform one update if the new and old account record types match.

MatthewBlanski: As an additional note, if you have the name of the original mapping here, you can directly go to the mapper for that one since you know the relevant data points to update, name, status, and role.

force-app/main/default/classes/AffiliationMappingsService.cls
```diff
@@ -101,6 +101,38 @@ public virtual with sharing class AffiliationMappingsService {
...
+        String mappingName,
+        String oldAccountRecordType,
+        String newAccountRecordType,
+        String autoProgramEnrollmentStatus,
+        String autoProgramEnrollmentRole
+    ) {
+        this.locateUserPermissionService().checkViewSetupForCurrentUser();
+        if (oldAccountRecordType != newAccountRecordType) {
+            List<AffiliationMappingsModel> afflMappingModels = this.locateAffiliationMappingsMapper()
```
MatthewBlanski, line 116: You still need to update the old mapping in this use case.

force-app/main/default/classes/AffiliationMappingsMapper.cls
```diff
@@ -107,6 +107,56 @@ public virtual with sharing class AffiliationMappingsMapper {
         update afflSettings;
     }
 
+    /*****************************************************************************
+     * @description Updates the auto enrollment mapping on affiliation mapping record
+     * @param mappingName
+     * @param autoProgramEnrollmentStatus
+     * @param autoProgramEnrollmentRole
+     *****************************************************************************/
+    public void updateAutoEnrollmentMappings(
```
MatthewBlanski, line 116: Consider making a method that takes a list of mappings to update and performs the conversion from models to mappings, then updates all mappings in that list.

force-app/main/default/classes/AffiliationMappingsMapper.cls
```diff
@@ -107,6 +107,56 @@ public virtual with sharing class AffiliationMappingsMapper {
...
+
+    /*****************************************************************************
+     * @description Copies the status and role from old auto enrollment mapping to
+     * new auto ernollment mapping.
+     * Also, disables auto enrollment for the old auto enrollment mapping
+     * @param mappingName
+     * @param autoProgramEnrollmentStatus
+     * @param autoProgramEnrollmentRole
+     *****************************************************************************/
+    public void copyStatusAndRoleAutoEnrollmentMapping(
```
MatthewBlanski, line 137: The logic for manipulation might be better served in the service. You can instantiate two models, set the appropriate values, then pass them to be updated.

force-app/main/default/classes/AffiliationMappingsService.cls
```diff
@@ -101,6 +101,47 @@ public virtual with sharing class AffiliationMappingsService {
...
+                .getAffiliationMappingsModel();
+            List<Affl_Mappings__c> affliliaitonMappingsToUpdateList = new List<Affl_Mappings__c>();
```
MatthewBlanski, line 118: This has a spelling issue.
```suggestion
            List<Affl_Mappings__c> affliliationMappingsToUpdateList = new List<Affl_Mappings__c>();
```

force-app/main/default/classes/AffiliationMappingsService.cls
```diff
@@ -101,6 +101,47 @@ public virtual with sharing class AffiliationMappingsService {
...
+     *****************************************************************************/
+    public virtual void updateAutoEnrollmentMappings(
+        String mappingName,
+        String oldAccountRecordType,
+        String newAccountRecordType,
+        String autoProgramEnrollmentStatus,
+        String autoProgramEnrollmentRole
+    ) {
+        this.locateUserPermissionService().checkViewSetupForCurrentUser();
+        if (oldAccountRecordType != newAccountRecordType) {
```
MatthewBlanski, line 115: Consider using the more simple use case first and make it function as a guard clause instead of an if, else...

force-app/main/default/classes/AffiliationMappingsService.cls
```diff
@@ -101,6 +101,47 @@ public virtual with sharing class AffiliationMappingsService {
...
+                    affliliaitonMappingsToUpdateList.add(newAfflSetting);
+                    break;
+                }
+            }
+            update affliliaitonMappingsToUpdateList;
+        } else {
+            Affl_Mappings__c afflSetting = Affl_Mappings__c.getInstance(mappingName);
+            afflSetting.Auto_Program_Enrollment_Status__c = autoProgramEnrollmentStatus;
+            afflSetting.Auto_Program_Enrollment_Role__c = autoProgramEnrollmentRole;
+            update afflSetting;
```
MatthewBlanski, line 141: This behavior should be moved to the mapper. We shouldn't be performing any DML operations in a service.

MatthewBlanski (changes requested):
We're not respecting the separation of concerns and abstraction between mapper and services in this PR.

MatthewBlanski (approved):

force-app/main/default/lwc/autoEnrollmentMappingModalBody/autoEnrollmentMappingModalBody.html
```diff
@@ -0,0 +1,55 @@
...
+    <div
+        class="slds-grid slds-grid--vertical slds-grid_vertical-align-start slds-wrap slds-p-left_large slds-p-right_large slds-p-bottom_xx-large"
+    >
+        <template if:true={modifyRecords}>
+            <div class="slds-col slds-size-1-of-1 slds-p-bottom_large">
+                <lightning-formatted-rich-text value={autoEnrollmentMappingModalDesc}> </lightning-formatted-rich-text>
+            </div>
+            <template if:true={accountRecordTypeComboboxVModel}>
+                <div class="slds-col slds-size_2-of-3">
+                    <lightning-combobox
```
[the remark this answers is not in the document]
MatthewBlanski: In this case we might be better suited to making a new form of settings component if we wanted to genericize, similar to what we did in Affiliations. At this juncture, I don't feel like it's worth the LOE to do so as the requirements are ever so slightly different with the API name display or not.

skosgi1 (approved):
Looks great!

force-app/main/default/lwc/autoEnrollmentMappingModalBody/autoEnrollmentMappingModalBody.js
```diff
@@ -0,0 +1,167 @@
...
+    @wire(getAccountRecordTypeComboboxVModel, {
+        accountRecordType: "$oldAccountRecordType",
+    })
+    accountRecordTypeComboboxVModelWire(result) {
+        this.accountRecordTypeComboboxWireResult = result;
+
+        if (result.data) {
+            this.accountRecordTypeComboboxVModel = result.data;
+        } else if (result.error) {
+            //console.log("error retrieving accountRecordTypeComboboxVModel");
```
iskander-sfdc, line 58: When/how do we display a callout error? Is there a separate story for this?

PR #1776, opened 2021-05-27
Custom labels for 232

skosgi1:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


MatthewBlanski (changes requested):
This is looking good so far. Let's add the appropriate labels to the navigation and clear up setup home in there so we can confirm that we do not have any hardcoded values.

Also, please update every instance called out in the spec for success toast titles with stgSuccess.

force-app/main/default/translations/ca.translation-meta.xml
```diff
@@ -1,5 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Translations xmlns="http://soap.sforce.com/2006/04/metadata">
+    <customApplications>
+        <label><!-- Education Data Architecture --></label>
+        <name>Education_Data_Architecture</name>
+    </customApplications>
```
MatthewBlanski, line 6: This needs to be removed in every translation file. The App is unmanaged.

force-app/main/default/translations/ca.translation-meta.xml
```diff
@@ -1698,8 +1922,16 @@
...
+        <name>Education_Cloud_Settings</name>
+    </customTabs>
     <customTabs>
         <label>Configuració d&apos;EDA</label>
         <name>HEDA_Settings</name>
     </customTabs>
+    <customTabs>
+        <label><!-- Settings Health Check --></label>
+        <name>Settings_Health_Check</name>
+    </customTabs>
```
MatthewBlanski, line 1936: This also needs to be removed in every file since it is unmanaged.
```suggestion
    <customTabs>
        <label><!-- Settings Health Check --></label>
        <name>Settings_Health_Check</name>
    </customTabs>
```

xinnanlisf (approved):
Looks good. I know you are double checking translations but I will approve now.

spinkelman:
QA passes - Merging

PR #1792, opened 2021-06-11
 [EDA] Relationship mappings new/edit/delete

iskander-sfdc:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/classes/RelationshipLookupMapper_TEST.cls
```diff
@@ -25,68 +25,273 @@
...
+            Name = 'test' + system.now(),
+            Active__c = true,
+            Female__c = 'wife',
+            Male__c = 'husband',
+            Neutral__c = 'spouse'
+        );
 
         Test.startTest();
-             List<RelationshipLookupModel> relationshipLookupModelsList = RelationshipLookupMapper.getInstance().getRelationshipLookupModels();
+        List<RelationshipLookupModel> relationshipLookupModelsList = RelationshipLookupMapper.getInstance()
```
MatthewBlanski, line 100: This is not a required change and something I've been commenting very loosely upon. Consider what you are testing uniquely within the test block and what metrics you are looking to retrieve? Is getting the instance part of those metrics or are we testing the performance of an individual method?

force-app/main/default/lwc/relationshipMappingModalBody/relationshipMappingModalBody.html
```diff
@@ -0,0 +1,86 @@
...
+                        type="toggle"
+                        label={labelReference.isActive}
+                        id={inputAttributeReference.isActive}
+                        name={inputAttributeReference.isActive}
+                        onchange={handleIsActiveChange}
+                        checked={isActive}
+                        value={isActive}
+                    ></lightning-input>
+                </div>
+            </template>
```
MatthewBlanski, line 80: Do we need to reassert the same condition across all of these? I would imagine we can wrap them in a single template if:true. This is functional, but causes more checks than are necessary on the render.

MatthewBlanski (changes requested):
I left two simple comments, however, the navigation label for this item should be stgTabReciprocalSettings in the nav menu. Right now it is "New Reciprocal Settings". Please change that over in the eda settings navigation component.

xinnanlisf (approved):

force-app/main/default/lwc/relationshipMappingModalBody/relationshipMappingModalBody.html
iskander-sfdc: Thanks, I changed the template to use only one if:true condition

iskander-sfdc:
> I left two simple comments, however, the navigation label for this item should be stgTabReciprocalSettings in the nav menu. Right now it is "New Reciprocal Settings". Please change that over in the eda settings navigation component.

Changed the label in the nav menu to stgTabReciprocalSettings

force-app/main/default/classes/RelationshipLookupMapper_TEST.cls
iskander-sfdc: I think this is just a formatting change in this specific place

MatthewBlanski (approved):

rhakeshj:
QA passed

PR #1802, opened 2021-06-24
Lead Conversion is not detecting pre-existing accounts

mrrigney:
# Changes
- We fixed an issue that caused Lead-converted Account naming not to work as expected when you select an existing Administrative Account during conversion. Now, if you select an existing Administrative Account when converting a Lead, the Account name doesn’t change.

# Issues Closed
[[W-9230954]](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000L48nYAC/view)
[[W-9229603]](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000L2U6YAK/view)

# Testing Notes


paulmalenke (approved):
👍  Good from my perspective.

agoolsby (approved):
@mrrigney I took a pass at drafting a release note (see the Changes area). Please review and let me know if I got it wrong. Thanks!

MatthewBlanski (approved):
Not only is this good, it's using a constant which gives us a lever to change rapidly if we decide we need to modify it. Thank you!

spinkelman:
QA Passes both the AC and EDA teams. Merging

PR #1815, opened 2021-07-06
Test Additions for Trust for Auto Enrollment Bug

MatthewBlanski:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
+++ b/force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
@@ -235,6 +235,7 @@ public virtual with sharing class AFFL_MultiRecordTypeMapper {
             autoRole = accTypeToEnrollCreateRole.get(affl.Affiliation_Type__c);
             autoStatus = accTypeToEnrollCreateStatus.get(affl.Affiliation_Type__c);
         }
+
         if (
             autoEnrollment &&
             affl != null &&
--- a/force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
+++ b/force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
@@ -28,74 +28,73 @@
     POSSIBILITY OF SUCH DAMAGE.
 */
 /**
-* @author Salesforce.org
-* @date 2015
-* @group Affiliations
-* @group-content ../../ApexDocContent/Affiliations.htm
-* @description Helps with the mapping of Account record types to Account lookup fields on Contact defined
-*				in the Affiliation Mapping custom settings.
-*/
-public with sharing class AFFL_MultiRecordTypeMapper {
+ * @author Salesforce.org
+ * @date 2015
+ * @group Affiliations
+ * @group-content ../../ApexDocContent/Affiliations.htm
+ * @description Helps with the mapping of Account record types to Account lookup fields on Contact defined
+ *				in the Affiliation Mapping custom settings.
+ */
+public virtual with sharing class AFFL_MultiRecordTypeMapper {
+    /*******************************************************************************************************
+     * @description Map of Contact field names to Contact field labels.
+     ********************************************************************************************************/
+    public Map<String, String> contactFieldAPINameByLabel;
 
     /*******************************************************************************************************
-    * @description Map of Contact field names to Contact field labels.
-    ********************************************************************************************************/
-    public Map<String, String> contactFieldAPINameByLabel;
-        
-    /*******************************************************************************************************
-    * @description Affiliation Account-record-type-to-Contact-field mappings.
-    ********************************************************************************************************/
+     * @description Affiliation Account-record-type-to-Contact-field mappings.
+     ********************************************************************************************************/
     public Map<String, String> primaryAfflFieldInMappingsByAccRecType;
-    
+
     /*******************************************************************************************************
-    * @description Affiliation Account-record-type-to-Academic-auto-create-field mappings.
-    ********************************************************************************************************/
+     * @description Affiliation Account-record-type-to-Academic-auto-create-field mappings.
+     ********************************************************************************************************/
     public Map<String, Boolean> accTypeToEnrollCreate;
     public Map<String, String> accTypeToEnrollCreateRole;
     public Map<String, String> accTypeToEnrollCreateStatus;
 
     /*******************************************************************************************************
-    * @description List of account record types in mappings.
-    ********************************************************************************************************/
+     * @description List of account record types in mappings.
+     ********************************************************************************************************/
     public List<String> accountRecordTypes;
 
     /*******************************************************************************************************
-    * @description List of valid account record types in mappings.
-    ********************************************************************************************************/
+     * @description List of valid account record types in mappings.
+     ********************************************************************************************************/
     public List<String> validAccRecordTypesInMappings;
 
     /*******************************************************************************************************
-    * @description List of primary affiliation lookup fields names on Contact.
-    ********************************************************************************************************/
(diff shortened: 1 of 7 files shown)


force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -186,76 +199,88 @@ public with sharing class AFFL_MultiRecordTypeMapper {
...
-    * @description Tells caller if Account record type needs Program Enrollment auto-creation.
-    * @param accountRecordType The name of the Account record type as entered in the Affiliation Mappings.
-    * @return Boolean Whether the Account record type has been flagged as needing Academic Record auto-creation.
-    ********************************************************************************************************/
-    public Boolean needsProgramEnrollment(Affiliation__c affl) {
+     * @description Tells caller if Account record type needs Program Enrollment auto-creation.
+     * @param accountRecordType The name of the Account record type as entered in the Affiliation Mappings.
+     * @return Boolean Whether the Account record type has been flagged as needing Academic Record auto-creation.
+     ********************************************************************************************************/
+    public virtual Boolean needsProgramEnrollment(Affiliation__c affl) {
```
MatthewBlanski, line 228: This is one of the two non-prettier changes in this class.

force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -28,74 +28,73 @@
...
-public with sharing class AFFL_MultiRecordTypeMapper {
-
+ * @author Salesforce.org
+ * @date 2015
+ * @group Affiliations
+ * @group-content ../../ApexDocContent/Affiliations.htm
+ * @description Helps with the mapping of Account record types to Account lookup fields on Contact defined
+ *				in the Affiliation Mapping custom settings.
+ */
+public virtual with sharing class AFFL_MultiRecordTypeMapper {
```
MatthewBlanski, line 38: This is one of the two non-prettier changes in this class.

force-app/main/default/classes/AFFL_MultiRecordType_TDTM.cls
```diff
@@ -154,53 +211,102 @@ public class AFFL_MultiRecordType_TDTM extends TDTM_Runnable {
...
+            TDTM_ProcessControl.resetRecursionFlag(
+                TDTM_ProcessControl.registeredTrigger.AFFL_MultiRecordType_TDTM_afflMadePrimary
+            );
+        }
 
         return dmlWrapper;
     }
 
-    public void processAfflDeleted(Affiliation__c afflOld, Contact relatedContact, String lookupFieldAPIName, DmlWrapper dmlWrapper) {
+    /*******************************************************************************************************
```
MatthewBlanski, line 259: Changes on this method include adding apex doc comments and having the method return a value for testability.

force-app/main/default/classes/AFFL_MultiRecordType_TDTM.cls
```diff
@@ -212,31 +318,59 @@ public class AFFL_MultiRecordType_TDTM extends TDTM_Runnable {
...
+                for (Contact contact : relatedContactsList) {
                     relatedContactsMap.put(contact.ID, contact);
                 }
             }
         }
         return relatedContactsMap;
     }
 
-    private void populateKeyAffls(Affiliation__c affl, String lookupFieldAPIName, Contact relatedContact, DmlWrapper dmlWrapper) {
+    /*******************************************************************************************************
```
MatthewBlanski, line 333: Changes on this method include adding apex doc comments and having the method return a value for testability.

force-app/main/default/classes/AFFL_MultiRecordType_TDTM.cls
```diff
@@ -212,31 +318,59 @@ public class AFFL_MultiRecordType_TDTM extends TDTM_Runnable {
...
                 dmlWrapper.objectsToUpdate.add(relatedContact);
+                return relatedContact;
             }
         }
+
+        return null;
     }
 
-    public void uncheckOtherPrimariesSameType(List<Affiliation__c> affls, DmlWrapper dmlWrapper) {
+    /*******************************************************************************************************
```
MatthewBlanski, line 366: Changes on this method include adding apex doc comments and having the method return a value for testability.

force-app/main/default/classes/AFFL_MultiRecordType_TDTM.cls
```diff
@@ -267,17 +410,32 @@ public class AFFL_MultiRecordType_TDTM extends TDTM_Runnable {
         for (Affiliation__c otherAffl : oldPrimaryAffls) {
             otherAffl.Primary__c = false;
         }
-        dmlWrapper.objectsToUpdate.addAll((List<SObject>)oldPrimaryAffls);
+        dmlWrapper.objectsToUpdate.addAll((List<SObject>) oldPrimaryAffls);
+        return oldPrimaryAffls;
     }
 
-    private void createProgramEnrollmentIfNecessary(Affiliation__c affl, DmlWrapper dmlWrapper) {
+    /*******************************************************************************************************
```
MatthewBlanski, line 417: Changes on this method include adding apex doc comments and having the method return a value for testability.

force-app/main/default/classes/AFFL_MultiRecordType_TEST.cls
```diff
@@ -976,4 +983,659 @@ public with sharing class AFFL_MultiRecordType_TEST {
...
     }
+
+    /**************************************************************************
+     * Unit TESTS
+     **************************************************************************/
+
+    /**
+     * @description helper class for initialization of tests for processAfflDeletedValid
+     */
+    private class ProcessAfflDeletedValidData {
```
xinnanlisf, line 994: Is there a reason why we need a helper class other than reusability. Imho the complexity this class is adding overshadows the reusability.

xinnanlisf (approved):
The new tests look good except that they are a bit too complicated since one of the test class has about 1.6k lines. Having inner helper class is fine but I feel like it is a bit overkilling if the only benefit is reusability and we are even implementing it for tdtm classes. Also, having multiple helper class makes the unit test a bit crowded and hard to read. One option is maybe we can extract the "essential" test data structure and let each single test method add their "flavor" on top of it.

A few other minor things really up to you:

- It is ok to add the "autoEnrollment" boolean parameter but might be a bit hard to figure out what that "false" value means when calling the method. We can potentially add a variable and pass it to the method. Also, if we need to add even more parameter to that data models, we will have to refactor all the test methods again. 
- Maybe we want to at least move those helper classes to the bottom of the test class. Potentially a section comment like we did for stub classes too.

force-app/main/default/classes/AFFL_MultiRecordType_TEST.cls
MatthewBlanski: I was having a good bit of trouble balancing the data setup in each class with readability of the tests. I opted for this route to make the data setup defined and then the individual tests more transparent about exactly what they are testing.

I can flatten it out, but the complexity of these methods has required the number of unit tests. I feel like it might become even less readable.

MatthewBlanski: Do you feel like this is a case where we might benefit from separating the test classes up even further? The level of complexity has to do with the functionality. If I were to work this completely, I would refactor it into tiny parts that were much more testable. Let me know how far you feel we should stretch on this one.

xinnanlisf:
Build passed. Merging to parent branch.

PR #1818, opened 2021-07-07
Fixed Auto Enrollment Mappings not working off of Record Type API Names

MatthewBlanski:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
+++ b/force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
@@ -313,7 +313,7 @@ public virtual with sharing class AFFL_MultiRecordTypeMapper {
 
     /*******************************************************************************************************
      * @description Whether or not the affiliation requires a program enrollment to be auto-created.
-     * @param affl The affiliation to check for program enrollment creations
+     * @param affiliation The affiliation to check for program enrollment creations
      * @return Whether or not the affiliation requires a program enrollment to be auto-created.
      ********************************************************************************************************/
     public virtual Boolean needsProgramEnrollment(Affiliation__c affiliation) {
--- a/force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
+++ b/force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
@@ -124,8 +124,6 @@ public virtual with sharing class AFFL_MultiRecordTypeMapper {
             AFFL_MultiRecordTypeMapper.accountRecordTypeInfosByDeveloperName
         );
 
-        this.populateMaps(affiliationMappings);
-
         //Contact Fields
         //List of primary affiliation lookup fields labels/API names from settings.
         List<String> primaryAfflFieldsInMappings = this.primaryAfflFieldInMappingsByAccRecType.values();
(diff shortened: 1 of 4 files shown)


force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -336,29 +319,30 @@ public virtual with sharing class AFFL_MultiRecordTypeMapper {
      * @param accountRecordType The name of the Account record type as entered in the Affiliation Mappings.
```
lifewithryan, line 319: Might wanna update the comments here? looks like the param at some point in time was supposed to be the account record type but now it looks like we are passing in an affiliation.

force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -333,29 +316,30 @@ public virtual with sharing class AFFL_MultiRecordTypeMapper {
      * @param affl The affiliation to check for program enrollment creations
```
xinnanlisf, line 316: We may need to fix the comment here since we have refactored the parameter.

force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -333,29 +316,30 @@ public virtual with sharing class AFFL_MultiRecordTypeMapper {
...
      * @return Whether or not the affiliation requires a program enrollment to be auto-created.
      ********************************************************************************************************/
-    public virtual Boolean needsProgramEnrollment(Affiliation__c affl) {
```
xinnanlisf, line 336: I am a bit worried here since we are changing the logic entirely here. It looked like we were passing affiliation and now affiliation mappings. Just want to double check if this is intended and has no risks. Maybe it was not designed this way but might already became a feature atm.

xinnanlisf (changes requested):

MatthewBlanski:
Resolved. Was an old comment.

force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
MatthewBlanski: This method is still passing in an Affiliation record and acting on it based on abstractions of Affiliation Mappings as configured in the constructor.

I retrofitted this a touch. Originally three maps were made from the Primary Affiliation Settings of the raw account record type value to the three auto enrollment settings. This map used the raw Account Record Type field in the method, meaning it wouldn't work on API names as it was always compared to Affiliation__c.Affiliation_Type__c, which resolves to the. Affiliation's Account's RecordType's Name (a label).

The solve implemented was to instead initialize a map of Primary Affiliation Mappings by their Account Record Type Name (resolving from Developer Name to Name and discarded invalid. mappings in the process). That map is then used in this method to pull its related and relevant fields by using map.get(Affiliation__c.Affiliation_Type__c).

Effectively I ended up trading three inadequate abstractions for one adequate map.

force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
MatthewBlanski: Updated.

force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -165,21 +163,6 @@ public virtual with sharing class AFFL_MultiRecordTypeMapper {
         return labelsnames;
     }
 
-    @testVisible
```
MatthewBlanski, line 168: This is no longer used in the context of this change. These maps were all using the raw values from the affiliation mappings.

force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -330,32 +313,30 @@ public virtual with sharing class AFFL_MultiRecordTypeMapper {
...
     /*******************************************************************************************************
      * @description Whether or not the affiliation requires a program enrollment to be auto-created.
-     * @param affl The affiliation to check for program enrollment creations
+     * @param affiliation The affiliation to check for program enrollment creations
      * @return Whether or not the affiliation requires a program enrollment to be auto-created.
      ********************************************************************************************************/
-    public virtual Boolean needsProgramEnrollment(Affiliation__c affl) {
-        if (affl.Affiliation_Type__c == null) {
+    public virtual Boolean needsProgramEnrollment(Affiliation__c affiliation) {
+        if (affiliation?.Affiliation_Type__c == null) {
```
MatthewBlanski, line 320: Note that we're checking to see if affiliation is null and if Affiliation_Type__c is null in one go here.

force-app/main/default/classes/AFFL_MultiRecordTypeMapper.cls
```diff
@@ -330,32 +313,30 @@ public virtual with sharing class AFFL_MultiRecordTypeMapper {
...
-            return false;
-        }
-
-        Boolean autoEnrollment = accTypeToEnrollCreate.get(affl.Affiliation_Type__c);
+        Affl_Mappings__c affiliationMapping = primaryAffiliationMappingsByAccountRecordTypeName.get(
+            affiliation.Affiliation_Type__c
+        );
 
-        if (autoEnrollment == false) {
+        if (affiliationMapping?.Auto_Program_Enrollment__c != true) {
```
MatthewBlanski, line 328: Note that we're checking to see if the affiliation mapping is null and if its auto enrollment field is not true

xinnanlisf (approved):
This looks good now. Merging it to upper branch.

PR #1822, opened 2021-07-12
Create Product Registry CMT

salvadorsfdc:
Created the Product Registry Custom Metadata Type and fields

# Critical Changes

# Changes

# Issues Closed
WI: [W-9539001](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000PzjWYAS/view)
# New Metadata

Custom Metadata Type:
- Product Registry

Custom Fields:
- Action__c
- Api_Version__c
- Class_Name__c
- Enabled__c
- Namespace__c

Layout:

- Product_Registry__mdt-Product Registry Layout

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/objects/Product_Registry__mdt/fields/Action__c.field-meta.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomField xmlns="http://soap.sforce.com/2006/04/metadata">
+    <fullName>Action__c</fullName>
+    <description>Defines the specific Action or Category this registry refers to.</description>
+    <externalId>false</externalId>
+    <fieldManageability>DeveloperControlled</fieldManageability>
```
iskander-sfdc, line 6: If we create the field as DeveloperControlled it means it can only be changed via package upgrade. Which means new products will be added there via EDA upgrade and admins won't be able to create products on their own, right?

force-app/main/default/objects/Product_Registry__mdt/fields/Enabled__c.field-meta.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomField xmlns="http://soap.sforce.com/2006/04/metadata">
+    <fullName>Enabled__c</fullName>
+    <defaultValue>false</defaultValue>
+    <description>Defines if the custom metadata record is enabled</description>
+    <externalId>false</externalId>
+    <fieldManageability>DeveloperControlled</fieldManageability>
```
iskander-sfdc, line 7: I think this one should be `SubscriberControlled` so admins could disable the class

iskander-sfdc (changes requested):

force-app/main/default/objects/Product_Registry__mdt/fields/Action__c.field-meta.xml
MatthewBlanski: I believe we want developer-controlled on this, but I would appreciate the behavior of a subscriber-created record being validated through 2gp builds. I believe that another package can add additional records there.

force-app/main/default/objects/Product_Registry__mdt/fields/Enabled__c.field-meta.xml
MatthewBlanski: This is one of the cases where SubscriberControlled is relevant. Leaving this open will give CCE a way to override behavior in the case of an emergency and effectively temporarily exclude a malfunctioning package, which is the purpose of this field.

salvadorsfdc: makes total sense, thanks

force-app/main/default/objects/Product_Registry__mdt/fields/Action__c.field-meta.xml
salvadorsfdc: ok, so leaving it DeveloperControlled. Thanks

iskander-sfdc (approved):

juliesnow (approved):
Edits that we approved in quip verified here. Thanks!

spinkelman:
QA Passes - Merging

PR #1823, opened 2021-07-12
[W-9393925, W-9495251] New date fields on application

vmarasow:
# Changes
## New Fields on the Application Object
To make it easy for you to track the date when an applicant is able to submit an application and the date by which they must submit one, we added the Open Date and Due Date fields to the Application object.

# Issues Closed
[W-9393925](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000NwjVYAS/view) - EDA
[W-9384949](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000Nt64YAC/view) - R&A

# New Metadata
## Fields
- Application -> Open Date
- Application -> Due Date

# Changed Metadata
## Layout
- EDA Application Layout

# Testing Notes

PR #1825, opened 2021-07-14
[W-9495251] Validate the open/due date range 

vmarasow:
I merged #1823 into this branch as it has a dependency on the new fields created there. As such, this PR should be merged after #1823.

# Changes
- We added a trigger to validate that an application due date is after its open date

# Issues Closed
[W-9495251](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000PACVYA4/view) - R&A

# New Metadata
## Apex
- Application_TDTM
- Application_TEST

## Custom Label
- stgErrorAppDueDate

# Testing Notes


RadSFDeveloper (changes requested):
Nice work @vmarasow . Below are comments to update trigger and unit tests for readability and descriptive names/messages. Please let me know if there is any question.

force-app/main/default/classes/Application_TDTM.cls
```diff
@@ -0,0 +1,77 @@
...
+                    }
+                }
+                if (application.Open_Date__c >= application.Due_Date__c) {
+                    application.addError(Label.stgErrorAppDueDate);
+                }
+            }
+        }
+
+        return new DmlWrapper();
+    }
```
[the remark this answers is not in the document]
vmarasow: 1. The validation is only happening in the last if statement, all the other conditionals are excluding records from being validated. Therefore, if I was to split the validation logic into a separate method, I would only want to extract the last if statement, and I think that's overkill.
2. I took the `oldMap` and `Label.*` convention from other trigger files in EDA, and I think that being consistent would be more valuable in this case, since these are widely used conventions.

MatthewBlanski (changes requested):
Following up on this, I would prefer we follow a logical pattern that we are using in new code for EDA Triggers. Similar to the partial implementation here: https://github.com/SalesforceFoundation/EDA/blob/main/force-app/main/default/classes/CON_Preferred_TDTM.cls

This is covered in our EDA TDTM Trigger Handler Pattern document.

Generally the logic flow is preferably similar to this:

Specifically I would prefer we see a method for:

1. The specific trigger context
2. Evaluating the relevance on each record (and adding it to a work to perform list in the trigger context method)
3. Performing work on the relevant list
4. Perform work for the individual record in the list.

force-app/main/default/classes/Application_TDTM.cls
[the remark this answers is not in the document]
RadSFDeveloper: 1. I understand it might look like another way to write the same thing, but it is more readable if the code is extracted into a meaningful method with descriptive name. Moreover when/if we change the logic or extend behaviour, we will be extended the code while still keeping it readable if we start with clean base.
2. I understand the variable naming take take from the existing/old naming convention, thus I kept the `run()` method arguments the same. However, since this is a new trigger and new code, let's follow the best Apex coding style as of today (it may/will change in future and that is ok). Variable naming is very important to convey the behaviour as well reduce bugs. Previous NPSP/EDA naming convention contained data type in variables and I would like to reduce those occurrences.

force-app/main/default/classes/Application_TDTM.cls
```diff
@@ -0,0 +1,129 @@
...
+     ********************************************************************************************************/
+    @TestVisible
+    private void handleBeforeUpdate(List<SObject> oldSObjectList, List<SObject> newSObjectList) {
+        List<Application__c> updatedApplications = (List<Application__c>) newSObjectList;
+        Map<Id, Application__c> oldApplicationsById = new Map<Id, Application__c>(
+            (List<Application__c>) oldSObjectList
+        );
+
+        for (Application__c updatedApplication : updatedApplications) {
+            Application__c oldApplication = (Application__c) oldApplicationsById.get(updatedApplication.Id);
```
MatthewBlanski, line 97: We no longer need to cast this given that we are using a strongly-typed map
```suggestion
            Application__c oldApplication = oldApplicationsById.get(updatedApplication.Id);
```

force-app/main/default/classes/Application_TEST.cls
```diff
@@ -0,0 +1,151 @@
...
+ * @group-content ../../ApexDocContent/Contacts.htm
+ * @description Test class for logic that prevents application due date to be prior to the open date.
+ */
+@isTest
+private with sharing class Application_TEST {
+    @TestSetup
+    static void makeData() {
+        List<Account> academicAccount = UTIL_UnitTestData_TEST.getMultipleTestAccounts(
+            1,
+            UTIL_Describe_API.getAcademicAccRecTypeID()
```
MatthewBlanski, line 43: A minor change. I would prefer we not call out to the global APIs in these cases and rely on internal methods.
```suggestion
            UTIL_Describe.getAcademicAccRecTypeID()
```

force-app/main/default/classes/Application_TEST.cls
```diff
@@ -0,0 +1,151 @@
...
+
+        Application__c openEndedApplication = new Application__c(
+            Applicant__c = student.Id,
+            Applying_To__c = academicAccount.Id,
+            Open_Date__c = System.today()
+        );
+        insert openEndedApplication;
+
+        Test.startTest();
+        openEndedApplication.Due_Date__c = System.today().addDays(-10);
```
MatthewBlanski, line 97: This is completely a nitpick and is not required. Given that the test block is used to create a separate set of limits and we can run performance metrics on that, anything that is not directly the test likely should remain outside the block.

```suggestion
        openEndedApplication.Due_Date__c = System.today().addDays(-10);

        Test.startTest();
```

force-app/main/default/classes/Application_TEST.cls
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+
+    private static Account getAcademicAccount() {
+        return [SELECT Id FROM Account LIMIT 1];
+    }
+
+    private static Contact getStudent() {
+        return [SELECT Id FROM Contact LIMIT 1];
+    }
```
MatthewBlanski, line 150: My experience with other tests is that these queries are not sufficiently selective and should use the Id of the Account or Contact generated. This can cause an issue when tests are run in parallel due to the joys of platform quirks.

I would prefer that we be selective in these cases. A different approach would be to insert the Contact and Account in these methods and return them rather than use the test setup. In those cases we are guaranteed to be working with the same records.

This is not a required change.

force-app/main/default/classes/Application_TEST.cls
```diff
@@ -0,0 +1,151 @@
...
+            updateResult.getErrors()[0].getMessage(),
+            'Should fail with the Open/Due Date validation error message'
+        );
+    }
+
+    /********************************************************
+     * @description Should succeed to insert an application with just one date set
+     *********************************************************/
+    @isTest
+    static void shouldSucceedWhenApplicationHasEitherDueOrOpenDate() {
```
MatthewBlanski, line 113: This test has a mixed context and is testing two different things. I would prefer that it test the individual behaviors separately so we have explicit and concise tests indicating exactly what behavior is functioning or failing.

This is not a required change, but I feel it would increase Trust.

MatthewBlanski (changes requested):
The code is functional, but we are performing an unnecessary cast in the TDTM class. The test class has some changes suggested in it, though none of them are required.

MatthewBlanski (approved):

RadSFDeveloper (approved):
Looks good to me. Great work @vmarasow !

spinkelman:
QA for both AC and EDA teams complete and passes. Merging

PR #1831, opened 2021-07-21
inactive address metadata better unit test

xinnanlisf:
- Added missing test class for ADDR_Account_TDTM. Mostly placeholder for now as we are not changing the tdtm class.
- Added missing test class for ADDR_Addresses_UTIL. 
-- We have some missing null pointer guard clause and thus those methods are commented out.
-- Some methods are not virtual and as a result chained methods are really hard to test against. Pending for refactoring. e.g. state picklist enabled or not.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+
+        System.assertEquals(false, isAddressEmpty);
+    }
+
+    /**************************************************************************************************************************
+     ****************************************************** isContactAccAddrInfoSame ******************************************
+     **************************************************************************************************************************/
+    @isTest
+    static void isContactAccAddrInfoSameTrue() {
+        Contact contact = UTIL_UnitTestData_TEST.getContact();
```
MatthewBlanski, line 84: Good use of not making DML updates here to let this test run a bit faster.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Test.startTest();
+        Boolean isContactAccAddrInfoSame = ADDR_Addresses_UTIL.isContactAccAddrInfoSame(contact, accounts[0]);
+        Test.stopTest();
+
+        System.assertEquals(false, isContactAccAddrInfoSame);
+    }
+
+    @isTest
+    static void isContactAccAddrInfoSameFalseDifferentStateCode() {
+        /*Almost no way we can test state code/country code here
```
MatthewBlanski, line 115: Given this use case, make a note for TODO at the appropriate locations in ADDR_Addresses_UTIL so we surface each of these use cases for posterity.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Test.startTest();
+        Boolean isContactAccAddrInfoSame = ADDR_Addresses_UTIL.isContactAccAddrInfoSame(null, accounts[0]);
+        Test.stopTest();
+
+        System.assertEquals(false, isContactAccAddrInfoSame);
+    }
+
+    @isTest
+    static void isContactAccAddrInfoSameNullContactNullAccount() {
+        //We don't have guard clause for the scenario when both contact and account are null
```
MatthewBlanski, line 147: If this is the case, make a test for this that covers the scenario and asserts for the current behavior. Then make a comment in the ADDR_Addresses_UTIL to highlight the problem scenario.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Boolean isContactAccAddrInfoSame = ADDR_Addresses_UTIL.isContactAccAddrInfoSame(null, null);
+        Test.stopTest();
+
+        System.assertEquals(true, isContactAccAddrInfoSame);*/
+    }
+
+    /**************************************************************************************************************************
+     ****************************************************** isAddressChanged **************************************************
+     **************************************************************************************************************************/
+    @isTest
```
MatthewBlanski, line 161: This is one of those use cases where technically we should be validating every potential scenario for address changing. This is why I tend to make a test and assertion helper for those to make the methods really small. We don't have to go that far this time, but it's worth noting.

Given that we're using case sensitivity, it might be worth changing the case for the last letter of the address to really capture that kind of change.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Test.startTest();
+        Boolean isContactAddressChanged = ADDR_Addresses_UTIL.isContactAddressChanged(con1, con2);
+        Test.stopTest();
+
+        System.assertEquals(true, isContactAddressChanged);
+    }
+
+    @isTest
+    static void testIsContactAddressChanged() {
+        /* return (!isContactAddressEmpty(con1) || String.isNotBlank(con1.Primary_Address_Type__c)); does not make any sense
```
MatthewBlanski, line 303: This is a distinct use case for test coverage. So we have 4 use cases for con1 or con2 being null:

1. `isContactAddressEmpty(con1)` returns true
2. `isContactAddressEmpty(con1)` returns false
3. `String.isNotBlank(con1.Primary_Address_Type__c)` returns true
4. `String.isNotBlank(con1.Primary_Address_Type__c)` returns false

So explicitly cover these scenarios.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        );
+        accounts[0].BillingStreet = 'some other billing street';
+
+        Test.startTest();
+        Boolean isAccountAddressChanged = ADDR_Addresses_UTIL.isAccountAddressChanged(accounts[0], accounts[1]);
+        Test.stopTest();
+
+        System.assertEquals(true, isAccountAddressChanged);
+    }
+
```
MatthewBlanski, line 360: We don't have coverage on the state and country code picklists, and we can only test for when they're turned off, based on org config when we get the ability to override. So make a note in the class about the testing gap.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Test.startTest();
+        Boolean isAccountAddressChanged = ADDR_Addresses_UTIL.isAccountAddressChanged(accounts[0], accounts[1]);
+        Test.stopTest();
+
+        System.assertEquals(false, isAccountAddressChanged);
+    }
+
+    @isTest
+    static void testIsAccountAddressChangedNotNullDifferentAddressInfoTrue() {
+        List<Account> accounts = UTIL_UnitTestData_TEST.getMultipleTestAccounts(
```
MatthewBlanski, line 348: In an ideal world, we'd have a test for each field that is also involving case sensitivity. I would advise changing the test to change the case on one character of the target field to capture that behavior so we don't inadvertently break it.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+
+        Test.startTest();
+        Boolean isContactAddressEmpty = ADDR_Addresses_UTIL.isContactAddressEmpty(contact);
+        Test.stopTest();
+
+        System.assertEquals(true, isContactAddressEmpty);
+    }
+
+    @isTest
+    static void testIsContactAddressEmptyNotNullNotEmptyFalse() {
```
MatthewBlanski, line 385: Again with in an ideal world we would go field by field for this to be explicit about what is expected here.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Boolean stringEquals = ADDR_Addresses_UTIL.equalsCaseSensitive('test string', '!test string');
+        Test.stopTest();
+
+        System.assertEquals(false, stringEquals);
+    }
+
+    /**************************************************************************************************************************
+     ****************************************************** isAddressMatchFuzzy ***********************************************
+     **************************************************************************************************************************/
+    //TODO: The method is not test visible.
```
MatthewBlanski, line 482: Making a method test visible is fine as it doesn't affect functionality outside of tests. Go ahead and do it then write the tests. This is especially important since this is one that's going to be critical for the upcoming work.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Boolean isAddressMatchFuzzy = ADDR_Addresses_UTIL.isAddressMatchFuzzy(null, null, null);
+        Test.stopTest();
+
+        System.assertEquals(false, isAddressMatchFuzzy);
+    }*/
+
+    /**************************************************************************************************************************
+     ****************************************************** strCleanup ********************************************************
+     **************************************************************************************************************************/
+    //TODO: The method is not test visible.
```
MatthewBlanski, line 495: Go ahead and make it test visible then write the tests.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+
+    /**************************************************************************************************************************
+     ****************************************************** copyAddressObj ****************************************************
+     **************************************************************************************************************************/
+    @isTest
+    static void testCopyAddressObjBothNull() {
+        //To Refactor: we don't have guard clause for this.
```
MatthewBlanski, line 502: Regardless, capture the expected behavior and assert for it, making a note on the method.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        /*Test.startTest();
+        ADDR_Addresses_UTIL.copyAddressObj(null, null);
+        Test.stopTest();*/
+
+        //Expecting null pointer
+    }
+
+    @isTest
+    static void testCopyAddressObjSourceNull() {
+        //To Refactor: we don't have guard clause for this.
```
MatthewBlanski, line 512: Capture the expected behavior and assert for it.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Test.startTest();
+        ADDR_Addresses_UTIL.copyAddressObj(addressSource, null);
+        Test.stopTest();*/
+
+        //Expecting null pointer
+    }
+
+    @isTest
+    static void testCopyAddressObjDestinationNull() {
+        //To Refactor: we don't have guard clause for this.
```
MatthewBlanski, line 524: Capture the expected behavior and assert for it.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        System.assertEquals(addressSource.Geolocation__Longitude__s, addressDestination.Geolocation__Longitude__s);
+        System.assertEquals('Work', addressDestination.Address_Type__c);
+    }
+
+    /**************************************************************************************************************************
+     ****************************************************** copyAddressStdSObj ************************************************
+     **************************************************************************************************************************/
+    @isTest
+    static void testCopyAddressObjSobjSourceRecordNull() {
+        //To Refactor: we don't have guard clause for this.
```
MatthewBlanski, line 586: Regardless, capture the expected behavior and assert for it to ensure future refactors work as expected. Then make a note in the method. Also about the terrible method name.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Test.startTest();
+        ADDR_Addresses_UTIL.copyAddressStdSObj(null, 'Mailing', account, 'Billing');
+        Test.stopTest();*/
+
+        //Expecting null pointer
+    }
+
+    @isTest
+    static void testCopyAddressObjSobjSourceFieldPrefixNull() {
+        //To Refactor: we don't have guard clause for this.
```
MatthewBlanski, line 600: Likewise here with capturing the expected behavior

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Test.startTest();
+        ADDR_Addresses_UTIL.copyAddressStdSObj(contact, 'Mailing', null, 'Billing');
+        Test.stopTest();*/
+
+        //Expecting null pointer
+    }
+
+    @isTest
+    static void testCopyAddressObjSobjDestinationFieldPrefixNull() {
+        //To Refactor: we don't have guard clause for this.
```
MatthewBlanski, line 625: Likewise here with capturing the expected behavior

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        ADDR_Addresses_UTIL.copyAddressStdSObj(contact, 'Mailing', account, null);
+        Test.stopTest();*/
+
+        //Expecting invalid field error
+    }
+
+    @isTest
+    static void testCopyAddressObjSobjValidField() {
+        //To Refactor: we don't have guard clause for this.
+        //To Refactor: No way we can test isStateCountryPicklistsEnabled.
```
MatthewBlanski, line 639: Make a note in the class

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        System.assertEquals(account.BillingPostalCode, contact.MailingPostalCode);
+        System.assertEquals(account.BillingState, contact.MailingState);
+        System.assertEquals(account.BillingCountry, contact.MailingCountry);
+        System.assertEquals(account.BillingLatitude, contact.MailingLatitude);
+        System.assertEquals(account.BillingLongitude, contact.MailingLongitude);
+    }
+
+    @isTest
+    static void testCopyAddressObjSobjInvalidField() {
+        //To Refactor: we don't have guard clause for this.
```
MatthewBlanski, line 659: Likewise here with capturing expected behavior

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Test.stopTest();*/
+
+        //Expecting invalid field error
+    }
+
+    /**************************************************************************************************************************
+     ****************************************************** initValidStateCountryCodes ****************************************
+     **************************************************************************************************************************/
+
+    //This method is not quite testable as it is using schema describe.
```
MatthewBlanski, line 827: Make a note in the actual class. Agreed on testability. This may be a good scenario for an unmanaged unit test.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1341 @@
...
+        Test.stopTest();
+
+        System.assertEquals('test', multilineStreet);
+    }
+
+    /**************************************************************************************************************************
+     ****************************************************** isStateCountryPicklistsEnabled ************************************
+     **************************************************************************************************************************/
+
+    //At this moment, it is really hard to test this method. We may consider make the methods virtual to override.
```
MatthewBlanski, line 909: Write a test for what we can know. Don't be afraid to make multiple tests that will exit early if the org state isn't what we need.

MatthewBlanski (changes requested):
There are some methods where we can have meaningful coverage to test current and existing behavior. Writing even partial unit tests for these will still bring value when we go for the refactor. Identify those places were are going to change in the coming work and make certain we have tests for those methods.

force-app/main/default/classes/ADDR_Account_TEST.cls
```diff
@@ -0,0 +1,120 @@
...
+    //This test class stays as placeholder for future refactoring as we are not making changes to the ADDR_Account_TDTM atm.
+
+    /**************************************************************************************************************************
+     ****************************************************** Recursion Flags ***************************************************
+     **************************************************************************************************************************/
+
+    @isTest
+    static void testTurnOffRecursionFlag() {
+        TDTM_ProcessControl.resetRecursionFlag(TDTM_ProcessControl.registeredTrigger.ADDR_Account_TDTM_After_Insert);
+        TDTM_ProcessControl.resetRecursionFlag(TDTM_ProcessControl.registeredTrigger.ADDR_Account_TDTM_After_Update);
```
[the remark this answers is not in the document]
xinnanlisf: My understanding is that static variables will retain its values among tests. Resetting them ensures us another layer of testing isolation.

force-app/main/default/classes/ADDR_Account_TEST.cls
```diff
@@ -0,0 +1,120 @@
...
+ */
+@isTest
+private with sharing class ADDR_Account_TEST {
+    //This test class stays as placeholder for future refactoring as we are not making changes to the ADDR_Account_TDTM atm.
+
+    /**************************************************************************************************************************
+     ****************************************************** Recursion Flags ***************************************************
+     **************************************************************************************************************************/
+
+    @isTest
```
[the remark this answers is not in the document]
xinnanlisf: I did this on purpose. I found those comments not very useful. Test method name should be good enough to tell us what the test is covering. Longer test method names are ok imo as we are not referencing it anyway. If we find the test method name too long, it usually means the methods are handling too many things.

force-app/main/default/classes/ADDR_Account_TEST.cls
```diff
@@ -0,0 +1,120 @@
...
+            false,
+            TDTM_ProcessControl.getRecursionFlag(TDTM_ProcessControl.registeredTrigger.ADDR_Account_TDTM_After_Update)
+        );
+    }
+
+    /**************************************************************************************************************************
+     ****************************************************** Main Method *******************************************************
+     **************************************************************************************************************************/
+    @isTest
+    static void testRunDMLWrapperSingleObjectType() {
```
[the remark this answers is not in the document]
xinnanlisf: This test class serves as a placeholder for now. We can work on it after future refactoring on account tdtm. e.g. breaking it into smaller methods.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
xinnanlisf: I came with the idea that we could add refactoring comments to test class:  1) Hoping that whenever we are refactoring any function, we go to the related unit tests first 2) We don't add TODO comments to our functional code so that it is nice and clean.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
xinnanlisf: You mean catch the null pointer and assert the exception type?

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
xinnanlisf: Yeah this is a mistake that i forgot to remove the test method. All the scenarios you mentioned were covered by test methods above.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
xinnanlisf: I left the comments in the unit test class.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
xinnanlisf: On it.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
xinnanlisf: Agreed. maybe something we can refactor in the functional class so that we don't have to repeat the "address fields" again and again.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
xinnanlisf: On it.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
xinnanlisf: On it.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
xinnanlisf: I tend to ignore the state picklist function at this moment since we won't touch them. Not ideal but refactoring it feels like a heavier lift and riskier.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
MatthewBlanski: I'm comfortable with that approach.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
MatthewBlanski: Yes

force-app/main/default/classes/ADDR_Addresses_UTIL.cls
```diff
@@ -28,14 +28,13 @@
...
-* @group-content ../../ApexDocContent/Addresses.htm
-* @description Addresses utility methods.
-*/
+ * @author Salesforce.org
+ * @date 2016
+ * @group Addresses
+ * @group-content ../../ApexDocContent/Addresses.htm
+ * @description Addresses utility methods.
+ */
 public with sharing class ADDR_Addresses_UTIL {
```
MatthewBlanski, line 37: Really happy to see this getting prettier applied to it so that meaningful changes are captured in other PRs and aren't lost in the shuffle on this one.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+     ****************************************************** copyAddressObj ****************************************************
+     **************************************************************************************************************************/
+    @isTest
+    static void testCopyAddressObjBothNull() {
+        //To Refactor: we don't have guard clause for this.
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressObj(null, null);
+            Test.stopTest();
+        } catch (NullPointerException e) {
```
MatthewBlanski, line 753: We'll want an assert here so we know the behavior changed when we fix this.
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
        } catch (NullPointerException e) {
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    @isTest
+    static void testCopyAddressObjSourceNull() {
+        //To Refactor: we don't have guard clause for this.
+        Address__c addressSource = UTIL_UnitTestData_TEST.getMultipleTestAddresses(1)[0];
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressObj(addressSource, null);
+            Test.stopTest();
+        } catch (NullPointerException e) {
```
MatthewBlanski, line 771:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
        } catch (NullPointerException e) {
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+
+    @isTest
+    static void testCopyAddressObjDestinationNull() {
+        //To Refactor: we don't have guard clause for this.
+        Address__c addressDestination = UTIL_UnitTestData_TEST.getMultipleTestAddresses(1)[0];
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressObj(null, addressDestination);
+            Test.stopTest();
```
MatthewBlanski, line 788:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    static void testCopyAddressObjSobjSourceRecordNull() {
+        //To Refactor: we don't have guard clause for this.
+        //To Refactor: terrible method name.
+        //To Refactor: Have to define the Address prefix is a loophole
+        Account account = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe_API.getBizAccRecTypeID())[0];
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressStdSObj(null, 'Mailing', account, 'Billing');
+            Test.stopTest();
```
MatthewBlanski, line 858:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    @isTest
+    static void testCopyAddressObjSobjSourceFieldPrefixNull() {
+        //To Refactor: we don't have guard clause for this.
+        Contact contact = UTIL_UnitTestData_TEST.getContact();
+        Account account = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe_API.getBizAccRecTypeID())[0];
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressStdSObj(contact, null, account, 'Billing');
+            Test.stopTest();
```
MatthewBlanski, line 877:
```suggestion
            Test.stopTest();
            System.assert(false,'An SObject Exception should be thrown as we do not handle it in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    @isTest
+    static void testCopyAddressObjSobjDestinationFieldPrefixNull() {
+        //To Refactor: we don't have guard clause for this.
+        Contact contact = UTIL_UnitTestData_TEST.getContact();
+        Account account = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe_API.getBizAccRecTypeID())[0];
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressStdSObj(contact, 'Mailing', account, null);
+            Test.stopTest();
```
MatthewBlanski, line 914:
```suggestion
            Test.stopTest();
            System.assert(false,'An SObject Exception should be thrown as we do not handle it in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    @isTest
+    static void testCopyAddressObjSobjInvalidField() {
+        //To Refactor: we don't have guard clause for this.
+        Contact contact = UTIL_UnitTestData_TEST.getContact();
+        Account account = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe_API.getBizAccRecTypeID())[0];
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressStdSObj(account, 'Mailing', contact, 'Billing');
+            Test.stopTest();
```
MatthewBlanski, line 953:
```suggestion
            Test.stopTest();
            System.assert(false,'An SObject Exception should be thrown as we do not handle it in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+     **************************************************************************************************************************/
+    @isTest
+    static void testCopyAddressObjSobjAddrAddressNull() {
+        //To Refactor: we don't have guard clause for this.
+        Account account = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe_API.getBizAccRecTypeID())[0];
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressStdSObjAddr(account, 'Billing', null, 'Home');
+            Test.stopTest();
```
MatthewBlanski, line 974:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    @isTest
+    static void testCopyAddressObjSobjAddrSourceFieldPrefixNull() {
+        //To Refactor: we don't have guard clause for this.
+        Account account = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe_API.getBizAccRecTypeID())[0];
+        Address__c address = UTIL_UnitTestData_TEST.getMultipleTestAddresses(1)[0];
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressStdSObjAddr(account, null, address, 'Home');
+            Test.stopTest();
```
MatthewBlanski, line 993:
```suggestion
            Test.stopTest();
            System.assert(false,'An SObject Exception should be thrown as we do not handle it in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    }
+
+    @isTest
+    static void testIsContactAccAddrInfoSameNullContactNullAccount() {
+        //TODO(Refactor): add guard clause for contact == null $$ account == null
+
+        try {
+            Test.startTest();
+            Boolean isContactAccAddrInfoSame = ADDR_Addresses_UTIL.isContactAccAddrInfoSame(null, null);
+            Test.stopTest();
```
MatthewBlanski, line 152:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+
+    @isTest
+    static void testIsAddressMatchFuzzyIncludeAddressTypeNullFalse() {
+        List<Address__c> addresses = UTIL_UnitTestData_TEST.getMultipleTestAddresses(2);
+
+        //To Refactor: add guard clause.
+        try {
+            Test.startTest();
+            Boolean isAddressMatchFuzzy = ADDR_Addresses_UTIL.isAddressMatchFuzzy(addresses[0], addresses[1], null);
+            Test.stopTest();
```
MatthewBlanski, line 526:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+
+    @isTest
+    static void testCopyAddressObjSobjDestinationRecordNull() {
+        //To Refactor: we don't have guard clause for this.
+        Contact contact = UTIL_UnitTestData_TEST.getContact();
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressStdSObj(contact, 'Mailing', null, 'Billing');
+            Test.stopTest();
```
MatthewBlanski, line 895:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+
+    @isTest
+    static void testCopyAddressObjSobjAddrSourceRecordNull() {
+        //To Refactor: we don't have guard clause for this.
+        Address__c address = UTIL_UnitTestData_TEST.getMultipleTestAddresses(1)[0];
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressStdSObjAddr(null, 'Billing', address, 'Home');
+            Test.stopTest();
```
MatthewBlanski, line 1011:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    @isTest
+    static void testCopyAddressObjSobjAddrInvalidField() {
+        //To Refactor: we don't have guard clause for this.
+        Account account = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe_API.getBizAccRecTypeID())[0];
+        Address__c address = new Address__c();
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressStdSObjAddr(account, 'Mailing', address, 'Home');
+            Test.stopTest();
```
MatthewBlanski, line 1050:
```suggestion
            Test.stopTest();
            System.assert(false,'An SObject Exception should be thrown as we do not handle it in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    @isTest
+    static void testCopyAddressAddrSObjAddressNull() {
+        //To Refactor: we don't have guard clause for this.
+        Account account = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe_API.getBizAccRecTypeID())[0];
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressAddrSObj(null, account, 'Billing', 'Primary_Address_Type__c', dmlWrapper);
+            Test.stopTest();
```
MatthewBlanski, line 1072:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    @isTest
+    static void testCopyAddressAddrSObjSourceRecordNull() {
+        //To Refactor: we don't have guard clause for this.
+        Address__c address = UTIL_UnitTestData_TEST.getMultipleTestAddresses(1)[0];
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressAddrSObj(address, null, 'Billing', 'Primary_Address_Type__c', dmlWrapper);
+            Test.stopTest();
```
MatthewBlanski, line 1091:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    static void testCopyAddressAddrSObjSourceRecordPrefixNull() {
+        //To Refactor: we don't have guard clause for this.
+        Address__c address = UTIL_UnitTestData_TEST.getMultipleTestAddresses(1)[0];
+        Account account = UTIL_UnitTestData_TEST.getMultipleTestAccounts(1, UTIL_Describe_API.getBizAccRecTypeID())[0];
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressAddrSObj(address, account, null, 'Primary_Address_Type__c', dmlWrapper);
+            Test.stopTest();
```
MatthewBlanski, line 1111:
```suggestion
            Test.stopTest();
            System.assert(false,'An SObject Exception should be thrown as we do not handle it in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    static void testCopyAddressAddrSObjInvalidSourceRecordField() {
+        //To Refactor: we don't have guard clause for this.
+        Address__c address = UTIL_UnitTestData_TEST.getMultipleTestAddresses(1)[0];
+        Contact contact = new Contact();
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.copyAddressAddrSObj(address, contact, 'Billing', 'Primary_Address_Type__c', dmlWrapper);
+            Test.stopTest();
```
MatthewBlanski, line 1154:
```suggestion
            Test.stopTest();
            System.assert(false,'An SObject Exception should be thrown as we do not handle it in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    /**************************************************************************************************************************
+     ****************************************************** handleMultilineStreet *********************************************
+     **************************************************************************************************************************/
+    @isTest
+    static void testHandleMultilineStreetNull() {
+        //To Refactor: we don't have guard clause for this.
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.handleMultilineStreet(null);
+            Test.stopTest();
```
MatthewBlanski, line 1179:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    /**************************************************************************************************************************
+     ****************************************************** getMultilineStreet ************************************************
+     **************************************************************************************************************************/
+    @isTest
+    static void testGetMultilineStreetNull() {
+        //To Refactor: we don't have guard clause for this.
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.getMultilineStreet(null);
+            Test.stopTest();
```
MatthewBlanski, line 1224:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    /**************************************************************************************************************************
+     ****************************************************** getParentIdToAddrs ************************************************
+     **************************************************************************************************************************/
+    @isTest
+    static void testGetParentIdToAddrsNull() {
+        //To Refactor: we don't have guard clause for this.
+        try {
+            Test.startTest();
+            Map<ID, List<Address__c>> parentIdToAddrs = ADDR_Addresses_UTIL.getParentIdToAddrs(null);
+            Test.stopTest();
```
MatthewBlanski, line 1285:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    /**************************************************************************************************************************
+     ****************************************************** getMapParentIdListAddrs *******************************************
+     **************************************************************************************************************************/
+    @isTest
+    static void testGetMapParentIdListAddrsNull() {
+        //To Refactor: we don't have guard clause for this.
+        try {
+            Test.startTest();
+            Map<ID, List<Address__c>> parentIdToAddrs = ADDR_Addresses_UTIL.getMapParentIdListAddrs(null);
+            Test.stopTest();
```
MatthewBlanski, line 1357:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+     **************************************************************************************************************************/
+    @isTest
+    static void testUpdateAddrIsOverrideMapNull() {
+        //To Refactor: we don't have guard clause for this.
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.updateAddrIsOverride(null, dmlWrapper);
+            Test.stopTest();
```
MatthewBlanski, line 1446:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+                mapAddrIdIsOverride.put(addresses[i].Id, true);
+            } else {
+                mapAddrIdIsOverride.put(addresses[i].Id, false);
+            }
+        }
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.updateAddrIsOverride(mapAddrIdIsOverride, null);
+            Test.stopTest();
```
MatthewBlanski, line 1484:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    @isTest
+    static void testUncheckDefaultOtherAddrsNewAddrNull() {
+        //To Refactor: we don't have guard clause for this.
+        List<Address__c> addresses = UTIL_UnitTestData_TEST.getMultipleTestAddresses(10);
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.uncheckDefaultOtherAddrs(null, addresses, dmlWrapper);
+            Test.stopTest();
```
MatthewBlanski, line 1552:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+    @isTest
+    static void testUncheckDefaultOtherAddrsAllAddrsNull() {
+        //To Refactor: we don't have guard clause for this.
+        List<Address__c> addresses = UTIL_UnitTestData_TEST.getMultipleTestAddresses(1);
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.uncheckDefaultOtherAddrs(addresses[0], null, dmlWrapper);
+            Test.stopTest();
```
MatthewBlanski, line 1571:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -0,0 +1,1731 @@
...
+        for (Integer i = 0; i < addresses.size(); i++) {
+            addresses[i].Parent_Account__c = accounts[i].Id;
+        }
+        insert addresses;
+        TDTM_Runnable.DmlWrapper dmlWrapper = new TDTM_Runnable.DmlWrapper();
+
+        try {
+            Test.startTest();
+            ADDR_Addresses_UTIL.uncheckDefaultOtherAddrs(addresses[0], addresses, null);
+            Test.stopTest();
```
MatthewBlanski, line 1600:
```suggestion
            Test.stopTest();
            System.assert(false,'A Null Pointer Exception should be thrown as we do not handle nulls in the method.');
```

MatthewBlanski (changes requested):
One set of changes. Tests in which we expect exceptions should fail if the exception does not occur. I've left a good amount of suggestions to incorporate into the code to validate that that behavior is occurring. Without those asserts, the unit tests would provide false positives and pass whether the exception was thrown or not.

MatthewBlanski (approved):

PR #1840, opened 2021-07-29
Add managed email field for credentialing to Account

MatthewBlanski:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/force-app/main/default/labels/CustomLabels.labels-meta.xml
+++ b/force-app/main/default/labels/CustomLabels.labels-meta.xml
@@ -663,14 +663,6 @@
         <value>Account Model</value>
     </labels>
     <labels>
-        <fullName>stgAccountRecordTypeGroupLabelTitle</fullName>
-        <categories>Settings</categories>
-        <language>en_US</language>
-        <protected>true</protected>
-        <shortDescription>Group label for the Account record type selection</shortDescription>
-        <value>Select Account Record Types</value>
-    </labels>
-    <labels>
         <fullName>stgAccountModelSettingsTitle</fullName>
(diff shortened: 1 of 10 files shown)


force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -662,14 +662,6 @@
         <shortDescription>Navigation title for Account Model</shortDescription>
         <value>Account Model</value>
     </labels>
-    <labels>
```
xinnanlisf, line 665: Looks like there are tons of label changes here mostly repositionings. Just want to make sure if they are on purpose.

xinnanlisf (approved):
looks good! very minor questions:

1) Do we want those label changes?
2) Are we adding the new field to the layout?

force-app/main/default/labels/CustomLabels.labels-meta.xml
MatthewBlanski: They are quite deliberate. We fell out of line with the default ordering in the labels file, so this is restoring that order.

MatthewBlanski:
> looks good! very minor questions:
> 
> 1. Do we want those label changes?
> 2. Are we adding the new field to the layout?

We do want the label readjustment. We should be keeping up with those in each PR, but fell out of line somehow. When I pulled down translations, it reordered them to the MDAPI expected order.

We are deliberately not adding this to a layout as we do not have an Educational Institution layout at this time.

spinkelman:
QA passes - Merging

PR #1842, opened 2021-08-03
Incorrect new/old Contact matching causing added cycles in TDTM

xinnanlisf:
Used a map to pass matching oldContact and process update data. 

Dev testing confirmed that update trigger wouldn't falsely run anymore.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/force-app/main/default/classes/ACCT_IndividualAccounts_TDTM.cls
+++ b/force-app/main/default/classes/ACCT_IndividualAccounts_TDTM.cls
@@ -246,26 +246,42 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
         List<Contact> contactsNeedAccounts = new List<Contact>();
         HandleUpdateWrapperLogic updateWrapper;
         HandleInsertWrapperLogic insertWrapper;
-        Map<Id, Contact> oldContactByContactId = new Map<Id, Contact>(oldContacts);
         Integer conProcessed = 0;
 
-        for (Contact con : newContacts) {
-            //Account needs to be created if (1) Account does not exist or (2) Account model is not blank
-            if (con.AccountId == null && defaultRecTypeId != null) {
-                contactsNeedAccounts.add(con);
-            }
+        if (triggerAction == TDTM_Runnable.Action.AfterInsert) {
+            for (Contact con : newContacts) {
+                //Account needs to be created if (1) Account does not exist or (2) Account model is not blank
+                if (con.AccountId == null && defaultRecTypeId != null) {
+                    contactsNeedAccounts.add(con);
+                }
 
-            //AFTER INSERT
-            if (triggerAction == TDTM_Runnable.Action.AfterInsert) {
                 if (insertWrapper == null) {
                     insertWrapper = new HandleInsertWrapperLogic();
                 }
 
                 insertWrapper.handleInsertData(con);
+
+                conProcessed += 1;
             }
 
-            //AFTER UPDATE
-            if (triggerAction == TDTM_Runnable.Action.AfterUpdate) {
+            TDTM_ProcessControl.setRecursionFlag(
+                TDTM_ProcessControl.registeredTrigger.ACCT_IndividualAccounts_TDTM_After_Insert,
+                true
+            );
+            handleInsertProcessing(insertWrapper, contactsNeedAccounts);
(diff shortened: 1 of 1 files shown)


force-app/main/default/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -277,7 +277,14 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
                     numOfContactsLeft = accountIdNumOfCons.get(con.AccountId);
                 }
 
-                updateWrapper.handleUpdateData(con, oldContacts[conProcessed], triggerAction, numOfContactsLeft);
+                Map<Id, Contact> oldContactByContactId = new Map<Id, Contact>(oldContacts);
```
MatthewBlanski, line 280: This map should be instantiated before the loop, likely at the beginning of the method. Within this code, you are instantiating the map once per contact rather than once at all.

MatthewBlanski (changes requested):
The map should be instantiated earlier in the method.

force-app/main/default/classes/ACCT_IndividualAccounts_TDTM.cls
xinnanlisf: That's good catch. On it.

force-app/main/default/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -246,25 +246,42 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
...
-            //Account needs to be created if (1) Account does not exist or (2) Account model is not blank
-            if (con.AccountId == null && defaultRecTypeId != null) {
-                contactsNeedAccounts.add(con);
-            }
 
-            //AFTER INSERT
-            if (triggerAction == TDTM_Runnable.Action.AfterInsert) {
+        if (triggerAction == TDTM_Runnable.Action.AfterInsert) {
+            for (Contact con : newContacts) {
+                //Account needs to be created if (1) Account does not exist or (2) Account model is not blank
```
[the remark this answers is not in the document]
xinnanlisf: Probably not. We can even get rid of the comment as the code explains itself.

force-app/main/default/classes/ACCT_IndividualAccounts_TDTM.cls
```diff
@@ -277,25 +294,32 @@ public class ACCT_IndividualAccounts_TDTM extends TDTM_Runnable {
...
+            }
 
-        //Set recursion flag
-        if (triggerAction == TDTM_Runnable.Action.AfterInsert) {
             TDTM_ProcessControl.setRecursionFlag(
-                TDTM_ProcessControl.registeredTrigger.ACCT_IndividualAccounts_TDTM_After_Insert,
+                TDTM_ProcessControl.registeredTrigger.ACCT_IndividualAccounts_TDTM_After_Update,
                 true
             );
+            handleInsertProcessing(insertWrapper, contactsNeedAccounts);
```
salvadorsfdc, line 311: I was just curious to know: if the if statement already passed and we know for sure this is an afterUpdate, why would we need to call handleInsertProcessing()?

salvadorsfdc (commented):
Just added a question, but the rest looks good to me.

salvadorsfdc (approved):
looks good!

force-app/main/default/classes/ACCT_IndividualAccounts_TDTM.cls
MatthewBlanski: We're making a minimum viable change here. So this is a single method refactor. In an ideal world we would evaluate the logic beyond this method and make a larger change. However, this is a small work item and the impact of goofing up this class can prevent contact inserts and updates, so we're playing this safe and small.

salvadorsfdc: sounds good, thanks for explaining the scenario.

xinnanlisf:
@MatthewBlanski, looks like you still have to approve this PR.

MatthewBlanski (approved):

PR #1845, opened 2021-08-05
adding changes to comparison

xinnanlisf:
We are adding `inactive` fields to address utility methods for comparisons and queries. Also adding them to other classes if related to queries. Note them some of the classes have to be prettier-ed as well.

- The reason why we are not adding the fields to empty check methods is that we don't consider inactive empty addresses as not empty.
- We are delaying some of the tests writing to resolve another bug.
- This functionality won't be tested until merged to actual logic branch and changes to the actual logic are in place.

# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


MatthewBlanski (changes requested):
This code is functional. We have a few methods where we are not handling syncing inactive address:

copyAddressObj
copyAddressStdSObj
copyAddressAddrSObj

Please address those use cases.

MatthewBlanski (approved):

PR #1852, opened 2021-08-16
inactive address logic

xinnanlisf:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/default/classes/ADDR_Addresses_UTIL.cls
```diff
@@ -44,6 +44,8 @@ public with sharing class ADDR_Addresses_UTIL {
     /* @description map from Country Code to Country Name */
     private static Map<String, String> validCountriesByCode = new Map<String, String>();
 
+    private final static Set<String> INACTIVE_ADDRESS_FIELDS_PREFIX = new Set<String>{ 'Mailing', 'Billing' };
```
MatthewBlanski, line 47: Using a constant here is a smart play. Perhaps add a comment here about how we are only supporting inactive address for Mailing Address on Contact and Billing Address on Account or extract logic checks to a method to make it abundantly clear what we are doing and why.

force-app/main/default/classes/ADDR_Addresses_UTIL.cls
```diff
@@ -404,6 +411,16 @@ public with sharing class ADDR_Addresses_UTIL {
         sobjDst.put(strFieldPrefixDst + 'Latitude', sobjSrc.get(strFieldPrefixSrc + 'Latitude'));
         sobjDst.put(strFieldPrefixDst + 'Longitude', sobjSrc.get(strFieldPrefixSrc + 'Longitude'));
 
+        if (
```
MatthewBlanski, line 414: This is a good addition for a minimum change. In an ideal world, we would go a lot further here, but we are working with our hands tied. It might be worth extracting this check to a method or changing the constant name to something that indicates it is the prefix for address types that support inactive address functionality.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -37,8 +37,130 @@ private with sharing class ADDR_Addresses_UTIL_TEST {
      ****************************************************** isAddressEmpty ****************************************************
      **************************************************************************************************************************/
     @isTest
-    static void testIsAddressEmptyFalse() {
+    static void testIsAddressEmptyFalseMailingStreetNotEmpty() {
         List<Address__c> addresses = UTIL_UnitTestData_TEST.getMultipleTestAddresses(1);
```
MatthewBlanski, line 41: I know this might be a little silly here, but are we getting significant value from making an address record and then setting everything to null? Why not start with a blank address record if we're not performing DML.

Not a needed change now, but maybe something to look at in the future. I feel like `Address__c testAddress = new Address__c(MailingStreet__c = 'some random mailing street');` might be an easier read.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -95,7 +217,7 @@ private with sharing class ADDR_Addresses_UTIL_TEST {
     }
 
     @isTest
-    static void testIsContactAccAddrInfoSameFalseDifferentAddressField() {
+    static void testIsContactAccAddrInfoSameDifferentMailingCity() {
```
MatthewBlanski, line 220: Thank you for making the test method names more clear.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -110,6 +232,86 @@ private with sharing class ADDR_Addresses_UTIL_TEST {
         System.assertEquals(false, isContactAccAddrInfoSame);
     }
 
+    @isTest
+    static void testIsContactAccAddrInfoSameDifferentMailingCounty() {
+        Contact contact = UTIL_UnitTestData_TEST.getContact();
+        List<Account> accounts = UTIL_UnitTestData_TEST.getMultipleTestAccounts(
+            1,
+            UTIL_Describe_API.getBizAccRecTypeID()
```
MatthewBlanski, line 240: Just a preference here, but do we still want to be relying on our API within our internal code? Not a necessary change.
```suggestion
            UTIL_Describe.getBizAccRecTypeID()
```

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
```diff
@@ -110,6 +232,86 @@ private with sharing class ADDR_Addresses_UTIL_TEST {
         System.assertEquals(false, isContactAccAddrInfoSame);
     }
 
+    @isTest
+    static void testIsContactAccAddrInfoSameDifferentMailingCounty() {
+        Contact contact = UTIL_UnitTestData_TEST.getContact();
+        List<Account> accounts = UTIL_UnitTestData_TEST.getMultipleTestAccounts(
```
MatthewBlanski, line 238: Just a note that I was initially concerned about this and had to sanity check that we weren't setting address information in our unit test data factory for contact and account creation. If we ever do that we will invalidate assumptions in these test classes.

force-app/main/default/classes/ADDR_Contact_TEST.cls
```diff
@@ -631,31 +734,67 @@ public with sharing class ADDR_Contact_TEST {
...
             } else {
-                System.assertEquals(False, addr.Default_Address__c); 
+                System.assertEquals(false, addr.Default_Address__c);
             }
-        }                
+        }
+    }
+
+    @isTest
+    private static void updateExistingContactMailingFieldsInactiveAddress() {
```
MatthewBlanski, line 775: This test is failing. Please rectify it.

MatthewBlanski (changes requested):
After examining this, I'm noticing a number of improvements in how we're handling tests and how they are being applied. There's a unit test failing and some oddities around formatting for the tests that might make them a faster read. Only the unit test change is truly necessary.

force-app/main/default/classes/ADDR_Addresses_UTIL_TEST.cls
xinnanlisf: That's a good point.

force-app/main/default/classes/ADDR_Addresses_UTIL.cls
```diff
@@ -714,7 +764,7 @@ public with sharing class ADDR_Addresses_UTIL {
                 Address__c addrMatch = null;
                 for (Address__c childAddr : childAddrs) {
                     //Only include address type in the test if the new one isn't empty.
-                    if (!isAddressChanged(addr, childAddr, addr.Address_Type__c != null)) {
+                    if (!isAddressInfoSame(addr, childAddr, addr.Address_Type__c != null)) {
```
xinnanlisf, line 767: @MatthewBlanski this is the trick i was referring to last week. Basically we will keep the same `same` comparison method so that it will consider address with different inactive status to share the same address info. Changes to inactive fields will still be recognized as we have an updated method for isAddressChanged. 

Sadly this is not confusing at all but it is how address is working right by always creating and deduplicating.

PR #1862, opened 2021-09-02
Release gating error handling and jest tests (W-9536676)

iskander-sfdc:
Release gating error handling and display
Unit tests
Jest tests for all release gating components
WI: [W-9536676](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000PwwCYAS/view)

# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes

### DIFF
--- a/force-app/main/default/labels/CustomLabels.labels-meta.xml
+++ b/force-app/main/default/labels/CustomLabels.labels-meta.xml
@@ -1951,6 +1951,14 @@
         <value>We couldn&apos;t access valid product information for {0}. Contact Salesforce Customer Support for help.</value>
     </labels>
     <labels>
+        <fullName>stgErrorReleaseGateActivateErrorResponse</fullName>
+        <categories>Settings</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Message when there was an error during the release gate activation</shortDescription>
+        <value>We coudn&apos;t activate the release gate {0}. Contact Salesforce Customer Support for help. Additional details: {1}</value>
+    </labels>
+    <labels>
         <fullName>stgErrorSettingsTitle</fullName>
         <categories>Settings</categories>
         <language>en_US</language>
@@ -4005,6 +4013,6 @@
         <language>en_US</language>
         <protected>true</protected>
         <shortDescription>Release gate in progress activation toast message</shortDescription>
-        <value>We are activating {0}. This process might take some time, please check back later.</value>
+        <value>We are activating {0}. This process can take some time. Check back in a bit.</value>
     </labels>    
 </CustomLabels>
--- a/force-app/main/default/labels/CustomLabels.labels-meta.xml
+++ b/force-app/main/default/labels/CustomLabels.labels-meta.xml
@@ -1956,7 +1956,7 @@
         <language>en_US</language>
         <protected>true</protected>
         <shortDescription>Message when there was an error during the release gate activation</shortDescription>
-        <value>We coudn&apos;t activate the release gate {0}. Contact Salesforce Customer Support for help. Additional details: {1}</value>
+        <value>We couldn&apos;t activate {0}. See Additional details for more information, or contact Salesforce Support for help.&#13;&#10;&#13;&#10;Additional details: {1}</value>
     </labels>
     <labels>
         <fullName>stgErrorSettingsTitle</fullName>
--- a/force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
+++ b/force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
@@ -49,9 +49,8 @@ public virtual with sharing class ReleaseGateAPIService {
      */
     public ReleaseGateAPIService(ProductRegistryModel releaseGateRegistry) {
         if (releaseGateRegistry.action != ProductRegistryMapper.RELEASE_GATING_ACTION) {
-            throw new ReleaseGateProductNotFoundException(
-                'Incorrect release gate action specified in product registry: ' + releaseGateRegistry.action,
-                null,
+            throw new ReleaseGateProductRegistryException(
+                Label.stgErrorSettingsProductInvalidAction,
                 releaseGateRegistry
             );
         }
@@ -74,14 +73,14 @@ public virtual with sharing class ReleaseGateAPIService {
                 this.callableInstance = this.createReleaseGateCallableInstance();
             } catch (System.NullPointerException npe) {
                 String errorMessage = String.format(
-                    'Class: {0} with namespace {1} could not get loaded',
+                    Label.stgErrorSettingsProductInvalidClass,
                     new List<object>{ this.releaseGateRegistry.classname, this.releaseGateRegistry.namespace }
                 );
                 UTIL_Debug.debug(System.LoggingLevel.ERROR, errorMessage);
                 throw new ReleaseGateProductNotFoundException(errorMessage, npe, this.releaseGateRegistry);
             } catch (System.TypeException te) {
                 String errorMessage = String.format(
-                    'Class: {0} with namespace {1} could not get loaded',
+                    Label.stgErrorSettingsProductInvalidClass,
                     new List<object>{ this.releaseGateRegistry.classname, this.releaseGateRegistry.namespace }
                 );
                 UTIL_Debug.debug(System.LoggingLevel.ERROR, errorMessage);
@@ -105,16 +104,31 @@ public virtual with sharing class ReleaseGateAPIService {
      * @return           A GetReleaseGateActionResultModel
      */
     public virtual GetReleaseGateActionResultModel getReleaseGates() {
-        Object result = this.makeCall(
-            GET_ACTION,
-            new Map<String, Object>{ VERSION_PARAMETER => this.releaseGateRegistry.apiVersion }
-        );
-        GetReleaseGateActionResultModel resultModel = (GetReleaseGateActionResultModel) JSON.deserialize(
-            (String) result,
-            GetReleaseGateActionResultModel.class
-        );
-        this.verifyCallableResult(resultModel, GET_ACTION);
-        return resultModel;
+        //Customize application perm is required for all release gating actions
+        this.locateUserPermissionService().checkCustomizeApplicationForCurrentUser();
+        try {
+            Object result = this.makeCall(
+                GET_ACTION,
+                new Map<String, Object>{ VERSION_PARAMETER => this.releaseGateRegistry.apiVersion }
+            );
+            GetReleaseGateActionResultModel resultModel = (GetReleaseGateActionResultModel) JSON.deserialize(
+                (String) result,
+                GetReleaseGateActionResultModel.class
+            );
+            return resultModel;
+        } catch (System.JSONException je) {
+            String errorMessage = String.format(
+                Label.stgErrorSettingsProductInvalidResponse,
+                new List<Object>{ this.releaseGateRegistry.name }
+            );
+            UTIL_Debug.debug(System.LoggingLevel.ERROR, je.getMessage());
+            throw new ReleaseGateCallableException(errorMessage, je, this.releaseGateRegistry, null);
+        } catch (ReleaseGateException rge) {
+            throw rge;
+        } catch (Exception ex) {
+            UTIL_Debug.debug(System.LoggingLevel.ERROR, ex.getMessage());
+            throw new ReleaseGateCallableException(ex.getMessage(), ex, this.releaseGateRegistry, null);
+        }
     }
 
     /**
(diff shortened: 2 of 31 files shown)


force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1950,6 +1950,14 @@
         <shortDescription>Message when EDA can&apos;t access valid product information</shortDescription>
         <value>We couldn&apos;t access valid product information for {0}. Contact Salesforce Customer Support for help.</value>
     </labels>
+    <labels>
+        <fullName>stgErrorReleaseGateActivateErrorResponse</fullName>
+        <categories>Settings</categories>
+        <language>en_US</language>
+        <protected>true</protected>
+        <shortDescription>Message when there was an error during the release gate activation</shortDescription>
+        <value>We couldn&apos;t activate {0}. See Additional details for more information, or contact Salesforce Support for help.&#13;&#10;&#13;&#10;Additional details: {1}</value>
```
salvadorsfdc, line 1959: I was just wondering if the "&#13;&#10;&#13;&#10;" is needed. Is that a return? I didn't see any wrap in the message.

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
```diff
@@ -105,16 +104,31 @@ public virtual with sharing class ReleaseGateAPIService {
...
+                (String) result,
+                GetReleaseGateActionResultModel.class
+            );
+            return resultModel;
+        } catch (System.JSONException je) {
+            String errorMessage = String.format(
+                Label.stgErrorSettingsProductInvalidResponse,
+                new List<Object>{ this.releaseGateRegistry.name }
+            );
+            UTIL_Debug.debug(System.LoggingLevel.ERROR, je.getMessage());
```
salvadorsfdc, line 124: I think we could remove this debugging line

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
```diff
@@ -105,16 +104,31 @@ public virtual with sharing class ReleaseGateAPIService {
...
+            throw new ReleaseGateCallableException(errorMessage, je, this.releaseGateRegistry, null);
+        } catch (ReleaseGateException rge) {
+            throw rge;
+        } catch (Exception ex) {
+            UTIL_Debug.debug(System.LoggingLevel.ERROR, ex.getMessage());
```
salvadorsfdc, line 129: I think we could remove this debugging line

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
```diff
@@ -124,20 +138,40 @@ public virtual with sharing class ReleaseGateAPIService {
...
+        try {
+            Object result = this.makeCall(
+                ACTIVATE_ACTION,
+                new Map<String, Object>{
+                    VERSION_PARAMETER => this.releaseGateRegistry.apiVersion,
+                    PRODUCT_PARAMETER => productName,
+                    GATE_PARAMETER => gateName
+                }
+            );
+            ActivateReleaseGateActionResultModel resultModel = (ActivateReleaseGateActionResultModel) JSON.deserialize(
```
salvadorsfdc, line 152: I think we might need to also catch the System.JSONException for the activateReleaseGate method logic.

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
```diff
@@ -147,8 +181,6 @@ public virtual with sharing class ReleaseGateAPIService {
      * @return            Callable return result
      */
     protected virtual Object makeCall(String action, Map<String, Object> parameters) {
-        //Customize application perm is required for all release gating actions
-        this.locateUserPermissionService().checkCustomizeApplicationForCurrentUser();
         Callable callableInstance = this.locateReleaseGateCallable();
         UTIL_Debug.debug(
```
salvadorsfdc, line 185: Some debugs we could potentially remove?

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
```diff
@@ -157,21 +189,7 @@ public virtual with sharing class ReleaseGateAPIService {
...
-            UTIL_Debug.debug(System.LoggingLevel.ERROR, errorMessage);
-            throw new ReleaseGateCallableException(
-                'An unhandled exception occurred during execution: ' + errorMessage,
-                ex,
-                this.releaseGateRegistry,
-                null
-            );
-        }
+        result = callableInstance.call(action, parameters);
         UTIL_Debug.debug(
```
salvadorsfdc, line 193: Some debugs we could potentially remove?

force-app/main/releaseGating/classes/ReleaseGateVMapper.cls
```diff
@@ -156,6 +158,22 @@ public virtual with sharing class ReleaseGateVMapper {
...
     }
 
+    /**
+     * @description Verify if the result is successful, throw ReleaseGateCallableException if not
+     * @param  resultModel Callable action result
+     */
+    private void verifyCallableResult(CallableResult resultModel) {
+        if (resultModel.success) {
+            return;
+        } else {
```
salvadorsfdc, line 168: following what Matt told me about guard clauses, we could potentially change the 'else' to be another if. something like `if(resultModel.success == false){...}`

force-app/main/releaseGating/classes/ReleaseGateVMapper.cls
```diff
@@ -156,6 +158,22 @@ public virtual with sharing class ReleaseGateVMapper {
...
+            String errorMessage = resultModel.error != null
+                ? String.format('{0} ({1})', new List<Object>{ resultModel.error.message, resultModel.error.code })
+                : 'Unspecified error';
+            UTIL_Debug.debug(System.LoggingLevel.ERROR, errorMessage);
```
salvadorsfdc, line 172: Some debugs we could potentially remove?

force-app/main/releaseGating/lwc/releaseManagement/releaseManagement.js
```diff
@@ -51,4 +54,39 @@ export default class EdcReleaseManagement extends NavigationMixin(LightningEleme
...
         }
     }
+
+    handleErrorMessage(event) {
+        event.stopPropagation();
+        this.displayErrorMessage(this.getErrorMessage(event.detail));
+    }
+
+    displayErrorMessage(errorMessage) {
+        console.error(errorMessage);
```
salvadorsfdc, line 64: It seems this line is throwing an error while running `yarn test:unit --coverage` on my mac.
image

salvadorsfdc (changes requested):
It looks good, just some minor changes regarding debug logs and some errors I noticed. Good job!

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
MatthewBlanski: This is one of those locations that I feel we may get value out of. CCE has requested that we incorporate more UTIL_Debug than less within our code, as it's only viewable if the debug logs are turned on and accessible within the package.

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
MatthewBlanski: Again here with a value added for CCE.

force-app/main/releaseGating/classes/ReleaseGateVMapper.cls
MatthewBlanski: In this case we don't even need the else condition.

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
iskander-sfdc: This exception is caught on line 167 in a generic `catch (Exception ex)`
The reason for this is that we need to use stgErrorReleaseGateActivateErrorResponse label for all activation exceptions including JSONException. This label has formatting with error details

force-app/main/default/labels/CustomLabels.labels-meta.xml
iskander-sfdc: The standard toast component doesn't support cr/lf so those will be replaced with space
If we decide to use a messagebox in future cr/lf will be visible

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
iskander-sfdc: Yes, here and below debug statements were added for Support/CCE to help with debugging

force-app/main/releaseGating/lwc/releaseManagement/releaseManagement.js
iskander-sfdc: This test is reporting a console error but not failing. The reason for this is to inform an advanced user about an error response from the server in the browser console
```
Test Suites: 11 passed, 11 total
Tests:       53 passed, 53 total
```

force-app/main/releaseGating/classes/ReleaseGateVMapper.cls
iskander-sfdc: I copied this method from another class which Matt reviewed before.
The reason for this statement was to enforce code coverage for both true and false flows

force-app/main/default/labels/CustomLabels.labels-meta.xml
salvadorsfdc: oh ok, sounds good. Thanks for explaining that.

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
salvadorsfdc: oh great! good to know, thanks for explaining it

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
salvadorsfdc: oh great! good to know, thanks for explaining it

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
salvadorsfdc: Thanks for explaining that @iskander-sfdc.
I normally like to make the exception catching handling more granular so that we can add specific logic later or to explain other devs that exception can be returned, but maybe that is a personal style. @MatthewBlanski any final thoughts?

iskander-sfdc: I can catch JSONException, but the catch code will be exactly the same as the generic catch on line 167
Or I can catch and rethrow, but this will alter the exception stack trace

force-app/main/releaseGating/classes/ReleaseGateVMapper.cls
salvadorsfdc: @iskander-sfdc I think Matt is referring to doing something like this:

```
    private void verifyCallableResult(CallableResult resultModel) {
        if (resultModel.success) {
            return;
        } 
            String errorMessage = resultModel.error != null
                ? String.format('{0} ({1})', new List<Object>{ resultModel.error.message, resultModel.error.code })
                : 'Unspecified error';
            UTIL_Debug.debug(System.LoggingLevel.ERROR, errorMessage);
            throw new ReleaseGateResultException(errorMessage);
        
    }
```

force-app/main/releaseGating/lwc/releaseManagement/releaseManagement.js
salvadorsfdc: Now that I know that logging the errors is really important for CCE, I see the need for this. I was just wondering why JEST doesn't like that line. This is not a show stopper, but we should probably look for a workaround or patch to make our test coverage results cleaner.  :-) hehe I originally thought there was something wrong, but I also noticed the test didn't fail.

force-app/main/releaseGating/classes/ReleaseGateVMapper.cls
salvadorsfdc: thanks :-)

salvadorsfdc (approved):
Looks good, just waiting for Matt to give a final thought on the JSONException handling piece.

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
MatthewBlanski: Please catch explicitly for the JSONException so we force ourselves to cover that code and communicate that it is an expected use case.

I feel it is best to tell a story. In cases where catching and rethrowing is more explicit, then do so with a custom exception. The stack trace is ultimately a means to communicate failure without ambiguity, so if a specific catch clears up ambiguity (such as in this use case), then we add value by handling it individually. We also have the opportunity to create a debug entry for that failure case. If it is ambiguous, then preserving the stack trace can be relevant.

In general, I would recommend saving generic exception catching for changing the message to be more explicit about where something went wrong for support and for when a single part should not break the whole.

Ultimately this is to the benefit of supporting the product.

MatthewBlanski (changes requested):
Changes requested on the JSON exception handling.

force-app/main/releaseGating/classes/ReleaseGateAPIService.cls
iskander-sfdc: Thanks @MatthewBlanski , I changed the code to catch this exception. Could you please check one more time?

salvadorsfdc: it looks good!

MatthewBlanski (approved):

PR #1872, opened 2021-09-27
[W-9536678] Release gating tile notifications

iskander-sfdc:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


force-app/main/educationCloudSettings/lwc/edcReleaseManagementCard/edcReleaseManagementCard.css
```diff
@@ -0,0 +1,30 @@
+.progress-container {
+    height: 30px;
+}
+.progressring-container {
+    max-width: 30px;
+}
+.progressring-content {
+    vertical-align: middle;
+}
```
MatthewBlanski, line 9: These classes are additional non lwc-token classes that we are adding into the design. I'm calling them out as typically we do want to avoid them. In particular, the pixel-specific widths are things I would prefer to avoid from a technical perspective. The relative vertical alignment may have a design token we can lean back on.

force-app/main/educationCloudSettings/lwc/edcReleaseManagementCard/edcReleaseManagementCard.css
```diff
@@ -0,0 +1,30 @@
...
+.notification-bell {
+    margin-top: -10px;
+    font-size: 15px;
+    display: block;
+    position: absolute;
```
MatthewBlanski, line 14: Consider replacing this with the `slds-is-absolute` class on the bell itself.
```suggestion
```

force-app/main/educationCloudSettings/lwc/edcReleaseManagementCard/edcReleaseManagementCard.css
```diff
@@ -0,0 +1,30 @@
...
+}
+.progressring-container {
+    max-width: 30px;
+}
+.progressring-content {
+    vertical-align: middle;
+}
+.notification-bell {
+    margin-top: -10px;
+    font-size: 15px;
```
MatthewBlanski, line 12: Absolute pixel references tend to be something we want to avoid. Is there a relative solution you can use in this case?

force-app/main/educationCloudSettings/lwc/edcReleaseManagementCard/edcReleaseManagementCard.html
```diff
@@ -22,14 +22,36 @@ <h3 slot="title" id="cardTitle">
...
-                ></lightning-button>
+                <div class="slds-float_right">
+                    <div class="slds-grid slds-grid_vertical">
+                        <div class="slds-col slds-size_1-of-1">
+                            <div if:false={allReleaseGatesActive} class="notification-bell">&#x1F514;</div>
+                            <lightning-button
+                                label={labelReference.releaseManagementButton}
+                                title={labelReference.releaseManagementButtonA11y}
+                                aria-described-by="cardTitle"
+                                class="btnGoToReleaseManagement slds-float_left"
```
MatthewBlanski, line 33: Consider adding a QA Locator data attribute to this item. Using the class to find it is something that we would use internally, but for browser testing, the data attribute is our tool of choice.

MatthewBlanski (changes requested):
A few minor suggestions on using slds-classes and an ask to look at pixel widths and avoid them if possible. I do not have context for UX feedback on them. If they were implemented as part of UX feedback then dismiss them.

force-app/main/educationCloudSettings/lwc/edcReleaseManagementCard/edcReleaseManagementCard.css
iskander-sfdc: Changed to relative units (em). Explicit height is needed here to avoid UI jumping when all release gates are finished loading

force-app/main/educationCloudSettings/lwc/edcReleaseManagementCard/edcReleaseManagementCard.css
iskander-sfdc: Done, also replaced pixel sizes with relative units

force-app/main/educationCloudSettings/lwc/edcReleaseManagementCard/edcReleaseManagementCard.css
iskander-sfdc: Replaced with relative units. These styles are needed to absolute position the bell icon

force-app/main/educationCloudSettings/lwc/edcReleaseManagementCard/edcReleaseManagementCard.html
iskander-sfdc: Done, added "data-qa" attributes in a few palces instead

MatthewBlanski (approved):

PR #1895, opened 2021-11-03
Reworked existing unpackaged unit tests into the current format

MatthewBlanski:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


xinnanlisf (approved):
looks good, approved!

cumulusci.yml
```diff
@@ -608,29 +582,41 @@ flows:
...
+    qa_org_2gp_ftests:
+        group: "Org Setup"
+        description: Set up an org and run functional tests as a QA environment using a second-generation package
+        steps:
+            1:
+                flow: qa_org_2gp
+            2:
+                flow: run_functional_tests
+
+    run_functional_tests:
```
xinnanlisf, line 594: One only concern here is that this will deploy all the unpackaged test metadata to the `default` scratch org. Does that "pollute" the development environment since we want it to be as close as the customer environment?

xinnanlisf: so ideally we should have function_test_org and unpackaged_test org like:

- function_test(unpackaged)_org
-- qa_org_2gp
-- deploy metadata for a certain test type

And then we should have run_functional_test and run_unpackaged_test "only" running the test.

MatthewBlanski: One of our targets for releng is to make the flow an all-in-one, similar to ci-feature. We can individually run run_functional_tests on any managed org.

cumulusci.yml
```diff
@@ -608,29 +582,41 @@ flows:
...
+                when: "'hed' in org_config.installed_packages"
             3:
-                task: execute_primary_affiliation_mapping_apex
-                when: "'hed' not in org_config.installed_packages"
-            4:
-                task: deploy_unpackaged_relationship_autocreate_config
+                # execute tests in unmanaged package
+                task: run_tests
+                options:
+                    test_name_match: '%\_FTST'
```
iskander-sfdc, line 607: Need to add a blank namespace to make `run_tests` task to run these ftests
```suggestion
                    test_name_match: '%\_FTST'
                    namespace: ''
```

PR #1905, opened 2021-12-01
[W-9269471] Display latest unmapped record types for auto enrollment program mappings

kg345:
# Changes

- Pull the latest unmapped record types when creating new auto enrollment program mappings in Education Cloud Settings.

# Issues Closed

-[ W-9269471](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000LnH0YAK/view)

# Testing Notes
Test scenario updated
[T-5147768
](https://gus.lightning.force.com/lightning/r/ADM_Test_Scenario__c/a80AH0000003cIrYAI/view)

Manual Test to be automated in [W-9425585](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000OCgkYAG/view)

1. Select auto enrollment program mappings setting in Education Cloud Settings
2. Click on new button
3. Add a mapping for business organization record type
4. Click Save
5. Click new button
6. Verify the list of available record types - Business Organization record type should not display.


force-app/main/default/lwc/autoEnrollmentMappingModalBody/autoEnrollmentMappingModalBody.js
```diff
@@ -90,6 +84,7 @@ export default class autoEnrollmentMappingModalBody extends LightningElement {
     }
 
     handleAccountRecordTypeChange(event) {
+        this.accountRecordTypeComboboxVModel.value = event.detail.value;
```
salvadorsfdc, line 87: I was just wondering if we need this line

salvadorsfdc (approved):
Looks good, just 1 question that I posted, but not a big deal.

force-app/main/default/lwc/autoEnrollmentMappingModalBody/autoEnrollmentMappingModalBody.js
kg345: I had added it as the html displays the API name of the selected record type through `accountRecordTypeComboboxVModel.value`, but this value isn't set anymore when when using imperative apex as that only runs once and before user interaction.

Before, with the wire, the LWC dispatches the event that the value has changed and then the aura wrapper would set `newAccountRecordType` to the selected combobox value which would cause the wire in LWC to re-run as it's passed as a parameter (and the return data appropriately sets the selected value in `accountRecordTypeComboboxVModel.value`). 

With imperative, the aura wrapper changing the value of `newAccountRecordType` doesn't cause anything to happen (with how I did it) so I had added this line to capture the value before dispatching the event. It technically goes against the framework - I had initially thought of using a getter/setter for `newAccountRecordType` so that I could have more control to call the imperative apex again when the selection changes to match what the framework does with wires? It's more calls to the server and more code so I had ended up going with just this one line. Open to suggestions!

salvadorsfdc: @kg345 you're totally right, no need to change anything :-). Thanks for explaining this.

salvadorsfdc (approved):
looks good, approving

aarsvoboda (commented):
Looks good!

PR #1927, opened 2021-12-30
[W-10059980] Automate Test for Release Management without Access

kg345:
# Issues Closed
[W-10059980
](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000HzwyZYAR/view
)

# New Metadata

- Permission Set (unpackaged)
   - Internal_Only_for_Testing

# Testing Notes

- [T-6951039](https://gus.lightning.force.com/lightning/r/ADM_Test_Scenario__c/a80EE000000DKSLYA4/view)

- Created new User with Standard profile
- Created new permission set for internal testing that only has tab visibility access for Education Cloud Settings

### DIFF
--- a/robot/EDA/resources/EducationCloudSettingsPageObject.py
+++ b/robot/EDA/resources/EducationCloudSettingsPageObject.py
@@ -44,4 +44,4 @@ class EducationCloudSettingsPage(BaseEDAPage, HomePage):
         """
         locator = eda_lex_locators["eda_settings_new"]["error_toast"].format(value)
         self.selenium.wait_until_page_contains_element(locator, timeout=60, error=f'{locator} is not available')
-        self.selenium.wait_until_element_is_visible(locator, error= "{locator} is not displayed for the user")
+        self.selenium.wait_until_element_is_visible(locator, error= f'{locator} is not displayed for the user')
--- a/robot/EDA/resources/EducationCloudSettingsPageObject.py
+++ b/robot/EDA/resources/EducationCloudSettingsPageObject.py
@@ -42,8 +42,6 @@ class EducationCloudSettingsPage(BaseEDAPage, HomePage):
     def verify_error_is_displayed(self,value):
         """Verifies an error message displayed in a toast within education cloud settings
         """
-        locator = eda_lex_locators["eda_settings_new"]["error_toast"]
-        self.selenium.wait_until_element_is_visible(locator)
-        toastErrorMessage = self.selenium.get_webelement(locator).text
-        if value not in toastErrorMessage:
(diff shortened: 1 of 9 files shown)


kg345:
@xinnanlisf - Let me know your thoughts on whether this scenario is already covered by pre-existing jest tests or if you believe it's worth creating a robot test for this. I added the pre-existing jest tests to the test scenario that I think already cover this testing scenario (with mocked errors). I ended up only noticing that one jest test was duplicated so I removed it (only change in this PR).

xinnanlisf:
> @xinnanlisf - Let me know your thoughts on whether this scenario is already covered by pre-existing jest tests or if you believe it's worth creating a robot test for this. I added the pre-existing jest tests to the test scenario that I think already cover this testing scenario (with mocked errors). I ended up noticing that one jest test was duplicated so I removed it (only change in this PR).

Well i guess they are not the same type of tests in my opinion. Jest is mainly unit while robot is more like acceptance? Ideally, we could write really "simple" jest tests to verify if the code is working properly or not but really up to you as we already have tons of "fake" unit tests in eda code base.

kg345:
@xinnanlisf This is ready now with a new robot test. I ended up creating a new user with the standard profile and a permission set that just has tab visibility to the new Education Cloud Settings. That was done to have the ability, in robot tests, to give that user class/field accesses as needed for impartial permission scenarios.

robot/EDA/resources/EducationCloudSettingsPageObject.py
```diff
@@ -37,4 +37,13 @@ def click_app_in_edc_home(self,value):
         self.selenium.wait_until_page_contains_element(locator, timeout=60, error=f'{locator} is not available')
         self.selenium.wait_until_element_is_visible(locator,
                                                 error= "Element is not displayed for the user")
-        self.selenium.click_element(locator)
\ No newline at end of file
+        self.selenium.click_element(locator)
+
+    def verify_error_is_displayed(self,value):
```
xinnanlisf, line 42: This should work. The other option you can do is by passing the `value` to the locator and see if the browser can find it. I think most of other methods do that but it is really up to you.

xinnanlisf (approved):
Hi @kg345, left a few minor comments and I will go ahead and approve this. Nice work!

robot/EDA/resources/EducationCloudSettingsPageObject.py
kg345: Updated to match similar methods that pass the value to the locator for the browser to see if it's on the page and visible. Thanks!

PR #1928, opened 2022-01-03
[W-10302209] Feature Parameter - Auto-Enrollment Mappings

ZeeshanSyedH:
This child branch contains the Feature Parameter for auto-enrollment mappings, the associated logic & unit tests.

# Issues Closed
- [W-10302209](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000XUCgzYAH/view)

# New Metadata
- Count of Auto Enrollment Mappings

# Testing Notes
- [T-8180507](https://gus.lightning.force.com/lightning/r/ADM_Test_Scenario__c/a80EE000000mIc9YAE/view)

_Note: [Parent branch](https://github.com/SalesforceFoundation/EDA/tree/feature/238__zhussain_telemetryEnhancements) will consist of all aggregated [telemetry 238 deliverables](https://gus.lightning.force.com/lightning/r/ADM_Epic__c/a3QEE0000002Z9l2AE/related/Work__r/view), to be tested/verified as part of [W-10302537](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000XUui3YAD/view)._

### DIFF
--- a/force-app/main/default/classes/UTIL_OrgTelemetry.cls
+++ b/force-app/main/default/classes/UTIL_OrgTelemetry.cls
@@ -294,209 +294,116 @@ public without sharing class UTIL_OrgTelemetry {
      * custom object record count.
      */
     private void handleObjectUsageTelemetry() {
-        Integer countBadgeCredentials = 0;
-        Integer countHouseholdAccounts = 0;
-        Integer countLicenseCredentials = 0;
-        Integer countAdministrativeAccounts = 0;
-        Integer countCertificationCredentials = 0;
-        Integer countDegreeAcademicCertifications = 0;
-        Integer countDiplomaAcademicCertifications = 0;
-        Integer countCertificateAcademicCertifications = 0;
-
-        Integer numErrors = [SELECT COUNT() FROM Error__c];
-        Integer numFacilities = [SELECT COUNT() FROM Facility__c];
-        Integer numTermGrades = [SELECT COUNT() FROM Term_Grade__c];
-        Integer numCredentials = [SELECT COUNT() FROM Credential__c];
-        Integer numProgramPlans = [SELECT COUNT() FROM Program_Plan__c];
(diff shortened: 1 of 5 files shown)


aarsvoboda (approved):

force-app/main/default/classes/UTIL_OrgTelemetry.cls
```diff
@@ -466,5 +467,12 @@ public without sharing class UTIL_OrgTelemetry {
                     AND CreatedDate = LAST_N_DAYS:365
             ]
         );
+
+        featureManager.setPackageIntegerValue(
+            TelemetryParameterName.Data_CountAutoEnrollmentMappings.name(),
+            [
+                SELECT COUNT() FROM Affl_Mappings__c WHERE Auto_Program_Enrollment__c = true
+            ]
+        );
```
MatthewBlanski, line 476: In an ideal world, I would want this isolated into its own method with a name that states its purpose. The state of the old code does not tell an effective story. about what we are doing nor separate the concerns.

This is also a location where we might be well served by having the SOQL query assign to a local variable before setting the value, again for readability and purpose.

Note that this is more stylistic. Note that keeping it like this can help as multiple assignments to the same feature parameter during a test suite run can cause row lock issues.

MatthewBlanski (approved):
I left stylistic feedback, but the code is solid, functional, and in the style of its surrounding code. Make changes at your discretion.

force-app/main/default/classes/UTIL_OrgTelemetry.cls
ZeeshanSyedH: Readability enhanced by introducing metrics model wrapper ([reference commit](https://github.com/SalesforceFoundation/EDA/pull/1928/commits/0e5228f8955451722bc44f18be7032c81d3b61ea)).

kg345 (approved):
Looks good to me!

PR #1953, opened 2022-02-04
W-10475111 create robot locators for api 54 (236)

iskander-sfdc:
# Critical Changes

# Changes

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes
create an or using `qa_org` flow
run stable robot tests
```
cci task run robot --suites robot/EDA/tests --exclude unstable
```
Should work on 234 and 236 orgs


robot/EDA/resources/locators_53.py
```diff
@@ -0,0 +1,414 @@
+""" Locators for Winter '22 """
```
salvadorsfdc, line 1: I think this needs to be something like:
```
from locators_54 import *
import copy
eda_lex_locators = copy.deepcopy(eda_lex_locators)
#Override the locator selector value
eda_lex_locators['eda_settings_new']['dropdown_input']="//label[text()='{}']/../descendant::input[@role='combobox']"
```

salvadorsfdc (changes requested):

robot/EDA/resources/locators_53.py
iskander-sfdc: Committed this change. Robot tests in meta ci are passing

robot/EDA/resources/EDA.robot
```diff
@@ -135,11 +135,11 @@ Get Affiliation Mappings
     ${NS} =                 Get EDA namespace prefix
     [Arguments]             ${field_name}
     ${result} =             SOQL Query
-    ...                     SELECT ${NS}${field_name} FROM ${NS}Affl_Mappings__c
```
salvadorsfdc, line 138: I'm not sure about removing the NS variable from here @jonathanjayaraj @iskander-sfdc . It seems it was working fine previously.

robot/EDA/resources/EDA.robot
```diff
@@ -135,11 +135,11 @@ Get Affiliation Mappings
...
+    ...                     SELECT ${field_name} FROM ${NS}Affl_Mappings__c
     ${count} =              Get Affl Records Count       ${field_name}
     @{ValueList}=           Create List
     FOR     ${index}        IN RANGE    ${count}
-        ${name} =               Set variable    ${result}[records][${index}][${NS}${field_name}]
```
salvadorsfdc, line 142: I'm not sure about removing the NS variable from here. It seems it was working fine previously.

robot/EDA/resources/EDA.robot
```diff
@@ -174,7 +174,7 @@ Get Affl Records Count
     ${NS} =                 Get EDA namespace prefix
     [Arguments]             ${field_name}
     ${result} =             SOQL Query
-    ...                     SELECT COUNT(${NS}${field_name}) FROM ${NS}Affl_Mappings__c
+    ...                     SELECT COUNT(${field_name}) FROM ${NS}Affl_Mappings__c
```
salvadorsfdc, line 177: I'm not sure about removing the NS variable from here. It seems it was working fine previously.

robot/EDA/resources/EDA.robot
jonathanjayaraj: Correct. we'll need the ${NS} variable to add the EDA namespace if it is in a managed org. This was working previously.I'm about to try it in a regression org which is managed.

iskander-sfdc: Yes, this change fixes this test for a regression org which uses managed package

salvadorsfdc (approved):
looks good

robot/EDA/resources/EDA.robot
salvadorsfdc: just for reference the field_name in this case is name which doesnt require namespace

PR #1958, opened 2022-02-12
fix: labels for Non-Degree translations match case of field

mshanemc:
Translations `masterLabel` for Desired_Degree_Level__c are had `Non-Degree` while the picklist had Non-degree` so that they throw deploy errors.

This could have also been fixed by changing the picklist's label instead of making all the translations match, but that has more potential for breaking other things.

# Critical Changes

# Changes
- We fixed a bug that was causing deploy errors due to translations for Desired_Degree_Level__c not matching the case of the picklist options.

# Issues Closed

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes


mshanemc:
thanks for considering

PR #1983, opened 2023-01-24
Remove session ID usage

jcschultz:
The user’s session ID was being used to retrieve the namespace. This is no longer permitted and would fail security review. I’ve updated the code to allow the apex controller to pass the namespace to the UI instead.

I’ve also added a unit test.

Signed-off-by: jcschultz <jschultz@salesforce.com>

# Critical Changes

# Changes

# Issues Closed
[W-12151125](https://gus.my.salesforce.com/a07EE00001FWGYLYA5)

# New Metadata

## Unpackaged Metadata

# Deleted Metadata

# Testing Notes
- Spin up a scratch org
- Go to the EDA app (not the Education Data Architecture app)
- Go to the EDA Settings tab and verify that the page loads properly (should show the map SVG image and a message about settings moving to the EDU Cloud settings experience page)


mrrigney (approved):
Looks good to me and tested as expected, approving!

PR #1995, opened 2024-01-19
Update CODEOWNERS

kg345:


CODEOWNERS
```diff
@@ -1,3 +1,3 @@
 #ECCN: Open Source
-#GUSINFO: Education Data Architecture (EDA), Education Data Architecture (EDA)
-
+#GUSINFO: Rad-ish, Education Data Architecture & K-12 (MP)
```
robertbwatson, line 2: Are spaces okay? The other codeowners PRs don't have spaces, so just want to make sure this is fine

kg345: It's a good question - I was going off the confluence page where the examples show spaces in the names of product tag, at least.
https://confluence.internal.salesforce.com/display/corescm/Create+a+GUS-Aware+CODEOWNERS+File

kg345:
Closing + opened a new PR with correct branch naming
https://github.com/SalesforceFoundation/EDA/pull/1996

CODEOWNERS
robertbwatson: Okay. I just didn't know if "#GUSINFO: Rad-ish, Education Data Architecture & K-12 (MP)" vs "#GUSINFO:Rad-ish,Education Data Architecture & K-12 (MP)" made a difference, as I noticed all of the others are the latter.

kg345: It seems they [trim the line ](https://git.soma.salesforce.com/scm/codeowners_parser/blob/master/src/main/java/com/salesforce/codeowners/parser/CodeownerParser.java#L314) when the split the string (based on `#GUSINFO`) so I believe it should be ok with a space.

